PR #15, opened 2019-12-18
Added adjoints for evaluate

### LINKED ISSUE #12: kerneldiagmatrix issue with Zygote
The current implementation of kerneldiagmatrix can create a SegFault when using Zygote.
No MWE at the moment

theogf:
Added adjoints and tests until https://github.com/FluxML/Zygote.jl/issues/429 and https://github.com/FluxML/Zygote.jl/issues/385 are fixed.
Fixes #12 
Improves a bit the performance of kerneldiagmatrix

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,13 @@ os:
   - osx
 julia:
   - 1.0
+  - 1.2
+  - 1.3
   - nightly
+# because of Zygote needs to allow failing on nightly
+matrix:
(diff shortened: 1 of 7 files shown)


theogf:
Reverted the lines using `eachcol` and `eachrow` since it is not compatible with Julia 1.0

devmotion:
You could add Compat if you would rather want to use `eachcol` and `eachrow`.

PR #21, opened 2020-01-14
Depend on ZygoteRules instead of Zygote

devmotion:
IMO KernelFunctions should not depend on Zygote and all its dependencies but rather use [ZygoteRules](https://github.com/FluxML/ZygoteRules.jl) as dependency.

The latest release of SpecialFunctions broke the tests (Pkg can't resolve a compatible set of packages since SpecialFunctions 0.10 is used but ForwardDiff, an implicit dependency of Zygote, does not support it yet), so I fixed SpecialFunctions compatibility to 0.9 and fixed a deprecation warning.

BTW I found [CompatHelper](https://github.com/bcbi/CompatHelper.jl) quite helpful for being notified about new releases of dependencies :slightly_smiling_face:


devmotion:
I added SpecialFunctions 0.8 since SpecialFunctions 0.9 is only compatible with Julia 1.3.

theogf:
That looks great thanks for the PR

theogf:
Thanks for the `CompatHelper` suggestion, I never used it but it's super useful!

devmotion:
You're welcome :smiley: I like it a lot, one just has to be a bit careful that the new version is actually used in the tests (sometimes some other dependencies might block it).

PR #22, opened 2020-01-14
Remove metric fields

devmotion:
It seems the `metric` fields are not needed, it is sufficient to define the `metric(::Kernel)` trait.

### DIFF
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -9,7 +9,7 @@ Base.iterate(k::Kernel, ::Any) = nothing
 for k in [:ExponentialKernel,:SqExponentialKernel,:GammaExponentialKernel,:MaternKernel,:Matern32Kernel,:Matern52Kernel,:LinearKernel,:PolynomialKernel,:ExponentiatedKernel,:ZeroKernel,:WhiteKernel,:ConstantKernel,:RationalQuadraticKernel,:GammaRationalQuadraticKernel]
     @eval begin
         @inline (κ::$k)(d::Real) = kappa(κ,d) #TODO Add test
-        @inline (κ::$k)(x::AbstractVector{<:Real},y::AbstractVector{<:Real}) = kappa(κ,evaluate(κ.metric,transform(κ,x),transform(κ,y)))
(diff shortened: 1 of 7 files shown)


theogf:
Doesn't it create unnecessary allocations ?
Maybe an alternative would be to create `const` objects of each metric in the package.
```
const m_sqeuclidean = SqEuclidean()
metric(::SqExponential) = m_sqeuclidean
```

devmotion:
I just assumed the compiler is smart enough (and the [implementation in Distances.jl](https://github.com/JuliaStats/Distances.jl/blob/7f3a28c0d1372e3b3edbcbc28f00ba5645e1bbdb/src/metrics.jl#L274-L275) would be different otherwise) :smile: To be sure that there are no allocations, I used the following benchmark script:
```julia
using Distances
using BenchmarkTools

struct A
    metric::SqEuclidean
end

A() = A(SqEuclidean())

metric(::A) = SqEuclidean()

const SQEUCLIDEAN = SqEuclidean()
metric_const(::A) = SQEUCLIDEAN

evaluate_field(a::A, x, y) = evaluate(a.metric, x, y)
evaluate_function(a::A, x, y) = evaluate(metric(a), x, y)
evaluate_const(a::A, x, y) = evaluate(metric_const(a), x, y)

function test(N::Int)
    a = A()
    x = rand(N)
    y = rand(N)

    @btime evaluate_field($a, $x, $y)
    @btime evaluate_function($a, $x, $y)
    @btime evaluate_const($a, $x, $y)

    nothing
end
```
I got the following results:
```julia
julia> test(10_000)
  1.778 μs (0 allocations: 0 bytes)
  1.783 μs (0 allocations: 0 bytes)
  1.785 μs (0 allocations: 0 bytes)

julia> test(100_000)
  23.819 μs (0 allocations: 0 bytes)
  23.808 μs (0 allocations: 0 bytes)
  23.824 μs (0 allocations: 0 bytes)
```
So the approach shouldn't introduce any allocations.

theogf:
Thanks a lot for making all the benchmarks and your contribution!

PR #27, opened 2020-01-16
Fix StackOverflowError

devmotion:
This PR fixes a StackOverflowError that was introduced in https://github.com/theogf/KernelFunctions.jl/pull/26.

### DIFF
--- a/src/kernels/constant.jl
+++ b/src/kernels/constant.jl
@@ -7,7 +7,7 @@ struct ZeroKernel{Tr} <: Kernel{Tr}
     transform::Tr
 end
 
-ZeroKernel(t::Transform=IdentityTransform()) = ZeroKernel(t)
+ZeroKernel() = ZeroKernel(IdentityTransform())
 
 @inline kappa(κ::ZeroKernel, d::T) where {T<:Real} = zero(T)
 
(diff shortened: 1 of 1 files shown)


theogf:
Oh I see the problem! But it's a funny bug no? Shouldn't Julia throw an error on the creation on the function? Maybe this should be reported on Base (or is it already known)?

devmotion:
Hmm not sure, if this should be addressed on the language/compiler level. Which behaviour/implementation error do you want to disallow exactly, by the way? To define a function that calls itself again with the same unchanged arguments? Intuitively it feels like it's difficult to address all possible cases anyway, e.g., `f(x) = f(g(x))` yields a stack overflow error if `g === identity` but can be completely reasonable in other cases.

theogf:
I meant that `Kernel(t::Transform) = Kernel(t)` overwrote the default constructor `Kernel{Tr}(t::Tr) where {Tr<:Transform}`

devmotion:
No, it didn't, it just redefined the default outer constructor in a bad way.
```julia
using Test

struct A{T}
    a::T
end

A(t) = A(t)

@test A{Int}(0) isa A
@test_throws StackOverflowError A(0)
```
That's why one could have fixed the bug also by defining something like
```julia
A(t = 0) = A{typeof(t)}(t)

@test A(0) isa A
@test A() isa A
```
Only if one defines a custom inner constructor then Julia does not provide the default outer constructor (see https://docs.julialang.org/en/v1/manual/constructors/#Inner-Constructor-Methods-1).

PR #28, opened 2020-01-16
Fix Travis link

devmotion:


theogf:
Ups, bad copy/pasting from me

Issue #17, opened 2019-12-19
Transformations of kernels


willtebbutt:
Currently, kernels have a field `transform` e.g. [here](https://github.com/theogf/KernelFunctions.jl/blob/460ed8685018a35bac2bceaa2401639668a41f6e/src/kernels/exponential.jl#L12).

This is unnecessary, and creates additional code to maintain. I would argue that a better approach is to orthogonalise the design by removing the `transform` field from the kernels and creating a combinator kernel `Transform` that wraps the kernel. i.e.

```julia
struct TransformedKernel{Tt<:Transform, Tk<:Kernel} <: Kernel
    t::Tt
    k::Tk
end

function kernel_matrix(t::TransformedKernel, x::AbstractVector{<:Real})
    return kernel_matrix(t.k, apply(t.t, x))
end
```

This has the distinct advantage that you can transform the inputs to any kernel. For example, the [KernelSum](https://github.com/theogf/KernelFunctions.jl/blob/460ed8685018a35bac2bceaa2401639668a41f6e/src/kernels/kernelsum.jl#L14) currently doesn't have a `transform` field, nor should it. This functionality would be obtained for free via this design.

Stheno currently does something similar to this with [stretch](https://github.com/willtebbutt/Stheno.jl/blob/61236affc381d5cba16fab67b9a0d93226e8995d/src/gp/kernel.jl#L438), but I'm planning to move more closely towards the above design over time.

theogf:
I think here we are mostly dealing with a UI problem. I still maintain the idea that the `Kernel` type should contain `transform`. I like to be able to directly see what kind of kernel is used without having to check subparameters. With the solution you propose literally everything would be a `TransformedKernel`... 
> This has the distinct advantage that you can transform the inputs to any kernel. For example, the KernelSum currently doesn't have a transform field, nor should it. This functionality would be obtained for free via this design.
About this how about (taking #22 into account)
```
abstract type Kernel end
abstract type SingleKernel{Tr<:Transform} <: Kernel end
abstract type CompositeKernel <: Kernel end
struct KernelSum <: CompositeKernel
...
end
```
I don't like `SingleKernel` but it's just for the example

willtebbutt:
I partly agree it's a UI problem, but I'm more concerned about the maintainability of the internals and composability.

Maintability: having the transformation being part of the type adds extra code to maintain to every single kernel. This is as opposed to having the transformations being completely separate from the "base" kernels.

[...]

PR #29, opened 2020-01-19
Ease implementation of custom kernels

devmotion:
The main purpose of this PR is to simplify the implementation of custom kernels, in particular if they use a fixed transform. I added a test with a reimplementation of a squared exponential kernel with fixed identity transform to illustrate the minimum implementation that is currently required.

The main changes in this PR are
- the removal of (hopefully) all accesses of a `transform` field in generic methods,
- the definition of a default method for evaluating kernels with two inputs that makes use of the metric and transform of a kernel and is not restricted to arrays (hence could be used with real numbers or distributions as well),
- the use of the identity transform as default transform for kernels without parameters instead of `ScaleTransform(1.0)`,
- the possibility to apply an `IdentityTransform` to arbitrary types (hence it can be applied, e.g., also to real numbers and distributions).

I still think that it might be better to not have to define a transformation at all, IMO it would be more natural to be able to just define
```julia
struct MyKernel <: Kernel end

KernelFunctions.kappa(::MyKernel, d2::Real) = exp(-d2)
KernelFunctions.metric(::MyKernel) = SqEuclidean()
```

### DIFF
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -5,11 +5,14 @@ Base.length(::Kernel) = 1
 Base.iterate(k::Kernel) = (k,nothing)
 Base.iterate(k::Kernel, ::Any) = nothing
 
+# default fallback for evaluating a kernel with two arguments (such as vectors etc)
+kappa(κ::Kernel, x, y) = kappa(κ, evaluate(metric(κ), transform(κ, x), transform(κ, y)))
+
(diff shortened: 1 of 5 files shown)


theogf:
Now I realised, I misunderstood your PR, first I thought you wanted to implement @willtebbutt solution from #17 but you are just removing explicit dependence on the `.transform` field.
This looks like a great step to make the `Transform` transition better!

Issue #19, opened 2019-12-23
Nystrom approximations


IsakFalk:
It would be nice to include Nystrom approximations if we are to do the same for Random Features, as per #14.

@trthatcher has already put in work here through the archived [MLKernels](https://github.com/trthatcher/MLKernels.jl) in [nystrom.jl](https://github.com/trthatcher/MLKernels.jl/blob/master/src/nystrom.jl) over there. I haven't looked through it in detail, but looks fine to apply here?

theogf:
Yes it would be nice to have it! If you want to submit a PR to transfer the work of @trthatcher that would be amazing!

IsakFalk:
I'm a bit busy with ICML stuff now, but I can look into this after. Or if someone else is interested in tackling this!

IsakFalk:
I've ported the code (thanks @trthatcher!) to work with KernelFunctions. It seems to work but I need to run and add some tests. I'm not really sure what the method `syrk_wrapper!` does and searching doesn't turn up many results. Could also be a leftover from earlier Julia versions?

trthatcher:
It's just an `transpose(A)*A` or `A*transpose(A)` operation. The reference BLAS implementation outlines what it does here:

> http://www.netlib.org/lapack/explore-html/dc/d05/dsyrk_8f_source.html

I implemented Nystrom before there was a lazy transpose, so there was actual overhead for calling either of those operations. Using `transpose` and `mul!` should be fine with Julia 1.0.

Sorry guys! I'm tied up with other obligations, otherwise I'd try to be an active contributor.

IsakFalk:
@trthatcher no worries, thanks for taking time to respond and explaining!

IsakFalk:
Linking to the PR #30

IsakFalk:
Since we merged the #30 I'll mark this issue as resolved.

PR #30, opened 2020-01-26
Nystrom Approximation

IsakFalk:
Implements the Nystrom kernel matrix approximation in response to #19.

### DIFF
--- a/src/KernelFunctions.jl
+++ b/src/KernelFunctions.jl
@@ -17,10 +17,12 @@ export Transform, SelectTransform, ChainTransform, ScaleTransform, LowRankTransf
 export NystromFact, nystrom
 
 using Compat
+using LinearAlgebra
 using Distances, LinearAlgebra
 using SpecialFunctions: logabsgamma, besselk
 using ZygoteRules: @adjoint
 using StatsFuns: logtwo
+using StatsBase
 using PDMats: PDMat
 
 const defaultobs = 2
--- a/src/KernelFunctions.jl
+++ b/src/KernelFunctions.jl
@@ -14,6 +14,8 @@ export KernelSum, KernelProduct
 
 export Transform, SelectTransform, ChainTransform, ScaleTransform, LowRankTransform, IdentityTransform, FunctionTransform
 
+export NystromFact, nystrom
+
 using Compat
 using Distances, LinearAlgebra
 using SpecialFunctions: logabsgamma, besselk
@@ -41,6 +43,7 @@ include("matrix/kernelmatrix.jl")
 include("matrix/kernelpdmat.jl")
 include("kernels/kernelsum.jl")
 include("kernels/kernelproduct.jl")
+include("approximations/nystrom.jl")
 
 include("generic.jl")
 
--- a/src/KernelFunctions.jl
+++ b/src/KernelFunctions.jl
@@ -17,7 +17,6 @@ export Transform, SelectTransform, ChainTransform, ScaleTransform, LowRankTransf
 export NystromFact, nystrom
 
 using Compat
-using LinearAlgebra
 using Distances, LinearAlgebra
 using SpecialFunctions: logabsgamma, besselk
 using ZygoteRules: @adjoint
--- a/src/approximations/nystrom.jl
+++ b/src/approximations/nystrom.jl
@@ -5,7 +5,7 @@ function sampleindex(X::AbstractMatrix, r::Real; obsdim::Integer=defaultobs)
     0 < r <= 1 || throw(ArgumentError("Sample rate `r` must be in range (0,1]"))
     n = size(X, obsdim)
     m = ceil(Int, n*r)
-    S = StatsBase.sample(collect(1:n), m; replace=false, ordered=true)
+    S = StatsBase.sample(1:n, m; replace=false, ordered=true)
     return S
 end
 
@@ -87,9 +87,7 @@ Returns a `NystromFact` struct which stores a Nystrom factorization satisfying:
 """
 function nystrom(k::Kernel, X::AbstractMatrix, r::AbstractFloat; obsdim::Int=defaultobs)
     S = sampleindex(X, r; obsdim=obsdim)
-    C, Cs = nystrom_sample(k, X, S; obsdim=obsdim)
-    W = nystrom_pinv!(Cs)
-    return NystromFact(W, C)
+    return nystrom(k, X, S; obsdim=obsdim)
 end
 
 """
--- a/src/approximations/nystrom.jl
+++ b/src/approximations/nystrom.jl
@@ -1,0 +1,103 @@
+using StatsBase
+using LinearAlgebra
+# Following the algorithm by William and Seeger, 2001
+# Cs is equivalent to X_mm and C to X_mn
+
+function sampleindex(X::AbstractMatrix, r::AbstractFloat; obsdim::Integer=defaultobs)
+    0 < r <= 1 || throw(ArgumentError("Sample rate `r` must be in range (0,1]"))
+    n = size(X, obsdim)
+    m = ceil(Int, n*r)
+    S = StatsBase.sample(collect(1:n), m; replace=false, ordered=true)
+    return S
+end
+
+function nystrom_sample(k::Kernel, X::AbstractMatrix, S::Vector{<:Integer}; obsdim::Integer=defaultobs)
+    obsdim ∈ [1, 2] || throw(ArgumentError("`obsdim` should be 1 or 2 (see docs of kernelmatrix))"))
+    Xₘ = obsdim == 1 ? getindex(X, S, :) : getindex(X, :, S)
+    C = k(Xₘ, X; obsdim=obsdim)
+    Cs = getindex(C, :, S)
+    return (C, Cs)
+end
+
+function nystrom_pinv!(Cs::Matrix{T}, tol::T=eps(T)*size(Cs,1)) where {T<:AbstractFloat}
+    # Compute eigendecomposition of sampled component of K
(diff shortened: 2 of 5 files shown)


src/approximations/nystrom.jl
```diff
@@ -0,0 +1,103 @@
+using StatsBase
+using LinearAlgebra
+# Following the algorithm by William and Seeger, 2001
+# Cs is equivalent to X_mm and C to X_mn
+
+function sampleindex(X::AbstractMatrix, r::AbstractFloat; obsdim::Integer=defaultobs)
```
devmotion: Maybe change this to
```suggestion
function sampleindex(X::AbstractMatrix, r::Real; obsdim::Integer=defaultobs)
```
?

src/approximations/nystrom.jl
```diff
@@ -0,0 +1,103 @@
+using StatsBase
+using LinearAlgebra
```
devmotion: Maybe move the `using` statements to `src/KernelFunctions.jl`?

src/approximations/nystrom.jl
```diff
@@ -0,0 +1,103 @@
...
+@doc raw"""
+    NystromFact
+
+Type for storing a Nystrom factorization. The factorization contains two fields: `W` and
+`C`, two matrices satisfying:
+```math
+\mathbf{K} \approx \mathbf{C}^{\intercal}\mathbf{W}\mathbf{C}
+```
+"""
+struct NystromFact{T<:AbstractFloat}
```
devmotion: Maybe also
```suggestion
struct NystromFact{T<:Real}
```
here?

src/approximations/nystrom.jl
```diff
@@ -0,0 +1,103 @@
...
+Nystrom factorization satisfying:
+```math
+\mathbf{K} \approx \mathbf{C}^{\intercal}\mathbf{W}\mathbf{C}
+```
+"""
+function nystrom(k::Kernel, X::AbstractMatrix, S::Vector{<:Integer}; obsdim::Int=defaultobs)
+    C, Cs = nystrom_sample(k, X, S; obsdim=obsdim)
+    W = nystrom_pinv!(Cs)
+    T = typeof(first(W))
+    return NystromFact{T}(W, C)
```
devmotion: It shouldn't be necessary to compute `T`, one should be able to just call `NystromFact(W, C)`. If `W` and `C` can happen to be of different types, then IMO it would make more sense to specify an additional constructor, i.e.,
```julia
function NystromFact(W::Matrix{<:Real}, C::Matrix{<:Real})
    T = Base.promote_eltypeof(W, C)
    return NystromFact(convert(Matrix{T}, W), convert(Matrix{T}, C))
end
```

src/approximations/nystrom.jl
```diff
@@ -0,0 +1,103 @@
...
+```math
+\mathbf{K} \approx \mathbf{C}^{\intercal}\mathbf{W}\mathbf{C}
+```
+"""
+function nystrom(k::Kernel, X::AbstractMatrix, r::AbstractFloat; obsdim::Int=defaultobs)
+    S = sampleindex(X, r; obsdim=obsdim)
+    C, Cs = nystrom_sample(k, X, S; obsdim=obsdim)
+    W = nystrom_pinv!(Cs)
+    T = typeof(first(W))
+    return NystromFact{T}(W, C)
```
devmotion: Same here, one shouldn't have to compute types I guess.

IsakFalk:
@devmotion thanks for the pointers, I'll update the code!

src/KernelFunctions.jl
```diff
@@ -14,11 +14,15 @@ export KernelSum, KernelProduct
 
 export Transform, SelectTransform, ChainTransform, ScaleTransform, LowRankTransform, IdentityTransform, FunctionTransform
 
+export NystromFact, nystrom
+
 using Compat
+using LinearAlgebra
```
devmotion: LinearAlgebra is already imported in the next line, so I'd suggest you just remove this line.
```suggestion
```

src/approximations/nystrom.jl
```diff
@@ -0,0 +1,104 @@
+# Following the algorithm by William and Seeger, 2001
+# Cs is equivalent to X_mm and C to X_mn
+
+function sampleindex(X::AbstractMatrix, r::Real; obsdim::Integer=defaultobs)
+    0 < r <= 1 || throw(ArgumentError("Sample rate `r` must be in range (0,1]"))
+    n = size(X, obsdim)
+    m = ceil(Int, n*r)
+    S = StatsBase.sample(collect(1:n), m; replace=false, ordered=true)
```
devmotion: If I'm not mistaken (I quickly skimmed through StatsBase), you don't need `collect` here. I think you can just write
```suggestion
    S = StatsBase.sample(1:n, m; replace=false, ordered=true)
```

src/approximations/nystrom.jl
```diff
@@ -0,0 +1,104 @@
...
+
+Computes a factorization of Nystrom approximation of the square kernel matrix of data
+matrix `X` with respect to kernel `k` using a sample ratio of `r`.
+Returns a `NystromFact` struct which stores a Nystrom factorization satisfying:
+```math
+\mathbf{K} \approx \mathbf{C}^{\intercal}\mathbf{W}\mathbf{C}
+```
+"""
+function nystrom(k::Kernel, X::AbstractMatrix, r::AbstractFloat; obsdim::Int=defaultobs)
+    S = sampleindex(X, r; obsdim=obsdim)
```
devmotion: I think it would be clearer (and remove redundant code) if you do not copy
```julia
    C, Cs = nystrom_sample(k, X, S; obsdim=obsdim)
    W = nystrom_pinv!(Cs)
    return NystromFact(W, C)
```
but replace it with
```julia
    return nystrom(k, X, S; obsdim=obsdim)
```

src/approximations/nystrom.jl
```diff
@@ -0,0 +1,104 @@
...
+    0 < r <= 1 || throw(ArgumentError("Sample rate `r` must be in range (0,1]"))
+    n = size(X, obsdim)
+    m = ceil(Int, n*r)
+    S = StatsBase.sample(collect(1:n), m; replace=false, ordered=true)
+    return S
+end
+
+function nystrom_sample(k::Kernel, X::AbstractMatrix, S::Vector{<:Integer}; obsdim::Integer=defaultobs)
+    obsdim ∈ [1, 2] || throw(ArgumentError("`obsdim` should be 1 or 2 (see docs of kernelmatrix))"))
+    Xₘ = obsdim == 1 ? getindex(X, S, :) : getindex(X, :, S)
```
devmotion: That's my personal preference, but I think it would be easier to read if you write
```suggestion
    Xₘ = obsdim == 1 ? X[S, :] : X[:, S]
```
and
```julia
    Cs = C[:, S]
```
two lines below.

src/approximations/nystrom.jl
```diff
@@ -0,0 +1,102 @@
...
+@doc raw"""
+    NystromFact
+
+Type for storing a Nystrom factorization. The factorization contains two fields: `W` and
+`C`, two matrices satisfying:
+```math
+\mathbf{K} \approx \mathbf{C}^{\intercal}\mathbf{W}\mathbf{C}
+```
+"""
+struct NystromFact{T<:Real}
```
theogf: Maybe we should parametrize `NystromFact` with the types of `W` and `C`, so we can eventually use special matrix types: so 
```
struct NystromFact{TW<:AbstractMatrix{<:Real},TC<:AbstractMatrix{<:Real}}
W::TW
C::TC
end
```

Also would it be interesting to create another struct where CWC is already precomputed and where it acts like an `AbstractMatrix`

src/approximations/nystrom.jl
```diff
@@ -0,0 +1,102 @@
...
+    S = sampleindex(X, r; obsdim=obsdim)
+    return nystrom(k, X, S; obsdim=obsdim)
+end
+
+"""
+    nystrom(CᵀWC::NystromFact)
+
+Compute the approximate kernel matrix based on the Nystrom factorization.
+"""
+function kernelmatrix(CᵀWC::NystromFact{<:AbstractFloat})
```
theogf: `NystromFact{<:Real}` ?

src/approximations/nystrom.jl
```diff
@@ -0,0 +1,102 @@
...
+    nystrom(k::Kernel, X::Matrix, r::AbstractFloat; obsdim::Int=defaultobs)
+
+Computes a factorization of Nystrom approximation of the square kernel matrix of data
+matrix `X` with respect to kernel `k` using a sample ratio of `r`.
+Returns a `NystromFact` struct which stores a Nystrom factorization satisfying:
+```math
+\mathbf{K} \approx \mathbf{C}^{\intercal}\mathbf{W}\mathbf{C}
+```
+"""
+function nystrom(k::Kernel, X::AbstractMatrix, r::AbstractFloat; obsdim::Int=defaultobs)
```
theogf: `r::Real` ?

IsakFalk: Changed all of the `AbstractFloat`'s to `Real`s.

src/approximations/nystrom.jl
IsakFalk: To be clear, implement the necessary methods as described in the [interface](https://docs.julialang.org/en/v1/manual/interfaces/#man-interface-array-1) for `AbstractArray` part of the julia docs?

theogf: The second part would be more of an additional suggestion. Maybe we can create a second PR to make a decision about it.

IsakFalk: How is this different from calling `kernelfunctions(nf::NystromFact)`?

Issue #31, opened 2020-01-27
Create a trainable function for Flux compatibility 


theogf:
It would be nice to have a `trainable` function that would avoid having to use `@functor` with Flux, selecting automatically parameters that can be optimized.

PR #32, opened 2020-01-28
Removing transform field and creating TransformedKernel (and ScaledKernel)

theogf:
Following the discussion in #17 I removed the transform field (and the parametrization on it). Now all base kernels default on `IdentityTransform`.
I created on top `TransformedKernel` which contains a transform and a kernel and `ScaledKernel` which contains a premultiplying scalar.
Left to do is 
- [x]  Decide on nomenclature for constructors (for now I implemented a toy example with the same names in lowercases
- [x] Create appropriate tests
- [x] Improve the printing of the kernels ( I have some ideas on this )
- [ ] Adapt documentation and examples


src/kernels/constant.jl
```diff
@@ -1,55 +1,50 @@
...
-params(k::ConstantKernel) = (params(k.transform),k.c)
-opt_params(k::ConstantKernel) = (opt_params(k.transform),k.c)
-
-ConstantKernel(c::Real=1.0) = ConstantKernel(IdentityTransform(),c)
+function ConstantKernel(c::Tc=1.0) where {Tc<:Real}
+    ConstantKernel{Tc}(c)
+end
 
-ConstantKernel(t::Tr,c::Tc=1.0) where {Tr<:Transform,Tc<:Real} = ConstantKernel{Tr,Tc}(t,c)
+params(k::ConstantKernel) = (k.c)
```
devmotion: Is `params` (and `opt_params`) supposed to return a tuple? This will just return a scalar value.

theogf: This is just bad leftovers I did for something else. This is related to #31 . I will make a more thorough PR at some point.

src/kernels/scaledkernel.jl
```diff
@@ -0,0 +1,20 @@
+struct ScaledKernel{Tk<:Kernel,Tσ<:Real} <: Kernel
+    kernel::Tk
+    σ::Vector{Tσ}
```
devmotion: I'm curious, why is the scaling parameter a vector? Couldn't you just use
```suggestion
    σ::Tσ
```
?

theogf: The idea is that it makes the types much more Zygote compatible. One can then optimize the parameters in place, by passing the reference of the parameter. A scalar would not allow this. One other option would be to use a `Ref{Ts}` but I find this quite ugly in general!

devmotion: Hmm, I'm pretty sure that not using a vector should be completely fine with Zygote - at least I would be surprised if it's not: https://fluxml.ai/Zygote.jl/dev/#Structs-and-Types-1

theogf: Sorry I was not clear. It's actually to be more compatible with `Flux`. 
The whole implicit gradient approach allows you to save the reference to parameters (via an `IdDict` and apply the gradients directly in place without having to build new objects. But this is only possible on mutable object so for a scalar it would not work. Or maybe I got everything wrong about Flux :sweat_smile:

devmotion:
Do we actually need an `IdentityTransform`? I thought one possibility would be to just implement kernels without considering transforms at all and not having to define any `transform` function.  One could just define general transformations such as `TransformedKernel` or `ScaledKernel` that implement all required methods (such as `kappa` etc.) by applying the transformations and otherwise forwarding to the corresponding function of the underlying kernel, such as e.g.
```julia
kappa(κ::TransformedKernel, d) = kappa(κ.kernel, d)
kappa(κ::TransformedKernel, x, y) = kappa(κ.kernel, κ.transform(x), κ.transform(y))
```
Maybe one could define some more convenient methods such as
```julia
transform(f, k::Kernel) = TransformedKernel(k, f)
Base.:*(s::Real, k::Kernel) = ScaledKernel(k, s)
Base.:*(k::Kernel, s::Real) = s * k
```
In general, I think it might be a bit confusing that TransformedKernel transforms the inputs whereas ScaledKernel transforms the evaluated kernel function. I'm not sure if this could be avoided by different or more explicit names.

devmotion:
> Why not for the first one although I start to feel we are overloading `transform` with a lot of unrelated stuff.

Oh I thought that could be the only implementation of `transform`. I assumed that the implementation of `TransformedKernel` could be such that one never has to worry or deal with transformations - everything regarding transformations is handled by `TransformedKernel`, one would always just evaluate `kappa`, `metric`, `kernelmatrix` etc. for a given kernel.

theogf:
Oh that's a perfectly fair point then!

src/kernels/transformedkernel.jl
```diff
@@ -0,0 +1,13 @@
+struct TransformedKernel{Tk<:Kernel,Tr<:Transform} <: Kernel
```
devmotion: Maybe one could remove the `Transform` type and allow arbitrary types and functions as transformations?

theogf: The idea is to have wrapper around all types.

src/kernels/transformedkernel.jl
```diff
@@ -0,0 +1,13 @@
...
+    kernel::Tk
+    transform::Tr
+end
+
+@inline transform(k::TransformedKernel) = k.transform
```
willtebbutt: Do we need a transform function anymore?

src/generic.jl
```diff
@@ -1,5 +1,4 @@
-@inline metric(κ::Kernel) = κ.metric
-
+@inline transform(::Kernel) = IdentityTransform()
```
willtebbutt: Do we need a transform function anymore?

src/kernels/kernelsum.jl
```diff
@@ -11,18 +11,18 @@ kernelmatrix(k,X) == kernelmatrix(k1+k2,X)
 kweighted = 0.5*k1 + 2.0*k2
 ```
 """
-struct KernelSum{Tr} <: Kernel{Tr}
+struct KernelSum <: Kernel
     kernels::Vector{Kernel}
     weights::Vector{Real}
-    function KernelSum{Tr}(kernels::AbstractVector{<:Kernel},weights::AbstractVector{<:Real}) where {Tr}
-        new{Tr}(kernels,weights)
+    function KernelSum(kernels::AbstractVector{<:Kernel},weights::AbstractVector{<:Real})
```
willtebbutt: Can probably get rid of this constructor entirely

src/kernels/scaledkernel.jl
willtebbutt: What's the issue with using a `Ref` rather than a `Vector` here? I would have thought that a reference to a `Real` would be more readable than a `Vector{<:Real}` that you have to know is  always of length `1`?

theogf:
> Do we actually need an IdentityTransform? I thought one possibility would be to just implement kernels without considering transforms at all and not having to define any transform function.

I went to make the changes to drop the need for `IdentityTransform` and faced an issue : 
we still want to compute the pairwise matrix using `Distances.jl` (more efficient this way) and for this we need to preapply a transform.
A solution would be that `transform(::BaseKernel,X)` simply returns `X`

theogf:
About the `Ref` vs `Vector` problem:
They have the same size :
```
sizeof(Ref(3.0)) ## 8
sizeof([3.0]) ## 8
```
But the functions used to set values to a `Ref` are different then from a `Vector`.
The advantage would be that when one collects all the parameters of a complex kernel, the results is a large collection of arrays and not a mix of ref and vectors which have to be treated differently

devmotion:
> I went to make the changes to drop the need for IdentityTransform and faced an issue :
we still want to compute the pairwise matrix using Distances.jl (more efficient this way) and for this we need to preapply a transform.
> A solution would be that transform(::BaseKernel,X) simply returns X

My suggestion would be to define `kernelmatrix!` etc. for TransformedKernel, I think that should solve the issue. E.g.,
```julia
# default implementation without considering transforms at all
function kernelmatrix!(
        K::Matrix,
        κ::Kernel,
        X::AbstractMatrix;
        obsdim::Int = defaultobs
        )
        @assert obsdim ∈ [1,2] "obsdim should be 1 or 2 (see docs of kernelmatrix))"
        if !check_dims(K,X,X,feature_dim(obsdim),obsdim)
            throw(DimensionMismatch("Dimensions of the target array K $(size(K)) are not consistent with X $(size(X))"))
        end
        map!(x->kappa(κ,x),K,pairwise(metric(κ),X,dims=obsdim))
end

# special implementation for TransformedKernel
function kernelmatrix!(
        K::Matrix,
        κ::TransformedKernel,
        X::AbstractMatrix;
        obsdim::Int = defaultobs
        )
       kernelmatrix!(K, κ.kernel, κ.transform(X, obsdim); obsdim = obsdim)
end
```

theogf:
> My suggestion would be to define kernelmatrix! etc. for TransformedKernel, I think that should solve the issue. E.g.,

The problem with this is that we end up having the exact problem we were trying to avoid! :laughing:
Dispatching `kernelmatrix` functions is really painful because you need to treat all cases (X,XY,diagX).
Doing :
```
transform(κ::Kernel,x;obsdim=defaultobs) = x
transform(κ::TransformedKernel,x;obsdim=defaultobs) = transform(κ.transform,x,obsdim=obsdim)
```
Would produce the exact same result.

devmotion:
By the way, one should probably better use `RefValue` instead of `Vector`: https://discourse.julialang.org/t/ref-vs-zero-dimensional-arrays/24434/11
```julia
julia> struct S
           s::Ref{Int}
       end

julia> struct T
           s::Base.RefValue{Int}
       end

julia> T(x::Int) = T(Ref(x))
T

julia> f(x) = x.s[]

julia> a = S(1)
S(Base.RefValue{Int64}(1))

julia> b = T(1)
T(Base.RefValue{Int64}(1))

julia> @code_warntype f(a)
Variables
  #self#::Core.Compiler.Const(f, false)
  x::S

Body::Any
1 ─ %1 = Base.getproperty(x, :s)::Ref{Int64}
│   %2 = Base.getindex(%1)::Any
└──      return %2

julia> @code_warntype f(b)
Variables
  #self#::Core.Compiler.Const(f, false)
  x::T

Body::Int64
1 ─ %1 = Base.getproperty(x, :s)::Base.RefValue{Int64}
│   %2 = Base.getindex(%1)::Int64
└──      return %2
```

willtebbutt:
Having now thought about this a bit more, I'm not sure that there's any way around what @devmotion proposes (which is also what I had imagined happening). The reason being that, for example, `KernelSum` doesn't define `kappa` or `metric` because it doesn't make sense for it, and I would very much like to be able to transform the inputs to a `KernelSum`.

AFAICT, the `kappa` + `metric` abstraction works really well for base kernels (I'm going to implement it in `Stheno` shortly so that I'm closer to being ready for `KernelFunctions` integration -- very excited about that), but doesn't work for composite kernels such as `KernelSum`, `KernelProduct` and `TransformedKernel`.

devmotion:
I thought about this a bit more as well, and not having a `transform` trait still feels like the more natural thing to do. I think @willtebbutt pointed out the strongest argument in favour of dropping `transform` and just implementing `kernelmatrix` and `kernelmatrix!` for `TransformedKernel`.

> The problem with this is that we end up having the exact problem we were trying to avoid! :laughing:

I thought this PR tries to solve the problem that we have to carry around `Transform` type parameters for all kernels since some kernels do not allow for a natural/reasonable definition of `Transform`s, so  it seems the same reason also motivates dropping the `transform` trait?

> Dispatching kernelmatrix functions is really painful because you need to treat all cases (X,XY,diagX).

IMO the amount of functions that have to be added is not unreasonably large - and at the same time it allows us to remove all uses of `transform` :slightly_smiling_face:

theogf:
Ok, I will implement these changes then.
I just need to put this on hold, as I have a deadline end of the week :)

willtebbutt:
No worries. No prizes for guessing what the deadline is though haha

theogf:
> My suggestion would be to define `kernelmatrix!` etc. for TransformedKernel, I think that should solve the issue. E.g.,
> 
> ```julia
> # default implementation without considering transforms at all
> function kernelmatrix!(
>         K::Matrix,
>         κ::Kernel,
>         X::AbstractMatrix;
>         obsdim::Int = defaultobs
>         )
>         @assert obsdim ∈ [1,2] "obsdim should be 1 or 2 (see docs of kernelmatrix))"
>         if !check_dims(K,X,X,feature_dim(obsdim),obsdim)
>             throw(DimensionMismatch("Dimensions of the target array K $(size(K)) are not consistent with X $(size(X))"))
>         end
>         map!(x->kappa(κ,x),K,pairwise(metric(κ),X,dims=obsdim))
> end
> 
> # special implementation for TransformedKernel
> function kernelmatrix!(
>         K::Matrix,
>         κ::TransformedKernel,
>         X::AbstractMatrix;
>         obsdim::Int = defaultobs
>         )
>        kernelmatrix!(K, κ.kernel, κ.transform(X, obsdim); obsdim = obsdim)
> end
> ```

`κ.transform(X,obsdim)` does not seem natural to me, looks a lot like OOP which we don't necessarily want. 
Here is my general take given all the inputs :
I would leave `transform` alone, but still overload `kernelmatrix` like : 
```
kernelmatrix(k::TransformedKernel,X,obsdim) = kernelmatrix(k.kernel,transform(k.transform,X,obsdim),obsdim)
```

theogf:
Maybe an `apply(k.transform,X,obsdim)` ?

willtebbutt:
> κ.transform(X,obsdim) does not seem natural to me, looks a lot like OOP which we don't necessarily want.

Seems like a matter of taste to me, can't say that I have a strong preference either way :)

devmotion:
> κ.transform(X,obsdim) does not seem natural to me, looks a lot like OOP which we don't necessarily want.
> Here is my general take given all the inputs :
> I replace transform by apply (not exported), and overload kernelmatrix like :

IMO it doesn't matter. It seems there's only an advantage in using `apply`, if the method is not only defined for `::TransformedKernel` (i.e., not specifically `kernelmatrix(k::TransformedKernel, ...)`) but some more general type of kernels which might require transforms to be applied in a different way (and in that case probably one might want to use `kernel(k)` instead of `k.kernel` for the same reasons).

devmotion:
I was playing around with this PR a bit since I would like to move (almost) all kernel-related functionality of my package to KernelFunctions. However, I noticed that the current implementation leads to a severe performance regression when using `ScaleTransform`. That is both due to the fact that its parameter `s` is not a scalar and (more importantly) due to the allocation of a large number of vectors when evaluating `kappa(kernel, x, y)`. A simple example that I benchmarked shows:
```julia
# using `exponentialkernel` (i.e. no `ScaleTransform`)
1.137 ms (0 allocations: 0 bytes)

# using `exponentialkernel(0.1)`
2.432 ms (39805 allocations: 6.07 MiB)
```
Maybe one could add more efficient implementations such as
```julia
function kappa(κ::TransformedKernel{<:ExponentialKernel,<:ScaleTransform}, x, y)
    k = κ.kernel
    s = κ.transform.s[]
    kappa(k, s * evaluate(metric(k), x, y))
end
```
(probably in a more modular/general way) for the default kernels? With this implementation and
using `RefValue` I get
```julia
1.009 ms (5 allocations: 80 bytes)
```
By switching to a scalar field `s` in `ScaleTransform` I can reduce the allocations to
```julia
1.002 ms (0 allocations: 0 bytes)
```

theogf:
When you say you use `kappa(k,x,y)` you mean you loop over all vectors x and y?
Could you share the benchmarking code you used?

devmotion:
It's different estimators and I don't need all pairwise distances for all of them, so I'm evaluating just the combinations I need, e.g., in https://github.com/devmotion/CalibrationErrors.jl/blob/1cad3a183ea4c0f34758dd938dfd3b8cec393409/src/skce/generic.jl#L224-L228. The code is a bit out-dated, I haven't pushed my latest changes.

theogf:
Ah I understand now what you propose, you want to have  specialized cases for some  specific `kappa` cases. It would only work for `ScaleTransform` actually
Maybe we can add this in a different PR.

devmotion:
> Ah I understand now what you propose, you want to have specialized cases for some specific kappa cases. It would only work for ScaleTransform actually
> Maybe we can add this in a different PR.

Yes, that was my idea. IMO the `ScaleTransform` is so simple that it should not lead to such a severe performance regression and hence could be improved (at least for the default kernels) by some special implementations. Just thought I bring it up right away, since I was a bit surprised about the regressions.

theogf:
Ok I got a two liner : 
```
kappa(κ::TransformedKernel{<:Kernel,<:ScaleTransform}, x, y) = kappa(kernel(κ), _scale(κ)*evaluate(metric(κ),x, y))
_scale(κ::TransformedKernel) = first(κ.transform.s)^(metric(κ) isa Euclidean ? 1 : 2)
```

devmotion:
> Ok I got a two liner :

I'm a bit worried about kernels with different metrics. Maybe we could dispatch on `metric(κ.kernel)` and add an implementation for `::SqEuclidean` and `::Euclidean` (these are the only metrics used in KernelFunctions so far?) and use a fallback in all other cases. Something like
```julia
kappa(κ::TransformedKernel{<:Kernel,<:ScaleTransform}, x, y) = kappa(κ.kernel, _scale(κ.transform, metric(κ.kernel), x, y))
_scale(t::ScaleTransform, metric::Euclidean, x, y) =  first(t.s) * evaluate(metric, x, y)
_scale(t::ScaleTransform, metric::Euclidean, x, y) =  first(t.s)^2 * evaluate(metric, x, y)
_scale(t::ScaleTransform, metric, x, y) = evaluate(metric, apply(t, x), apply(t, y))
```

src/kernels/constant.jl
```diff
@@ -1,55 +1,49 @@
...
-
-ConstantKernel(t::Tr,c::Tc=1.0) where {Tr<:Transform,Tc<:Real} = ConstantKernel{Tr,Tc}(t,c)
+params(k::ConstantKernel) = (k.c,)
+opt_params(k::ConstantKernel) = (k.c,)
 
-@inline kappa(κ::ConstantKernel,x::Real) = κ.c
+@inline kappa(κ::ConstantKernel,x::Real) = κ.c*one(x)
 
-metric(::ConstantKernel) = Delta()
+@inline metric(::ConstantKernel) = Delta()
```
devmotion: As a side remark, IMO usually it's not required to add `@inline`, in particular not for small functions such as `metric`. In these cases, I think usually one can trust the compiler, as stated in the docstring:
```julia
help?> @inline
  @inline

  Give a hint to the compiler that this function is worth inlining.

  Small functions typically do not need the @inline annotation, as the compiler does it automatically. By using @inline on bigger
  functions, an extra nudge can be given to the compiler to inline it. 
```

theogf:
To add up on the massive changes of this PR : 
- Creation of the abstract type `BaseKernel`, which makes it more practical to create subsequent generic functions
- `BaseKernel` with parameters, for example `ConstantKernel`, have now a constructor based on keywords, to avoid confusion when `constantkernel(3.0)` is called for example. To get the old behaviour one needs to call `constantkernel(3.0,c=1.0)`. Related to this, I kept the same naming for the variables, but this means some keywords are unicode characters (alpha,gamma etc), I will change it to verbose names.
- I went with the approach of naming the `TransformedKernel` wrappers by changing the name to lowercase, but I  think we can maybe remove the `kernel` at the end.
- I put some deprecations for calling the previous constructors with a `Transform`

src/generic.jl
```diff
@@ -1,33 +1,36 @@
...
+kappa(κ::Kernel, x, y) = kappa(κ, evaluate(metric(κ), x, y))
+kappa(κ::TransformedKernel, x, y) = kappa(kernel(κ), apply(κ.transform,x), apply(κ.transform,y))
+kappa(κ::TransformedKernel{<:BaseKernel,<:ScaleTransform}, x, y) = kappa(κ, _scale(κ.transform, metric(κ), x, y))
+_scale(t::ScaleTransform, metric::Euclidean, x, y) =  first(t.s) * evaluate(metric, x, y)
+_scale(t::ScaleTransform, metric::Union{SqEuclidean,DotProduct}, x, y) =  first(t.s)^2 * evaluate(metric, x, y)
+_scale(t::ScaleTransform, metric, x, y) = evaluate(metric, apply(t, x), apply(t, y))
 
 ### Syntactic sugar for creating matrices and using kernel functions
-for k in [:ExponentialKernel,:SqExponentialKernel,:GammaExponentialKernel,:MaternKernel,:Matern32Kernel,:Matern52Kernel,:LinearKernel,:PolynomialKernel,:ExponentiatedKernel,:ZeroKernel,:WhiteKernel,:ConstantKernel,:RationalQuadraticKernel,:GammaRationalQuadraticKernel]
+for k in subtypes(BaseKernel)
```
devmotion: Unfortunately, that does not add the syntactic sugar for user-defined `BaseKernel`s - in the future (when support for Julia < 1.3 is dropped) this can be done in a nice way due to https://github.com/JuliaLang/julia/pull/31916: one can just defines
```julia
(κ::BaseKernel)(d::Real) = kappa(κ, d)
```
and so on.

src/generic.jl
```diff
@@ -1,33 +1,36 @@
...
 
-### Transform generics
-@inline transform(κ::Kernel) = κ.transform
-@inline transform(κ::Kernel, x) = transform(transform(κ), x)
-@inline transform(κ::Kernel, x, obsdim::Int) = transform(transform(κ), x, obsdim)
-
-## Constructors for kernels without parameters
-for kernel in [:ExponentialKernel,:SqExponentialKernel,:Matern32Kernel,:Matern52Kernel,:ExponentiatedKernel]
+for k in Symbol.(subtypes(BaseKernel))
+    k = Symbol(string(k)[17:end])
```
devmotion: I guess slightly less confusing would be
```julia
for k in nameof.(subtypes(BaseKernel))
```

src/generic.jl
theogf: Oh I did not know there was this new feature in 1.3.
I was thinking for now and as an alternative to create a macro that would do the same on a user-defined kernel. Including creating the `TransformedKernel` wrapper

src/generic.jl
theogf: Damn I spent hours trying to look for a solution on this :sweat_smile:

src/generic.jl
```diff
@@ -1,33 +1,36 @@
...
+    new_k = Symbol(lowercase(string(k)))
```
devmotion: I'm not completely sold on the idea of using lowercase functions here but I'm also not fully convinced by using the uppercase `k` either. IMO, the lowercase definitions add quite a bit of inconsistency, since now some kernels can be constructed with lowercase function whereas others don't, everyone who implements custom `BaseKernel`s has to copy these definitions (and in contrast to the syntactic sugar above I don't see how this problem could ever be solved by KernelFunctions, even on Julia >= 1.3), and one can construct kernels now in two different ways (e.g., `sqexponentialkernel()` and `SqExponentialKernel()`) which feels weird. That's why I'm inclined to say, maybe one should rather implement the uppercase version, such as `$k(ρ::Real; args...) = TransformedKernel($k(;args...),ScaleTransform(ρ))` - even though in this case it might come as a surprise that the return value is not of type `$k`. On the other hand, that should be fine I guess?

theogf: I think that's the real problem, this would create a very problematic type inconsistency...
For the user defined case, again I was thinking on making a macro. That would automatically create these functions.

theogf:
```julia
function wrap_kernel(k) #::KernelFunctions.BaseKernel
    new_k = Symbol(lowercase(string(k)))
    return @eval begin
        (κ::$k)(d::Real) = kappa(κ,d) #TODO Add test
        (κ::$k)(x::AbstractVector{<:Real}, y::AbstractVector{<:Real}) = kappa(κ, x, y)
        (κ::$k)(X::AbstractMatrix{T}, Y::AbstractMatrix{T}; obsdim::Integer=defaultobs) where {T} = kernelmatrix(κ, X, Y, obsdim=obsdim)
        (κ::$k)(X::AbstractMatrix{T}; obsdim::Integer=defaultobs) where {T} = kernelmatrix(κ, X, obsdim=obsdim)
        $new_k(;args...) = $k(;args...)
        $new_k(ρ::Real;args...) = TransformedKernel($k(;args...),ScaleTransform(ρ))
        $new_k(ρ::AbstractVector{<:Real};args...) = TransformedKernel($k(;args...),ARDTransform(ρ))
        $new_k(t::Transform;args...) = TransformedKernel($k(;args...),t)
        export $new_k
    end
end
```

devmotion: That (`@eval` inside a function) does not feel right, IMO this is the perfect use case for a proper macro, i.e., `macro wrap_kernel(T) ... end` where `T` is the type of the custom user-defined `BaseKernel`. That feels a lot cleaner to me.

theogf: I am working on this on Slack right now . I am very bad at understanding macros

theogf:
```
macro wrap_kernel(k) #::KernelFunctions.BaseKernel
    new_k = esc(Symbol(lowercase(string(k))))
    k = esc(k)
    return quote
        (κ::$k)(d::Real) = kappa(κ,d) #TODO Add test
        (κ::$k)(x::AbstractVector{<:Real}, y::AbstractVector{<:Real}) = kappa(κ, x, y)
        (κ::$k)(X::AbstractMatrix{T}, Y::AbstractMatrix{T}; obsdim::Integer=defaultobs) where {T} = kernelmatrix(κ, X, Y, obsdim=obsdim)
        (κ::$k)(X::AbstractMatrix{T}; obsdim::Integer=defaultobs) where {T} = kernelmatrix(κ, X, obsdim=obsdim)
        $new_k(;args...) = $k(;args...)
        $new_k(ρ::Real;args...) = TransformedKernel($k(;args...),ScaleTransform(ρ))
        $new_k(ρ::AbstractVector{<:Real};args...) = TransformedKernel($k(;args...),ARDTransform(ρ))
        $new_k(t::Transform;args...) = TransformedKernel($k(;args...),t)
        export $new_k
    end
end
```
With some support

devmotion: I guess
```julia
macro wrap_kernel(k)
    new_k = Symbol(lowercase(string(k)))
    return quote
        (κ::$k)(d::Real) = kappa(κ, d)
        (κ::$k)(x::AbstractVector{<:Real}, y::AbstractVector{<:Real}) = kappa(κ, x, y)
        (κ::$k)(X::AbstractMatrix{<:Real}, Y::AbstractMatrix{<:Real}; obsdim::Integer = defaultobs) = kernelmatrix(κ, X, Y, obsdim=obsdim)
        (κ::$k)(X::AbstractMatrix{<:Real}; obsdim::Integer=defaultobs) = kernelmatrix(κ, X, obsdim=obsdim)
        $new_k(; args...) = $k(; args...)
        $new_k(ρ::Real; args...) = TransformedKernel($k(; args...), ScaleTransform(ρ))
        $new_k(ρ::AbstractVector{<:Real};args...) = TransformedKernel($k(;args...),ARDTransform(ρ))
        $new_k(t::Transform;args...) = TransformedKernel($k(;args...),t)
        export $new_k
    end |> esc
end
```
should work as well (maybe one has to add `KernelFunctions.` to the non-exported types? :thinking:). I guess it could also be nice to wrap it inside an `if` block like
```julia
    if VERSION < v"..."
        return quote end
     else
        return quote
            ....
        end
    end
```
and define the methods on the abstract type `BaseKernel` in KernelFunctions. The correct version number can be found with the script https://github.com/JuliaLang/julia/blob/master/contrib/commit-name.sh.

willtebbutt: I've been battling with a similar UI issue over on Stheno for similar reasons.

As of v0.6.0, there are no lower-case names, and no extra methods that let you provide length scales / variances to kernels. Rather, I've just added a function called `kernel` that is just shorthand for stretching and scaling. i.e.
```julia
kernel(Matern52(); l=5.0, s=1.2)
```
is equivalent to
```julia
1.2 * stretch(Matern52(), 1 / 5.0)
```
This should be read as "give me a Matern - 5/2 kernel with length scale 5 and variance 1.2".

Dunno what you guys think to this approach -- I like it a lot from a consistency perspective, and how nicely it plays with new user-defined kernels. I like it less from the UI perspective, but I've been unable to come up with a better alternative.

I agree with @devmotion that you either have to make lower-case _or_ upper-case names user-facing, but not both because that creates ambiguity.

devmotion: > Dunno what you guys think to this approach -- I like it a lot from a consistency perspective, and how nicely it plays with new user-defined kernels. I like it less from the UI perspective, but I've been unable to come up with a better alternative.

I agree, this approach seems to be quite consistent and clean from a developer perspective, but maybe a bit too verbose from a user perspective. It would solve the issues we discussed above, I guess, but unfortunately I don't have an immediate idea for how to come up with a better UI. Maybe one could define a macro that rewrites
```julia
@kernel 1.2 * Matern52(1 / 5.0)
```
to `1.2 * stretch(Matern52(), 1 / 5.0)`? But it seems that could not handle the case `Matern52(v)` where `v` is a vector?

theogf: > I agree, this approach seems to be quite consistent and clean from a developer perspective, but maybe a bit too verbose from a user perspective.

I completely agree, this is a tad much to ask from a lambda user.

> ```@kernel 1.2 * Matern52(1 / 5.0)```

I like the macro approach! This would simplify and unify the creation of kernels. We just need to see how to deal with parametrized kernels like `Matern(nu=2.0)` or `GammaExponential(gamma=3.0)`

willtebbutt: To be consistent with the UI discussed above, you would probably want
```julia
@kernel 1.2 * Matern52(5.0)
```
to become
```julia
1.2 * stretch(Matern52(), 1 ./ 5.0)
```
so it should be possible to make it work with a vector just by always using the dot. Definitely couldn't handle a matrix of parameters, but I think that's fine as it doesn't really make sense anyway, (the inverse not being defined).

The other small issue is what to do about kernels that themselves have parameters e.g. rational quadratic? Would we assume that, if there's more than one argument, everything but the last should be passed directly to the kernel constructor? One of the reasons I quite like the function-based approach is because it's quite clear what happens in this case:
```julia
kernel(RationalQuadraticKernel(2.0); l=l, s=s)
```

Again, I'm not sold on all of the properties of this approach, but I think it offers a reasonable tradeoff between providing a consistent UI that is predictable and simple to understand from a user-perspective, ease of maintenance, and verbosity.

willtebbutt: > I completely agree, this is a tad much to ask from a lambda user.

Could you elaborate a bit on why you think this is the case? I agree it's a little more verbose than is ideal, but it's unclear to me why it would be problematic for a new user to understand? Maybe I'm missing something...

devmotion: > The other small issue is what to do about kernels that themselves have parameters e.g. rational quadratic? Would we assume that, if there's more than one argument, everything but the last should be passed directly to the kernel constructor? One of the reasons I quite like the function-based approach is because it's quite clear what happens in this case:
>
> kernel(RationalQuadraticKernel(2.0); l=l, s=s)

An alternative would be to always require the length scale to be provided as keyword argument, i.e.,
```julia
@kernel s * RationalQuadraticKernel(2.0; l=l)
```
would become
```julia
stretch(RationalQuadraticKernel(2.0); l=l, s=s)
```
That's slightly more verbose for cases such as
```julia
@kernel Matern52(; l=l)
```
but would at least make it completely clear that the provided `l` specifies the length scale (maybe one should even use the more verbose `lengthscale` instead of `l`?).

theogf: It just feels weird to have to call a `kernel` function on what is supposed to be already a kernel.
Somehow a `@kernel` macro feels to me to have a more generative meaning (I know it does not necessarily make sense)
I checked what famous Python packages (GPFlow, GPytorch etc) do, but they all have their transformations contained in their kernels. No inspiration to take from there.
Here are other propositions : 
- Remove the `Kernel` suffix on all `BaseKernel` (`SqExponentialKernel` -> `SqExponential`) types and use the suffix as the general call to create transformed kernel (`SqExponentialKernel(3.0)`->`TransformedKernel(SqExponential(),3.0)`)
- Do the opposite, keep all kernels this way and use their name without suffix to make `TransformedKernel`

I know that the issue with this is that it is not compatible with custom kernel. But we could also just have both... Have a `kernel` function/macro for general use. And have specific constructors for the `BaseKernel` we define

willtebbutt: > It just feels weird to have to call a kernel function on what is supposed to be already a kernel.

Hmmm yeah -- it also feels weird to me! Maybe there's just a better name we could use? Possibly something really generic like
```julia
transform(Matern52(); length_scale=l, variance=s)
```

> Somehow a @kernel macro feels to me to have a more generative meaning (I know it does not necessarily make sense)

I also agree with this.

>  But we could also just have both... Have a kernel function/macro for general use. And have specific constructors for the BaseKernel we define

Also a good point. Still seems like a shame to have to have more than one way of doing things though -- makes for a less straightforward user-experience imho, as there's more stuff for them to learn. Can also see the value of convenience though, just not convinced that the amount of convenience obtained is worth it.

theogf: To me it is important to have both because we can have two type of users:
- Users who just want to use an isotropic RBF kernel and do not want to read the whole docs to understand how to do it
- Users who want to create complex kernels combinations, eventually creating their own, and will be ready to read and understand docs to go further
A bit in the same way that Flux.jl allows to create a `Dense` layer as well as creating your own custom complex layers.
The `@kernel` approach seems good for creating such complex kernels, while having a `SqExponential(3.0)` seems the best way to have the fastest (and most intuitive) way to get a simple kernel

willtebbutt: > Users who just want to use an isotropic RBF kernel and do not want to read the whole docs to understand how to do it

I completely agree that a new user should definitely _not_ have to read all of the documentation to know how to construct a simple isotropric RBF kernel, but I don't agree that only providing the `@kernel` interface necessitates this. The argument being that

1. it's reasonable to expect users to read the most basic tutorial / whatever is on the package readme, so
2. if you have the `@kernel` example used consistently throughout the readme, then it is clear that this is the correct way to construct a kernel with this length scale.
3. You can make a point in the readme of saying that you shouldn't expect `KernelFunctions.jl` to have the same interface as other packages, and that this is okay.

theogf: Fair enough, you got me convinced.
So now we need to find the correct syntax for `@kernel`.
If we come to propose example I propose to use the `LinearKernel` as it requires a parameter `c`
I am not convinced by `l=` or `lengthscale=` because it does not generalize well to use different type of `Transform`
My favorite option would be to go with
```@kernel 2.0*Linear(3.0,c=2.0)```

Comment (willtebbutt) [reply src/generic.jl]: Are you proposing the general pattern
```julia
@kernel s * Kernel(transform, kernel_params...)
```
?

theogf: Exactly!
We can even extend it to more general formulas like:
```
@kernel s1* Kernel1(transform1, params1...) + s2*Kernel2(transform2,params2...)
```

theogf: So I will remove the lowercase constructors and finish the PR and maybe someone can make a PR for the `@kernel` macro, I am not good enough with macro to make it I think...

devmotion: What should
```julia
@kernel s * Kernel(transform, kernel_params...)
```
be rewritten to? To
```julia
s * TransformedKernel(Kernel(kernel_params...), transform)
```
? I'm still wondering if this does not lead to problems in cases such as
```julia
@kernel s * MaternKernel(42)
```
It seems a bit unintuitive to assume that a user would actually want to use a length scale of `42` in this case instead of the default constructor of a `MaternKernel` with parameter `42`. IMO in that example an explicit keyword argument such as `lengthscale` would be less confusing.

theogf: >What should
`@kernel s * Kernel(transform, kernel_params...)` be rewritten to? To `
s * TransformedKernel(Kernel(kernel_params...), transform)` ?

Yes exactly

> It seems a bit unintuitive to assume that a user would actually want to use a length scale of 42 in this case instead of the default constructor of a MaternKernel with parameter 42

Hmm I dont understand your point, it seems as unintuitive to me a user would use a nu = 42 for MaternKernel. That's why I think using keywords for kernel parameters is necessary. It is even more evident for kernel having multiple parameters like `PolynomialKernel`. What should be `PolynomialKernel(2.0,3.0)` should be ? Whereas `PolynomialKernel(c=2.0,d=3.0)` does not leave any doubt.
Having `lengthscale` as keyword could be an additional way to do it. But I would be quite against having kernel parameters as positional parameters, (as it is now the case)

devmotion: > Hmm I dont understand your point, it seems as unintuitive to me a user would use a nu = 42 for MaternKernel. That's why I think using keywords for kernel parameters is necessary.

Sure, using keyword arguments for kernel parameters might be more explicit and more intuitive, in particular in the case of kernels with multiple parameters. However, IMO it feels unintuitive and confusing if adding `@kernel` changes completely what the 42 in `MaternKernel(42)` means (even if one wants to encourage the use of keyword arguments, usually the default constructors still exist, so `MaternKernel(42)` might be a completely valid call of the default outer constructor).

> But I would be quite against having kernel parameters as positional parameters, (as it is now the case)

The problem I see is that you might be able to enforce that for the kernels defined in this package (by implementing an inner constructor that invalidates the default outer constructor and only adding an outer constructor with keyword arguments) but it is impossible to ensure that for user-defined kernels. Actually, I'm not even sure if it is a good idea to try to enforce that in the case of kernels with a single clearly defined parameter (such as maybe even `MaternKernel`).

theogf: > usually the default constructors still exist, so MaternKernel(42) might be a completely valid call of the default outer constructor

My thought was to change the default constructor as well.

>  it is impossible to ensure that for user-defined kernels.

We can be explicit about that in the docs

theogf: I think I will finish the PR without the `@kernel` wrapper and remove all other wrappers (the lowercase thing). We can add `@kernel` in a different PR. I will just correct the docs and add tests and merge the PR.

PR #36, opened 2020-02-21
Implemented Kronecker matrices with Kronecker.jl

theogf:
Added Requires to avoid overloading


Project.toml
```diff
@@ -5,8 +5,10 @@ version = "0.2.4"
 [deps]
 Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
 Distances = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
+Kronecker = "2c470bb0-bcc8-11e8-3dad-c9649493f05e"
```
devmotion:
```suggestion
```
If it is an optional dependency, it shouldn't be included among the dependencies. In the same way, I guess you want to remove PDMats as a dependency.

theogf: Oh I am not super familiar with `Requires.jl`. Does it mean I should just add it for the tests?

devmotion: Yes, exactly.

PR #38, opened 2020-02-28
@kernel macro for creating kernel

theogf:
So I started creating a first version of the `@kernel` macro which for now follows this syntax:
```julia
@kernel scale*Kernel(keywords arguments of the kernel) <keywordargument/argument for a transform>
```
So the transform is given either by a keyword `l=2.0`, `l=[2.0,3.0]` etc or directly by a parameter
For the kernel parameters they are restricted to be in the Kernel constructor

Edit : Left to do :

- [x] Add docs
- [x] Add tests


willtebbutt:
Am busy this week. Will take a look early next week.

willtebbutt:
Unable to review at the minute as the `@kernel` macro gives the error:
```julia
julia> @kernel 5 * Matern32Kernel(l=5.0)
ERROR: LoadError: UndefVarError: kw not defined
Stacktrace:
 [1] macro expansion at ./show.jl:562 [inlined]
 [2] @kernel(::LineNumberNode, ::Module, ::Expr, ::Any) at /Users/willtebbutt/.julia/dev/KernelFunctions/src/kernels/kernel_macro.jl:8
 [3] @kernel(::LineNumberNode, ::Module, ::Expr) at /Users/willtebbutt/.julia/dev/KernelFunctions/src/kernels/kernel_macro.jl:7
in expression starting at REPL[12]:1
```

theogf:
Sorry about that! it should be fixed now

willtebbutt:
This seems good:
```julia
julia> (@kernel Matern32Kernel() l=[5.0, 2.0]) + (@kernel Matern52Kernel() l=[4.0, 3.0])
Sum of 2 kernels:
	- (w=1.0) Matern32Kernel
			- ARDTransform{Float64,2}([5.0, 2.0])
	- (w=1.0) Matern52Kernel
			- ARDTransform{Float64,2}([4.0, 3.0])
```
(the printing is already quite cool)

This I found surprising:
```julia
julia> @kernel (Matern32Kernel() + Matern52Kernel()) l=[5.0, 4.0]
ERROR: MethodError: no method matching transform(::KernelSum, ::Array{Float64,1})
Closest candidates are:
  transform(::BaseKernel, ::AbstractArray{T,1} where T) at /Users/willtebbutt/.julia/dev/KernelFunctions/src/kernels/transformedkernel.jl:22
Stacktrace:
 [1] top-level scope at REPL[26]:1
```
Maybe it's just the case that `transform`'s type constraints need to be relaxed to `Kernel` from `BaseKernel`?

Also, this doesn't seem right:
```julia
julia> @kernel Matern32Kernel() l=5.0
Matern32Kernel
	- Scale Transform s=5.0
```

theogf:
> (the printing is already quite cool)
> 
> This I found surprising:
> 
> ```julia
> julia> @kernel (Matern32Kernel() + Matern52Kernel()) l=[5.0, 4.0]
> ERROR: MethodError: no method matching transform(::KernelSum, ::Array{Float64,1})
> Closest candidates are:
>   transform(::BaseKernel, ::AbstractArray{T,1} where T) at /Users/willtebbutt/.julia/dev/KernelFunctions/src/kernels/transformedkernel.jl:22
> Stacktrace:
>  [1] top-level scope at REPL[26]:1
> ```
> 
> Maybe it's just the case that `transform`'s type constraints need to be relaxed to `Kernel` from `BaseKernel`?

Yeah somehow I got fixed in the idea that `Transform` could only be applied on a `BaseKernel` I will extend the `transform` to all kernels.

> Also, this doesn't seem right:
> 
> ```julia
> julia> @kernel Matern32Kernel() l=5.0
> Matern32Kernel
> 	- Scale Transform s=5.0
> ```

I don't understand why? Should we start consider the parameter `l` as the actual lengthscale ? (`1/l`)

willtebbutt:
> I don't understand why? Should we start consider the parameter l as the actual lengthscale ? (1/l)

Sorry, no, I just misunderstood what is going on. I think it might be better to modify how this is printed a little bit, but that can be addressed in a separate issue. Please ignore my third point :)

docs/src/userguide.md
```diff
@@ -4,18 +4,24 @@
...
+
 ```julia
   k = SqExponentialKernel()
 ```
+
+Instead of having lengthscale(s) for each kernel we use `Transform` objects (see [Transform](@ref)). The transform operations are going to be applied on the inputs before they are passed to the kernel.
+For example, the [ScaleTransform](@ref) multiply every sample by a scalar $\rho$. A `SqExponentialKernel` with a `ScaleTransform(ρ)`, is therefore equivalent to have a `SqExponentialKernel` with lengthscale `1/ρ`.
+Here are some examples on how to use these transformations and are all equivalent:
```
devmotion, line 14:
```suggestion
Instead of having lengthscale(s) for each kernel we use `Transform` objects (see [`Transform`](@ref)). The transformations are applied to the inputs before the kernel is evaluated.
For example, the [`ScaleTransform`](@ref) multiplies every sample with a scalar. A `SqExponentialKernel` with a `ScaleTransform(ρ)` is therefore equivalent to a `SqExponentialKernel` with lengthscale `1/ρ`.
Here are some examples of how to use these transformations that are all equivalent:
```

src/kernels/kernel_macro.jl
```diff
@@ -0,0 +1,39 @@
+"""
+    @kernel [variance *]kernel::Kernel [l=Real/Vector / t=transform::Transform / transform::Transform]
```
devmotion, line 2: IMO this looks a bit confusing, in particular with all the alternatives. Maybe remove the type annotations and keep the signature as simple as possible?

src/kernels/kernel_macro.jl
```diff
@@ -0,0 +1,39 @@
...
+
+The `@kernel` macro is an helping alias to the [`transform`](@ref) function.
+The first argument should be a kernel multiplied (or not) by a scalar (variance of the kernel).
+The second argument (optional) can be a keyword :
+ - `l=ρ` where `ρ` is a positive scalar or a vector of scalar
+ - `t=transform` where `transform` is a [`Transform`](@ref) object
+One can also directly use a `Transform` object without a keyword.
+Here are some examples :
+```julia
```
devmotion, line 11: It would be nice to put this in a `# Examples` section and use "jldoctest" (as suggested in the [documentation](https://docs.julialang.org/en/latest/manual/documentation/)).

src/kernels/kernel_macro.jl
```diff
@@ -0,0 +1,39 @@
...
+            throw(error("The additional argument could not be intepreted. Please see documentation of `@kernel`"))
+        end
+    else
+        arg
+    end
+    if isnothing(scale)
+        return esc(:(transform($k, $t)))
+    else
+        return esc(:($scale * transform($k, $t)))
+    end
```
devmotion, line 38: Maybe just (untested, so might have to be changed slightly...)
```suggestion
    t = @capture(arg, (l = val_ | t = val_)) ? val : arg
    if scale === nothing
        return :(transform($(esc(k)), $(esc(t))))
    else
        return :($(esc(scale)) * transform($(esc(k)), $(esc(t))))
    end
```
Usually it is preferable to test equality with `nothing` with `===` (even though it shouldn't make a difference here) and to escape only user-provided input in macro expressions. Otherwise one has to qualify everything else (here, e.g., `transform`) to avoid problems if a user defined a variable (or function) `transform`.

src/kernels/transformedkernel.jl
```diff
@@ -41,11 +44,13 @@ _scale(t::ScaleTransform, metric, x, y) = evaluate(metric, t(x), t(y))
...
-transform(k::BaseKernel, t::Transform) = TransformedKernel(k, t)
+transform(k::Kernel, t::Transform) = TransformedKernel(k, t)
+
+transform(k::Kernel, ρ::Real) = TransformedKernel(k, ScaleTransform(ρ))
 
-transform(k::BaseKernel, ρ::Real) = TransformedKernel(k, ScaleTransform(ρ))
+transform(k::Kernel,ρ::AbstractVector) = TransformedKernel(k, ARDTransform(ρ))
 
-transform(k::BaseKernel,ρ::AbstractVector) = TransformedKernel(k, ARDTransform(ρ))
+transform(k::BaseKernel,::Nothing) = k
```
devmotion, line 53: Is there a reason for restricting this to `BaseKernel`s?
```suggestion
transform(k::Kernel, ::Nothing) = k
```

src/kernels/kernel_macro.jl
theogf: What do you think about leaving these small examples there (for understanding) but still creating a larger `# Examples` section with a big collection of varied examples?

src/kernels/kernel_macro.jl
theogf: In your code is there a way to still be able to throw an understandable error message?

src/kernels/transformedkernel.jl
theogf: Nope thanks for catching this

src/kernels/kernel_macro.jl
devmotion: I guess something like
```suggestion
    if arg === nothing
        t = nothing
    else
        if @capture(arg, (l = val_ | t = val_))
            t = val
        else
            error("the additional argument of `@kernel` is incorrect")
        end
    end

    if scale === nothing
        return :(transform($(esc(k)), $(esc(t))))
    else
        return :($(esc(scale)) * transform($(esc(k)), $(esc(t))))
    end
```
would be possible.

theogf: Btw `isnothing` is equivalent to `===`, in the docs of `isnothing` : 
"""
    isnothing(x)

Return `true` if `x === nothing`, and return `false` if not.
"""

devmotion: The result of the computation is the same but `===` is better for the compiler (see e.g. https://github.com/JuliaLang/julia/issues/35585#issuecomment-619053871).

theogf:
In terms in UI design, is it preferable to force to have a keyword, e.g. `@kernel SEKernel() l = 3.0` or is it okay to throw in any argument to make it an equivalent of `transform`, e.g. `@kernel SEKernel 3.0` ?

src/kernels/kernel_macro.jl
theogf: Oh I thought you were commenting on the docs page. Yes I will do this!

devmotion:
> In terms in UI design, is it preferable to force to have a keyword, e.g. @kernel SEKernel() l = 3.0 or is it okay to throw in any argument to make it an equivalent of transform, e.g. @kernel SEKernel 3.0 ?

Initially I thought `@kernel SEKernel() 3.0` would be too cryptic. However, I'm not so sure anymore since `l = ...` or `t = ...` is also not a completely intuitive syntax. That got me thinking if we need `@kernel` at all if one can just write `transform(SEKernel(), 3.0)` instead (but arguably the explicit syntax is completely user-friendly). At least it seems it's not needed to handle the scale with `@kernel` since the same syntax can be used outside of `@kernel` by writing `3 * @kernel SEKernel() l = 3.0` instead?

theogf:
This is true I can completely remove the scaling. If one put it inside of outside would not matter.
`@kernel` is indeed an alias for `transform`, but should be more intuitive. But there is no absolute best solution here. I tend to prefer `l=3.0` but I guess it is personal.

theogf:
Goodbye little one!

PR #39, opened 2020-03-02
Remove adjoints of Euclidean and SqEuclidean

devmotion:
Fixes https://github.com/theogf/KernelFunctions.jl/issues/37, now that Zygote 0.4.8 has been released.


theogf:
Thanks!

PR #40, opened 2020-03-03
Fix PDMats warning

devmotion:
This PR fixes the warning
```julia
┌ Warning: Error requiring PDMats from KernelFunctions:
│ LoadError: ArgumentError: Package KernelFunctions does not have PDMats in its dependencies:
│ - If you have KernelFunctions checked out for development and have
│   added PDMats as a dependency but haven't updated your primary
│   environment's manifest file, try `Pkg.resolve()`.
│ - Otherwise you may need to report an issue with KernelFunctions
```

### DIFF
--- a/src/matrix/kernelpdmat.jl
+++ b/src/matrix/kernelpdmat.jl
@@ -1,4 +1,4 @@
-using PDMats: PDMat
+using .PDMats: PDMat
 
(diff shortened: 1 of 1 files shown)


theogf:
Oh nice catch thanks!

PR #41, opened 2020-03-10
params(), Flux/Zygote style

theogf:
I replaced all scalars by vectors so they can be mutable, and copied the Flux approach of using `trainable` and `params` to get out an array of array of parameters to be optimized. Left to be done is to figure out how to fix parameters. @devmotion I saw you helped in making a `FixedKernel` in GaussianProcesses.jl do you think this is the right solution?

### DIFF
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -14,21 +14,7 @@ _scale(t::ScaleTransform, metric, x, y) = evaluate(metric, apply(t, x), apply(t,
 printshifted(io::IO,κ::Kernel,shift::Int) = print(io,"$κ")
 Base.show(io::IO,κ::Kernel) = print(io,nameof(typeof(κ)))
 
-function params(k::Kernel)
-    ps = []
-    params!(ps,k)
-    return ps
-end
-
-function params!(ps,k::Kernel)
-    for child in trainable(k)
-        params!(ps,k)
-    end
-end
-
-params!(ps,x::AbstractArray) = push!(ps,x)
-
-trainable(x) = ()
+_trainable(x) = ()
 
 ### Syntactic sugar for creating matrices and using kernel functions
 for k in subtypes(BaseKernel)
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -14,8 +14,6 @@ _scale(t::ScaleTransform, metric, x, y) = evaluate(metric, apply(t, x), apply(t,
 printshifted(io::IO,κ::Kernel,shift::Int) = print(io,"$κ")
 Base.show(io::IO,κ::Kernel) = print(io,nameof(typeof(κ)))
 
-_trainable(x) = ()
-
 ### Syntactic sugar for creating matrices and using kernel functions
 for k in subtypes(BaseKernel)
     @eval begin
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -14,6 +14,22 @@ _scale(t::ScaleTransform, metric, x, y) = evaluate(metric, apply(t, x), apply(t,
 printshifted(io::IO,κ::Kernel,shift::Int) = print(io,"$κ")
 Base.show(io::IO,κ::Kernel) = print(io,nameof(typeof(κ)))
 
+function params(k::Kernel)
+    ps = []
+    params!(ps,k)
+    return ps
+end
+
+function params!(ps,k::Kernel)
+    for child in trainable(k)
+        params!(ps,k)
+    end
+end
+
+params!(ps,x::AbstractArray) = push!(ps,x)
+
+trainable(x) = ()
+
 ### Syntactic sugar for creating matrices and using kernel functions
 for k in subtypes(BaseKernel)
     @eval begin
--- a/src/trainable.jl
+++ b/src/trainable.jl
@@ -1,42 +1,39 @@
-using .Flux: trainable
-
-Flux.trainable(::Kernel) = () # By default no parameters are returned
-Flux.trainable(::Transform) = ()
+import .Flux.trainable
 
 ### Base Kernels
 
-Flux.trainable(k::ConstantKernel) = (k.c,)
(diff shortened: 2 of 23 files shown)


devmotion:
Hmm that was quite some time ago, so I don't remember anything about `FixedKernel`. I think the correct approach would be to follow [the instruction in the documentation of Flux](https://fluxml.ai/Flux.jl/stable/models/advanced/#Freezing-Layer-Parameters-1) if you want to fix specific parameters.

devmotion:
BTW quite disappointing that all parameters have to be vectors. I guess there's no way around that currently?

src/generic.jl
```diff
@@ -14,6 +14,22 @@ _scale(t::ScaleTransform, metric, x, y) = evaluate(metric, apply(t, x), apply(t,
...
+
+function params!(ps,k::Kernel)
+    for child in trainable(k)
+        params!(ps,k)
+    end
+end
+
+params!(ps,x::AbstractArray) = push!(ps,x)
+
+trainable(x) = ()
```
devmotion:
```suggestion
```
IMO the implementation of `trainable` should be removed, since it is already defined in https://github.com/FluxML/Flux.jl/blob/a874bef6f9d8eb1ab2fe376cb8ad068eb36baf33/src/functor.jl#L40.

src/generic.jl
```diff
@@ -14,6 +14,22 @@ _scale(t::ScaleTransform, metric, x, y) = evaluate(metric, apply(t, x), apply(t,
...
+    return ps
+end
+
+function params!(ps,k::Kernel)
+    for child in trainable(k)
+        params!(ps,k)
+    end
+end
+
+params!(ps,x::AbstractArray) = push!(ps,x)
```
devmotion:
```suggestion
```
Isn't that already defined in https://github.com/FluxML/Flux.jl/blob/a874bef6f9d8eb1ab2fe376cb8ad068eb36baf33/src/functor.jl#L76-L88? It should be sufficient to define `trainable` for kernels that have trainable parameters.

src/utils.jl
```diff
@@ -38,4 +38,4 @@ base_transform(t::Transform) = eval(nameof(typeof(t)))
 For a kernel return a tuple with parameters of the transform followed by the specific parameters of the kernel
 For a transform return its parameters, for a `ChainTransform` return a vector of `params(t)`.
 """
-params(k::Kernel) = (params(transform(k)),)
+params
```
devmotion: Is it needed to have a custom `params` implementation? Couldn't one just use the one in Flux?

theogf:
> BTW quite disappointing that all parameters have to be vectors. I guess there's no way around that currently?

I don't think so, `Ref` would be definitely a better option. But maybe at some point Flux.jl will support it and we can revert it.

theogf:
I reimplemented `Flux` params to avoid the dependency, `Flux` being super heavy...
Do you have an alternative solution?

devmotion:
Ah I see. Are `params` and `trainable` supposed to be used without Flux at all? If not, one could make Flux an optional dependency and define them only if Flux loaded using Requires.

theogf:
> Ah I see. Are `params` and `trainable` supposed to be used without Flux at all? If not, one could make Flux an optional dependency and define them only if Flux loaded using Requires.

Not necessarily. The problem with `Requires` is that the need to move all the trainable in a different file. It actually feels weird that it is not possible to do all this without importing `Flux.jl` that would be great that they move this in a different package.

theogf:
Ah one easy option is to design an internal function `_trainable` for each kernel and with `Requires` to add simply `Flux.trainable(k::Kernel)= _trainable(k)`

devmotion:
If the whole purpose of `trainable` is the integration with Flux, IMO it would be more natural to only define `trainable`, possibly with `Requires`. Since all implementations are very short, I don't see an issue with having them in a separate file.

theogf:
Ok then that's what I did. I just need to write tests now.

src/trainable.jl
```diff
@@ -0,0 +1,42 @@
+using .Flux: trainable
+
+Flux.trainable(::Kernel) = () # By default no parameters are returned
+Flux.trainable(::Transform) = ()
```
devmotion: That's the default in Flux, so it's not really needed.

src/trainable.jl
```diff
@@ -0,0 +1,42 @@
+using .Flux: trainable
```
devmotion: This is also not needed, it seems?

src/trainable.jl
theogf: Yep good point

src/trainable.jl
theogf: Isn't it good practice to show which function you are going to dispatch on?

devmotion: Isn't that indicated quite clearly by writing `Flux.trainable(k::ExponentialKernel) = ...` etc.? I thought the line here could just be replaced with
```suggestion
import .Flux
```

src/generic.jl
```diff
@@ -14,6 +14,8 @@ _scale(t::ScaleTransform, metric, x, y) = evaluate(metric, apply(t, x), apply(t,
 printshifted(io::IO,κ::Kernel,shift::Int) = print(io,"$κ")
 Base.show(io::IO,κ::Kernel) = print(io,nameof(typeof(κ)))
 
+_trainable(x) = ()
+
```
devmotion:
```suggestion
```

src/transform/scaletransform.jl
```diff
@@ -16,7 +16,7 @@ function ScaleTransform(s::T=1.0) where {T<:Real}
 end
 
 set!(t::ScaleTransform,ρ::Real) = t.s .= [ρ]
-params(t::ScaleTransform) = t.s
+_trainable(t::ScaleTransform) = (t.s,)
```
devmotion:
```suggestion
```

PR #42, opened 2020-03-11
Fix typo

willtebbutt:


willtebbutt:
This actually might require renaming the internal parameter of `ScaledKernel` to avoid weird naming inconsistencies, but this is definitely the right idea. The parameter of the scaled kernel is its variance, as opposed to std. dev.

willtebbutt:
> That sounds like a reasonable change. I will then do all the renaming of ScaledKernel

@theogf no need, I've just done it!

theogf:
Just saw that. Maybe that would be good to also add a `variance` function that returns sigma for a ScaledKernel and 1 for other kernels? It just might end up complicated for nested kernels...

willtebbutt:
> Just saw that. Maybe that would be good to also add a variance function that returns sigma for a ScaledKernel and 1 for other kernels? It just might end up complicated for nested kernels...

This sounds reasonable. Could we do it in a different PR though please?

Issue #44, opened 2020-03-12
Add missing kernels from GPML


sharanry:

theogf:
Do you have a list somewhere?

sharanry:
From a quick look, these seem to be the missing kernels


- covCos
- covDiscrete
- covEye
- covFBM
- covGabor
- covGaborard
- covGaboriso
- covLINard
- covLINiso
- covLINone
- covMaha
- covMask
- covMaternard
- covMaterniso
- covNNone
- covOU
- covPER
- covPERard
- covPERiso
- covPP
- covPPard
- covPPiso
- covPeriodic
- covPeriodicNoDC
- covPref
- covRQard
- covRQiso
- covSEard
- covSEiso
- covSEisoU
- covSEproj
- covSEvlen
- covSM
- covULL
- covW
- covWarp

sharanry:
@theogf @willtebbutt Do you suggest I make separate PRs for each kernel or just one for all of them combined?

theogf:
A lot of this kernel are actually already implemented `covSEard` etc. I think having one PR per kernel seems fine. Maybe you can start by opening an issue for the ones missing so anyone can participate?

willtebbutt:
Yeah, you really just need to implement the "base" kernel -- the input transformations provided by `KernelFunctions` gives you the rest for free (see the `@kernel` macro in #38 and more generally the various `Transform`s).

I think maybe `covSEvlen` might need a special implementation, but the various `---ard`, `---iso`, `---proj` etc should be fine.

I agree that we don't have the `Cosine`, `Gabor`, `Eye`, `FBM`, `Maha`, `NN`, `PP`, `Pref`, `SM`, `ULL`, or `W` (wiener process). (I'm not even sure what some of these are lol)

There are also more recent kernels that GPML doesn't have, but we might want to implement. I'm thinking the various DNN kernels that have cropped up in the literature in recent years (I can't remember their names off the top of my head though)

theogf:
Haha same for me it's the first time I hear about some of those.
I will try to write some documentation on how to make custom kernels. But you can already have a look at the beginning of  https://github.com/theogf/KernelFunctions.jl/blob/master/test/test_custom.jl

sharanry:
Yeah. The function names aren't the most helpful. I have added a short description to them. Hope that helps.

willtebbutt:
[...]

PR #48, opened 2020-03-17
Add Fractional Brownian motion kernel

sharanry:
Issue #44

I am trying to implement [Fractional Brownian motion covariance ](https://github.com/alshedivat/gpml/blob/master/cov/covFBM.m) given by
```
    κ(x,y) =  ( |x|²ʰ + |z|²ʰ - |x-z|²ʰ ) / 2
``` 
 with Hurst index `h` from (0,1).

I don't think the current structure of `kappa` and `metric` can be used directly implement this. I am considering implementing a modified `kappa` function to serve such cases. Any thoughts on this?


sharanry:
@willtebbutt @theogf @devmotion

willtebbutt:
Yeah -- looks like it will be tricky.

The best way to go about implementing this will be to implement `kernelmatrix` and `kerneldiagmatrix` directly. The `kappa` and `metric` abstraction is just a convenient way of doing that anyway. See eg. the `KernelSum` implementation for an example of how this would work.

src/kernels/fbm.jl
```diff
@@ -37,6 +37,16 @@ function kernelmatrix(
...
     modXY = pairwise(SqEuclidean(),X-Y,dims=obsdim)
-    @info modXY
     K = map((x)->_fbm(x[1], x[2], x[3], κ.h), zip(modX, modY, modXY))
 end
+
+#Syntactic Sugar
+function (κ::FBMKernel)(x::AbstractVector{<:Real}, y::AbstractVector{<:Real})
+    modX = evaluate(SqEuclidean(), x, x)
+    modY = evaluate(SqEuclidean(), y, y)
+    modXY = evaluate(SqEuclidean(), x-y, x-y)
```
devmotion: This seems wrong, all distances here should evaluate to zero?

src/kernels/fbm.jl
```diff
@@ -0,0 +1,52 @@
...
+"""
+struct FBMKernel{T<:Real} <: BaseKernel
+    h::T
+    function FBMKernel(;h::T=0.5) where {T<:Real}
+        @assert h<=1.0 && h>=0.0 "FBMKernel: Given Hurst index h is invalid."
+        new{T}(h)
+    end
+end
+
+kappa(κ::FBMKernel, d::Real) = error("Not Implemented: Please use `kernelmatrix` or `kerneldiagmatrix` instead.")
```
devmotion: IMO there's no need to throw an error, I'd suggest just not implementing `kappa` in this case (similar to `metric` just not being defined).

Moreover, I think the error message is misleading since IMO it should also be possible to evaluate the kernel with scalar- and vector-valued arguments, two cases that are not captured by `kernelmatrix` and `kerneldiagmatrix`. It seems vector-valued arguments can be evaluated using the implementation below in the section "syntactic sugar" but I think currently it's not possible to evaluate it for scalar-valued inputs? In general (not restricted to this PR), I'm wondering what's the desired API - do we want a user to use the "syntactic sugar" in these cases (which then actually wouldn't be syntactic sugar but somewhat mandatory part of the API)? Or do we expect them to use `kappa` with three arguments whose generic version is implemented in https://github.com/theogf/KernelFunctions.jl/blob/214b83fdbf785b4ca71c1b2d5e4238d796938cc6/src/generic.jl#L7?

src/kernels/fbm.jl
sharanry: Sorry, my bad. Does it make sense now?

src/kernels/fbm.jl
```diff
@@ -42,9 +42,9 @@ end
 
 #Syntactic Sugar
 function (κ::FBMKernel)(x::AbstractVector{<:Real}, y::AbstractVector{<:Real})
-    modX = evaluate(SqEuclidean(), x, x)
-    modY = evaluate(SqEuclidean(), y, y)
-    modXY = evaluate(SqEuclidean(), x-y, x-y)
+    modX = evaluate(SqEuclidean(), x, zero(x))
+    modY = evaluate(SqEuclidean(), y, zero(y))
+    modXY = evaluate(SqEuclidean(), x-y, zero(x-y))
```
devmotion: I'd suggest
```suggestion
    modX = sum(abs2, x)
    modY = sum(abs2, y)
    modXY = sqeuclidean(x, y)
```

src/kernels/fbm.jl
sharanry: I think the "Syntactic Sugar" covers the scalar-valued inputs in general. We should implement it in specific for this kernel.

sharanry: I am assuming the preferred usage to be what we provide using the "syntactic sugar". As it seems to be the most natural way to apply a kernel.

sharanry: I have added the scalar case.

src/kernels/fbm.jl
```diff
@@ -0,0 +1,51 @@
...
+    X::AbstractMatrix,
+    Y::AbstractMatrix;
+    obsdim::Int = defaultobs,
+)
+    @assert obsdim ∈ [1,2] "obsdim should be 1 or 2 (see docs of kernelmatrix))"
+    
+    modX = pairwise(SqEuclidean(),X,dims=obsdim)
+    modY = pairwise(SqEuclidean(),Y,dims=obsdim)
+    modXY = pairwise(SqEuclidean(),X-Y,dims=obsdim)
+    K = map((x)->_fbm(x[1], x[2], x[3], κ.h), zip(modX, modY, modXY))
```
devmotion: Shouldn't it rather be
```suggestion
    modX = sum(abs2, X, dims=3-obsdim)
    modY = sum(abs2, Y, dims=3-obsdim)
    modXY = pairwise(SqEuclidean(), X, Y,dims=obsdim)
    K = _fbm.(vec(modX), reshape(modY, 1, :), modXY, κ.h)
```
?

src/kernels/fbm.jl
```diff
@@ -0,0 +1,51 @@
...
+        @assert h<=1.0 && h>=0.0 "FBMKernel: Given Hurst index h is invalid."
+        new{T}(h)
+    end
+end
+
+_fbm(modX, modY, modXY, h) = (modX^h + modY^h - modXY^h)/2
+
+function kernelmatrix(κ::FBMKernel, X::AbstractMatrix; obsdim::Int = defaultobs)
+    @assert obsdim ∈ [1,2] "obsdim should be 1 or 2 (see docs of kernelmatrix))"
+    K = map(modX->_fbm(modX, modX, 0, κ.h), pairwise(SqEuclidean(),X,dims=obsdim))
```
devmotion: Shouldn't this rather be
```suggestion
    modX = sum(abs2, X; dims = 3 - obsdim)
    modXX = pairwise(SqEuclidean(), X, dims = obsdim)
    K = _fbm.(vec(modX), reshape(modX, 1, :), modXX, κ.h)
```
?

src/kernels/fbm.jl
sharanry: Yeah I think this makes sense.

src/kernels/fbm.jl
```diff
@@ -0,0 +1,53 @@
...
+#Syntactic Sugar
+function (κ::FBMKernel)(x::AbstractVector{<:Real}, y::AbstractVector{<:Real})
+    modX = sum(abs2, x)
+    modY = sum(abs2, y)
+    modXY = sqeuclidean(x, y)
+    (modX^κ.h + modY^κ.h - modXY^κ.h)/2
+end
+
+(κ::FBMKernel)(x::Real, y::Real) = (abs2(x)^κ.h + abs2(y)^κ.h - abs2(x-y)^κ.h)/2
+(κ::FBMKernel)(X::AbstractMatrix{T}, Y::AbstractMatrix{T}; obsdim::Integer=defaultobs) where {T} = kernelmatrix(κ, X, Y, obsdim=obsdim)
```
willtebbutt: This probably doesn't need the type constraints, and is better written as
```suggestion

function (κ::FBMKernel)(X::AbstractMatrix{<:Real}, Y::AbstractMatrix{<:Real}; obsdim::Integer=defaultobs)
    return kernelmatrix(κ, X, Y, obsdim=obsdim)
end
```
to avoid excessively long lines.

src/kernels/fbm.jl
```diff
@@ -0,0 +1,53 @@
...
+(κ::FBMKernel)(X::AbstractMatrix{T}; obsdim::Integer=defaultobs) where {T} = kernelmatrix(κ, X, obsdim=obsdim)
```
willtebbutt: Same here:
```suggestion

function (κ::FBMKernel)(X::AbstractMatrix{<:Real}; obsdim::Integer=defaultobs)
    return kernelmatrix(κ, X, obsdim=obsdim)
end
```

src/kernels/fbm.jl
```diff
@@ -0,0 +1,53 @@
...
+    modXY = pairwise(SqEuclidean(), X, Y,dims=obsdim)
+    return _fbm.(vec(modX), reshape(modY, 1, :), modXY, κ.h)
+end
+
+#Syntactic Sugar
+function (κ::FBMKernel)(x::AbstractVector{<:Real}, y::AbstractVector{<:Real})
+    modX = sum(abs2, x)
+    modY = sum(abs2, y)
+    modXY = sqeuclidean(x, y)
+    (modX^κ.h + modY^κ.h - modXY^κ.h)/2
```
willtebbutt:
```suggestion
    return (modX^κ.h + modY^κ.h - modXY^κ.h)/2
```

src/kernels/fbm.jl
```diff
@@ -0,0 +1,59 @@
+"""
+    FBMKernel(; h::Real=0.5)
+Fractional Brownian motion kernel with Hurst index h from (0,1) given by
```
willtebbutt:
```suggestion

Fractional Brownian motion kernel with Hurst index h from (0,1) given by
```

src/kernels/fbm.jl
```diff
@@ -0,0 +1,59 @@
...
+```
+    κ(x,y) =  ( |x|²ʰ + |y|²ʰ - |x-y|²ʰ ) / 2
+```
+
+For h=1/2, this is the Wiener Kernel, for h>1/2, the increments are
+positively correlated and for h<1/2 the increments are negatively correlated.
+%
```
willtebbutt:
```suggestion
```

src/kernels/fbm.jl
```diff
@@ -0,0 +1,59 @@
...
+"""
+struct FBMKernel{T<:Real} <: BaseKernel
+    h::T
+    function FBMKernel(;h::T=0.5) where {T<:Real}
+        @assert h<=1.0 && h>=0.0 "FBMKernel: Given Hurst index h is invalid."
+        new{T}(h)
```
willtebbutt:
```suggestion
        return new{T}(h)
```

test/test_kernels.jl
```diff
@@ -25,6 +25,19 @@ x = rand()*2; v1 = rand(3); v2 = rand(3); id = IdentityTransform()
             @test kappa(k,0.5) == c
         end
     end
+    @testset "FBM" begin
+        k = FBMKernel(h=0.3)
+        @test k(v1,v2) ≈ (sqeuclidean(v1, zero(v1))^0.3 + sqeuclidean(v2, zero(v2))^0.3 - sqeuclidean(v1-v2, zero(v1-v2))^0.3)/2 atol=1e-5
+        
+        # kernelmatrix tests
+        m1 = rand(3,3); m2 = rand(3,3);
```
willtebbutt:
```suggestion
        m1 = rand(3,3)
        m2 = rand(3,3)
```

test/test_kernels.jl
```diff
@@ -25,6 +25,19 @@ x = rand()*2; v1 = rand(3); v2 = rand(3); id = IdentityTransform()
...
+        @test kernelmatrix(k, m1, m1) ≈ kernelmatrix(k, m1)
+        @test kernelmatrix(k, m1, m2) ≈ k(m1, m2)
+
+        
+        x1 = rand(); x2 = rand()
```
willtebbutt:
```suggestion
        x1 = rand()
        x2 = rand()
```

sharanry:
@willtebbutt Can we merge this now?

devmotion:
I guess it would be good to implement the inplace variants `kernelmatrix!` and `kerneldiagmatrix!` as well.

src/kernels/fbm.jl
```diff
@@ -26,6 +26,10 @@ function kernelmatrix(κ::FBMKernel, X::AbstractMatrix; obsdim::Int = defaultobs
     return _fbm.(vec(modX), reshape(modX, 1, :), modXX, κ.h)
 end
 
+function kernelmatrix!(K::AbstractMatrix, κ::FBMKernel, X::AbstractMatrix; obsdim::Int = defaultobs)
+    K = kernelmatrix(κ, X; obsdim = obsdim)
```
devmotion: That's not correct, you're supposed to update the entries of `K` here. Currently you just create a new matrix which you also call `K`.

src/kernels/fbm.jl
```diff
@@ -40,6 +44,16 @@ function kernelmatrix(
...
 end
 
+function kernelmatrix!(
+    K::AbstractMatrix,
+    κ::FBMKernel,
+    X::AbstractMatrix,
+    Y::AbstractMatrix;
+    obsdim::Int = defaultobs,
+)
+    K = kernelmatrix(κ, X, Y; obsdim = obsdim)
```
devmotion: The same here, you have to update the entries of `K`.

src/kernels/fbm.jl
sharanry: Would this work?
 `K[:] = kernelmatrix(κ, X; obsdim = obsdim)`

devmotion: Maybe that's a good reason for also adding tests for the inplace variants BTW :smiley:

sharanry:
@theogf @devmotion 
What is the purpose of `kerneldiagmatrix` and `kerneldiagmatrix!`?
We currently have implementations in general for 
```julia
kerneldiagmatrix(κ::Kernel, X::Matrix; obsdim::Int=2)
kerneldiagmatrix!(K::AbstractVector,κ::Kernel, X::Matrix; obsdim::Int=2)
```
however we do not have implementations for
```julia
kerneldiagmatrix(κ::Kernel, X::Matrix,y::Matrix; obsdim::Int=2)
kerneldiagmatrix!(K::AbstractVector,κ::Kernel, X::Matrix,y::Matrix; obsdim::Int=2)
```
Any particular reason begin this?

willtebbutt:
> Any particular reason begin this?

This is definitely something that I want, because it's crucial for Stheno.jl (see the docs for the rationale). I don't think this PR is the right place for it though -- I would stick with implementing the current interface, and we can re-visit later.

sharanry:
@willtebbutt So are we going to implement `kerneldiagmatrix` and `kerneldiagmatrix!` for `FBMKernel` before merging or not?

theogf:
`kerneldiagmatrix(k, X)` aims at computing the diagonal only of `kernelmatrix(k, X)`. As @willtebbutt pointed out, this extremely useful for Gaussian Processes (and probably other stuff)

willtebbutt:
> @willtebbutt So are we going to implement kerneldiagmatrix and kerneldiagmatrix! for FBMKernel before merging or not?

Yes, but only the single-argument ones.

sharanry:
@theogf @willtebbutt I was confused because it always seems to return an array of ones. As it probably should. But I guess this is not the right place. I will discuss this over slack.
```julia
julia> kerneldiagmatrix(CosineKernel(), rand(3,3))
3-element Array{Float64,1}:
 1.0
 1.0
 1.0

julia> kerneldiagmatrix(SqExponentialKernel(), rand(3,3))
3-element Array{Float64,1}:
 1.0
 1.0
 1.0

julia> kerneldiagmatrix(ExponentialKernel(), rand(3,3))
3-element Array{Float64,1}:
 1.0
 1.0
 1.0

```

theogf:
Well it's because all these kernels are based on Euclidean distance and you will always get a distance of 0, hence all the ones! If you try with the LinearKernel, it won't be the case

sharanry:
@theogf Thanks a lot for that! Now it makes sense :)

src/kernels/fbm.jl
```diff
@@ -26,8 +26,8 @@ function kernelmatrix(κ::FBMKernel, X::AbstractMatrix; obsdim::Int = defaultobs
     return _fbm.(vec(modX), reshape(modX, 1, :), modXX, κ.h)
 end
 
-function kernelmatrix!(K::AbstractMatrix, κ::FBMKernel, X::AbstractMatrix; obsdim::Int = defaultobs)
-    K = kernelmatrix(κ, X; obsdim = obsdim)
+function kernelmatrix!(K::Matrix, κ::FBMKernel, X::AbstractMatrix; obsdim::Int = defaultobs)
+    K[:,:] = kernelmatrix(κ, X; obsdim = obsdim)
```
devmotion: That's not correct either, unfortunately. We shouldn't restrict it to `Matrix` (e.g., then a view can't be passed), and additionally it should not be necessary to generate a matrix with the results and then copy them to `K`. Instead the results should end up in `K` immediately. One way to achieve this is copying the code from `kernelmatrix` and changing the final line to
```julia
    K .= _fbm.(vec(modX), reshape(modX, 1, :), modXX, κ.h)
    return K
```

src/kernels/fbm.jl
```diff
@@ -51,7 +51,7 @@ function kernelmatrix!(
     Y::AbstractMatrix;
     obsdim::Int = defaultobs,
 )
-    K = kernelmatrix(κ, X, Y; obsdim = obsdim)
+    K[:,:] = kernelmatrix(κ, X, Y; obsdim = obsdim)
```
devmotion: Same here.

src/kernels/fbm.jl
sharanry: The reason I had to go with `Matrix` instead of `AbstractMatrix` is that current [implementation](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/0654f2fd5bceed8600873158c251629c9ea4ed5f/src/matrix/kernelmatrix.jl#L12) uses it. I was facing ambiguity errors when I went with   `AbstractMatrix`. 

Is the current implementation in master a mistake too? I wasn't sure.

willtebbutt: Yeah, that looks like a mistake. I would go with `AbstractMatrix`.

sharanry:
@willtebbutt Please take a look. Implementing just `_kernel` specific to `FBMKernel` seems to suffice.

willtebbutt:
Actually, sorry, I've just thought: do we actually need to implement `_kernel` to make `kerneldiagmatrix` work? Looks like the `_kernel` implementations near [here](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/1081e46399a3a3682b6f121fd7317a4d79a55f5b/src/matrix/kernelmatrix.jl#L45) should do the trick.

sharanry:
@willtebbutt Yeah but that uses `kappa` which we don't have.

PR #49, opened 2020-03-17
Add non-unicode keywords aliases

theogf:
Add alpha vs \alpha, gamma vs \gamma etc, for the concerned kernels.


devmotion:
IMO in general it's preferable to have only one way for specifying the kernel parameters. Maybe only allow non-unicode keyword arguments?

theogf:
Do you really think it would be bad? There is no chance some one is going to have a `alpha` and `\alpha` keyword at the same time

devmotion:
Sure, it's unlikely that someone will specify both keyword arguments. I just think that generally it can be confusing, both for developers and users, if there are multiple different ways for achieving or specifying the same thing.

Issue #34, opened 2020-02-02
Add periodic kernels


IsakFalk:
I was playing around with periodic data and noticed that the periodic kernel, a general form can be found in [this document, equation 47](http://www.inference.org.uk/mackay/gpB.pdf), isn't implemented.

I think you could use a transformation together with one of the existing kernels that yields this kernel, but it would be nice to have easy access to it.

theogf:
Another possibility is to define a new metric `Periodic` and leave the `r_i` as parameters of the kernel.

PR #50, opened 2020-03-17
Add Periodic Kernel

theogf:
Following the request #34 I implemented a first version of the periodic kernel described in http://www.inference.org.uk/mackay/gpB.pdf equation 47.
I also added the unconventional Sinus metric to keep the same style as other kernels.


src/distances/sinus.jl
```diff
@@ -0,0 +1,13 @@
+struct Sinus{T} <: Distances.PreMetric
```
devmotion: By using
```suggestion
struct Sinus{T} <: Distances.SemiMetric
```
Distances exploits the symmetry when calling `pairwise`.

src/distances/sinus.jl
```diff
@@ -0,0 +1,13 @@
...
+    r::Vector{T}
+end
+
+@inline function Distances._evaluate(d::Sinus,a::AbstractVector{T},b::AbstractVector{T}) where {T}
```
devmotion: In general, I think one should rather define `Distances.parameters` and `Distances.eval_op` (I guess the default definitions of `eval_start`, `eval_reduce`, and `eval_end` should be fine). A Zygote-compatible implementation could then be added in the definition of the custom adjoint.
Here's an example how these functions are used when evaluating the distance between two arrays:
https://github.com/JuliaStats/Distances.jl/blob/master/src/metrics.jl#L181-L210
Using these definitions, one could even just define
```julia
(dist::Sinus)(a::Number, b::Number) = Distances.eval_op(dist, a, b, first(dist.r))
```
similar to https://github.com/JuliaStats/Distances.jl/blob/7f3a28c0d1372e3b3edbcbc28f00ba5645e1bbdb/src/metrics.jl#L267.

src/kernels/periodic.jl
```diff
@@ -0,0 +1,15 @@
+"""
+PeriodicKernel()
```
devmotion: According to the [Julia documentation](https://docs.julialang.org/en/latest/manual/documentation/), one should write this as
```suggestion
    PeriodicKernel

```

> so that it is printed as Julia code.

Alternatively, of course, one could specify the function signature of the constructor, but I guess that would involve either `dims` or `r`?

src/distances/sinus.jl
theogf: Thanks I will also use it for `DotProduct` then

src/distances/sinus.jl
theogf: Thanks I implemented `eval_op` only now

theogf: Ah but this would not be taken into account, since `UnionMetrics` is fixed : https://github.com/JuliaStats/Distances.jl/blob/7f3a28c0d1372e3b3edbcbc28f00ba5645e1bbdb/src/metrics.jl#L107

devmotion: Ah sorry I completely missed that :see_no_evil: Somehow I assumed https://github.com/JuliaStats/Distances.jl/issues/126 was fixed already.

src/kernels/periodic.jl
theogf: Corrected!

src/distances/sinus.jl
theogf: I will leave it as it is for now

dkarrasch: The "`UnionMetrics` as a `Union`" issue is now resolved on master of `Distances.jl`, so @devmotion's suggestion should now be doable. There is a similar minimal example in https://github.com/JuliaStats/Distances.jl/issues/151#issuecomment-612460855.

theogf: @dkarrasch Thanks! We need to wait for a new release of Distances.jl now...

theogf:
I could not put the `eval_op` approach as there is no release of `Distances` for it. I put it as commented and will leave it for another PR (solving DotProduct at the same time)

dkarrasch:
Not sure where to post it, so I'll do it here. Distances.jl has released a new version, that allows to subtype `Union*Metric`s. If you do that, however, you'll need a lower compat bound for Distances.jl "0.9".

devmotion:
I've already added a lower bound for 0.9 (https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/29b5ad793312b3127fb714a96940898730d9dfc6/Project.toml#L18) to get rid of the SqEuclidean hacks (was needed to avoid the negative results).

PR #51, opened 2020-03-20
Add alias to WhiteKernel, EyeKernel

sharanry:
Issue #44

From what I gather, [EyeKernel](https://github.com/alshedivat/gpml/blob/master/cov/covEye.m) is the same as already implemented `WhiteKernel`. Simply adding an alias.

### DIFF
--- a/src/kernels/constant.jl
+++ b/src/kernels/constant.jl
@@ -24,6 +24,7 @@ Kernel function working as an equivalent to add white noise.
 struct WhiteKernel <: BaseKernel end
 
 const EyeKernel = WhiteKernel
+export EyeKernel
 
 kappa(κ::WhiteKernel,δₓₓ::Real) = δₓₓ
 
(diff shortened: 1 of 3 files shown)


theogf:
Sounds good. Could you just add a test? And also add it in export

src/kernels/constant.jl
```diff
@@ -23,6 +23,9 @@ Kernel function working as an equivalent to add white noise.
 """
 struct WhiteKernel <: BaseKernel end
 
+const EyeKernel = WhiteKernel
+export EyeKernel
```
devmotion: Can you add it to the exports in https://github.com/theogf/KernelFunctions.jl/blob/master/src/KernelFunctions.jl instead? In that way it's a lot easier to spot what is exported by KernelFunctions.

sharanry: @devmotion Sure. Didn't notice those exports at the top. Thanks for pointing it out!

sharanry:
@theogf I think we can merge this if everything is fine?

theogf:
Yep!

PR #52, opened 2020-03-20
Add Gabor Kernel

sharanry:
Issue #44 

Gabor kernel with length scale ell and period p. Given by
```
    κ(x,y) =  h(x-z), h(t) = exp(-sum(t.^2./(2*ell.^2)))*cos(2*pi*sum(t./p))
```

I don't think we can get rid of period and length scale using `ScaledKernel` like we did in #45. Please let me know if I am wrong or if there any other way.

References:
- http://www.gaussianprocess.org/gpml/code/matlab/doc/manual.pdf pg44
- https://github.com/alshedivat/gpml/blob/master/cov/covGabor.m


devmotion:
Actually, I think we do not need the length scale and could only keep the period, since if `d` is rescaled to `d/ell` one could just use `p/ell` to obtain the same results, it seems. That would be nice in a way since it would allow to reuse the existing input transformations in KernelFunctions.

src/kernels/gabor.jl
```diff
@@ -0,0 +1,20 @@
+"""
+    GaborKernel(; ell::Real=1.0, p::Real=1.0)
+
+Gabor kernel with length scale ell and period p. Given by
+```
```
devmotion: Just a minor thing since it is a formula:
```suggestion
```math
```

theogf:
Can't we just define it as a product of a Squared exponential and cosine actually?

sharanry:
@theogf I am not exactly sure to how to accommodate the the different scaling for inputs of `SqExponentialKernel` and `CosineKernel`. Did you have something in mind?

theogf:
I think this would look like `transform(SqExponential(), ell)*transform(CosineKernel(), p)`

sharanry:
@theogf Shouldn't it be this?
`transform(SqExponentialKernel(), sqrt(2)*ell)*transform(CosineKernel(), p)`

theogf:
`transform(SqExponentialKernel(), 1/(sqrt(2)*ell))*transform(CosineKernel(), 1/p)`. The scale transform multiply by the inputs

devmotion:
I guess the `sqrt(2)` is not necessarily needed here. The division by 2 is also used in the squared exponential kernel in GPML whereas in KernelFunctions the squared exponential kernel is defined without scaling. So without `sqrt(2)` it would be consistent with our definition of `SqExponentialKernel` whereas with `sqrt(2)` it would be according to the definition in GPML.

devmotion:
I'm wondering, if we actually need a separate type for the Gabor kernel since it can be composed of `SqExponentialKernel` and `CosineKernel`. We could even try to remove the use of unneeded transforms with something like
```julia
function GaborKernel(; ell = nothing, p = nothing)
    if ell === nothing
        if p === nothing
            return SqExponentialKernel() * CosineKernel()
        else
            return SqExponentialKernel() * transform(CosineKernel(), 1 ./ p)
        end
    elseif p === nothing
        return transform(SqExponentialKernel(), 1 ./ ell) * CosineKernel()
    else
        return transform(SqExponentialKernel(), 1 ./ ell) * transform(CosineKernel(), 1 ./ p)
    end
end
```

sharanry:
IMO it would be good to be able to use the `GaborKernel` out of box. 
Would removing the transforms make much of a difference in terms of performance?

devmotion:
> IMO it would be good to be able to use the GaborKernel out of box.

Isn't that "out of the box"? You just call `GaborKernel()` (possibly with length scale and period) and you get a Gabor kernel.

> transform(SqExponentialKernel(), 1 ./ ell) * transform(CosineKernel(), 1 ./ p)

I haven't benchmarked it but I would assume that most of the time (in particular for higher dimensional inputs) is spent in evaluating the kernels and not in computing the transforms. It just felt a bit unnecessary to perform these additional computations if it's so easy to avoid them.

sharanry:
@devmotion Sorry, I think I misinterpreted what you meant. 

If you mean to say that we construct just a function called `GaborKernel` which gives a `KernelProduct` object, I think I agree with you. This should be sufficient. However, we need to keep in mind that we can't access the parameters, `ell` and `p` easily at a later point if required for whatever reason.

theogf:
I see @sharanry point, we could eventually write the GaborKernel as a wrapper around what you described :
```julia
struct GaborKernel{K<:Kernel} <: Kernel
 k::Kernel
end
```
And create the constructor as @devmotion described it. One can then create `get` functions...

sharanry:
@theogf But would this also create problems if we intend to make the parameters learnable?

theogf:
No it would be fine as it would just be a wrapper, one would need to create the trainable function on `GaborKernel` and the training parameters would be propagated through

src/kernels/gabor.jl
```diff
@@ -11,9 +11,25 @@ struct GaborKernel{T<:Real, K<:Kernel} <: BaseKernel
     ell::T
     p::T
```
theogf: I would not even store `p` and `ell`. It makes more sense to just make an adaptive get function for each case

sharanry: Yeah makes sense. Will do that!

src/kernels/gabor.jl
```diff
@@ -33,6 +33,22 @@ function _gabor(; ell = nothing, p = nothing)
...
+    elseif v == :p && typeof(k.κ.kernels[2]) <: CosineKernel
+        return 1.0
+    elseif v == :p && typeof(k.κ.kernels[2]) <: TransformedKernel
+        return 1 ./ k.κ.kernels[2].transform.s[1]
+    else
+        error("Invalid Property")
+    end
+end
+
 kappa(κ::GaborKernel, x, y) where {T<:Real} = kappa(κ.κ, x ,y)
```
theogf: You don't need the `where {T}` here

src/kernels/gabor.jl
```diff
@@ -0,0 +1,74 @@
+import Base.getproperty
+
```
devmotion: This is not needed.
```suggestion
```

src/kernels/gabor.jl
```diff
@@ -0,0 +1,74 @@
...
+```
+
+"""
+struct GaborKernel{T<:Real, K<:Kernel} <: BaseKernel
+    κ::K
+    function GaborKernel(;ell=nothing, p=nothing)
+        k = _gabor(ell=ell, p=p)
+        if ell == nothing ell=1.0 end
+        if p == nothing p=1.0 end
+        new{Union{typeof(ell),typeof(p)}, typeof(k)}(k)
```
devmotion: It's not a good idea to define `T` is a union since that's an abstract type. One could use `promote_type` here, but I suggest removing `T` completely since it is not needed.
```suggestion
        k = _gabor(ell=ell, p=p)
        new{typeof(k)}(k)
```

devmotion: Of course, then `T` has to be removed in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/52/files#diff-d820aea21c04d68f343a510d8e894cf4R14.

src/kernels/gabor.jl
```diff
@@ -0,0 +1,74 @@
...
+    if v == :κ 
+        return getfield(k, v)
+    elseif v == :ell && typeof(k.κ.kernels[1]) <: SqExponentialKernel
+        return 1.0
+    elseif v == :ell && typeof(k.κ.kernels[1]) <: TransformedKernel
+        return 1 ./ k.κ.kernels[1].transform.s[1]
+    elseif v == :p && typeof(k.κ.kernels[2]) <: CosineKernel
+        return 1.0
+    elseif v == :p && typeof(k.κ.kernels[2]) <: TransformedKernel
+        return 1 ./ k.κ.kernels[2].transform.s[1]
```
devmotion: IMO it would be easier to read if you write
```suggestion
    elseif v == :ell
        kernel1 = k.κ.kernels[1]
    	if kernel1 isa TransformedKernel
    	    return 1 ./ kernel1.transform.s[1]
    	else
    	    return 1.0
    	end
    elseif v == :p
        kernel2 = k.κ.kernels[2]
        if kernel2 isa TransformedKernel
            return 1 ./ kernel2.transform.s[1]
        else
            return 1.0
        end
```

src/kernels/gabor.jl
```diff
@@ -0,0 +1,74 @@
...
+    GaborKernel(; ell::Real=1.0, p::Real=1.0)
+
+Gabor kernel with length scale ell and period p. Given by
+```math
+    κ(x,y) =  h(x-z), h(t) = exp(-sum(t.^2./(ell.^2)))*cos(pi*sum(t./p))
+```
+
+"""
+struct GaborKernel{T<:Real, K<:Kernel} <: BaseKernel
+    κ::K
```
devmotion: Maybe use something more descriptive such as
```suggestion
    kernel::K
```
?

sharanry:
@theogf @devmotion Can we merge this?

PR #54, opened 2020-03-23
Add Mahalanobis distance-based kernel

sharanry:
Issue #44 

Adding Mahalanobis distance-based kernel given by,

```math
    κ(x,y) =  kappa(r^2), r^2 = maha(x,P,y) = (x-y)'*inv(P)*(x-y)
```

Not sure about what the name should be. `MahaKernel` or `MahalanobisKernel`?

**References**
https://github.com/alshedivat/gpml/blob/master/cov/covMaha.m
www.gaussianprocess.org/gpml/code/matlab/doc/manual.pdf pg44

### DIFF
--- a/src/kernels/maha.jl
+++ b/src/kernels/maha.jl
@@ -1,22 +1,21 @@
 """
-    MahaKernel(P::AbstractMatrix, kappa::Function)
+    MahalanobisKernel(P::AbstractMatrix, kappa::Function)
 
     Mahalanobis distance-based kernel given by
 ```math
-    κ(x,y) =  kappa(r^2), r^2 = maha(x,P,y) = (x-y)'*inv(P)*(x-y)
+    κ(x,y) =  exp(-r^2), r^2 = maha(x,P,y) = (x-y)'*inv(P)*(x-y)
 ```
-where the matrix P is the metric and kappa is a user defined kernel function.
+where the matrix P is the metric.
 
 """
-struct MahaKernel{T<:Real} <: BaseKernel
-    P::AbstractMatrix{T}
-    kappa::Function
-    function MahaKernel(P::AbstractMatrix{T}, kappa::Function) where {T<:Real}
+struct MahalanobisKernel{T<:Real,A<:AbstractMatrix{T}} <: BaseKernel
+    P::A
+    function MahalanobisKernel(P::AbstractMatrix{T}) where {T<:Real}
         LinearAlgebra.checksquare(P)
-        new{T}(P, kappa)
+        new{T}(P)
     end
 end
 
-kappa(κ::MahaKernel, d::T) where {T<:Real} = κ.kappa(d) 
+kappa(κ::MahalanobisKernel, d::T) where {T<:Real} = exp(-d) 
 
-metric(κ::MahaKernel) = SqMahalanobis(κ.P)
+metric(κ::MahalanobisKernel) = SqMahalanobis(κ.P)
--- a/src/kernels/maha.jl
+++ b/src/kernels/maha.jl
@@ -12,7 +12,7 @@ struct MahaKernel{T<:Real} <: BaseKernel
     P::AbstractMatrix{T}
     kappa::Function
     function MahaKernel(P::AbstractMatrix{T}, kappa::Function) where {T<:Real}
-        @assert size(P)[1] == size(P)[2] "P should be a square matrix"
+        LinearAlgebra.checksquare(P)
         new{T}(P, kappa)
     end
 end
--- a/src/kernels/maha.jl
+++ b/src/kernels/maha.jl
@@ -1,0 +1,21 @@
+"""
+    MahaKernel(; ell::Real=1.0, p::Real=1.0)
+
+    Mahalanobis distance-based kernel given by
+```math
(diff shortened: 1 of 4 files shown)


devmotion:
:+1: for MahalanobisKernel. I prefer clear and descriptive names :smiley:

src/kernels/maha.jl
```diff
@@ -0,0 +1,22 @@
...
+    Mahalanobis distance-based kernel given by
+```math
+    κ(x,y) =  kappa(r^2), r^2 = maha(x,P,y) = (x-y)'*inv(P)*(x-y)
+```
+where the matrix P is the metric and kappa is a user defined kernel function.
+
+"""
+struct MahaKernel{T<:Real} <: BaseKernel
+    P::AbstractMatrix{T}
+    kappa::Function
```
devmotion: This will lead to bad performance, see https://docs.julialang.org/en/v1/manual/performance-tips/index.html#Avoid-fields-with-abstract-type-1. Please change it to
```suggestion
struct MahaKernel{T<:Real,A<:AbstractMatrix{T},F} <: BaseKernel
    P::A
    kappa::F
```

I'm also wondering if we actually want to allow a flexible `kappa`. IIRC this would be the first kernel where you could specify a positive definite function, so it does not seem completely in line with the current design. What's the general opinion on this?

src/kernels/maha.jl
```diff
@@ -0,0 +1,22 @@
...
+    function MahaKernel(P::AbstractMatrix{T}, kappa::Function) where {T<:Real}
+        @assert size(P)[1] == size(P)[2] "P should be a square matrix"
+        new{T}(P, kappa)
+    end
+end
+
+kappa(κ::MahaKernel, d::T) where {T<:Real} = κ.kappa(d) 
+
+metric(κ::MahaKernel) = Mahalanobis(κ.P)
```
devmotion: I think you want to use
```suggestion
metric(κ::MahaKernel) = SqMahalanobis(κ.P)
```
here.

src/kernels/maha.jl
```diff
@@ -0,0 +1,22 @@
...
+    κ(x,y) =  kappa(r^2), r^2 = maha(x,P,y) = (x-y)'*inv(P)*(x-y)
+```
+where the matrix P is the metric and kappa is a user defined kernel function.
+
+"""
+struct MahaKernel{T<:Real} <: BaseKernel
+    P::AbstractMatrix{T}
+    kappa::Function
+    function MahaKernel(P::AbstractMatrix{T}, kappa::Function) where {T<:Real}
+        @assert size(P)[1] == size(P)[2] "P should be a square matrix"
```
devmotion: I'd suggest
```suggestion
        @assert size(P, 1) == size(P, 2) "P should be a square matrix"
```
or just
```suggestion
        LinearAlgebra.checksquare(P)
```

theogf:
Given all these specific kernels, I start to wonder if we should start to define a new type of distance that directly take two inputs. Btw couldn't Maha be defined by a transform of a projection matrix L where P=LL^T?

sharanry:
@theogf I am not exactly sure what you mean. Doesn't distance measure always take in two inputs?

theogf:
> Btw couldn't Maha be defined by a transform of a projection matrix L where P=LL^T?

Oh sorry about the confusion I did not see you were already using the Mahalanobis distance from Distances.jl. 

> if we should start to define a new type of distance that directly take two inputs

I am not sure what I meant anymore.... It was probably late.

theogf:
About the `kappa` problem, I would of the opinion to use the `exp(-d)` kappa as a standard (like in the SEKernel).

src/kernels/maha.jl
sharanry: Oh thanks for catching this!

sharanry:
@theogf I like the idea of using `exp(-d)` if it is a standard. This way, we would not be diverging from our current design as @devmotion point out.

devmotion:
I agree, I think that's a good solution. Another advantage: it's also easier to dispatch on the kernels if we do not allow arbitrary `kappa`. Otherwise IMO one would have to consider coming up with dedicated types or traits for the positive definite function.

theogf:
One more thing! In `trainable.jl` can you add that a function for `MahalanobisKernel` ?

sharanry:
@theogf Is this ready to merge?

src/kernels/maha.jl
```diff
@@ -0,0 +1,21 @@
...
+```
+where the matrix P is the metric.
+
+"""
+struct MahalanobisKernel{T<:Real, A<:AbstractMatrix{T}} <: BaseKernel
+    P::A
+    function MahalanobisKernel(P::AbstractMatrix{T}) where {T<:Real}
+        LinearAlgebra.checksquare(P)
+        new{T,AbstractMatrix{T}}(P)
+    end
```
theogf: Here the type of `A` is going to be `AbstractMatrix{T}` which is not optimal. 
Instead do something like this :
`new{T, typeof(P)}(P)`

sharanry:
@theogf Few of the builds seem to have failed. Not sure why exactly. I am guessing it is because of some external reason. Could you restart it?

PR #63, opened 2020-03-27
Refactors the test suite

willtebbutt:
Doesn't modify any of the tests beyond adding the occasional random number generator explicitly.

Now that we're having more contributions to the repo, it makes sense to formalise the structure of our tests. The one I'm proposing is widely used and I have always found it to work well.

edit: ~~I appear to have lost two tests somewhere, but I'm not sure where 😬~~ I found them, and some extra ones!


test/runtests.jl
```diff
@@ -1,18 +1,101 @@
...
+
+    @testset "matrix" begin
+        include(joinpath("matrix", "kernelmatrix.jl"))
+        include(joinpath("matrix", "kernelkroneckermat.jl"))
+        include(joinpath("matrix", "kernelpdmat.jl"))
+    end
+
+    @testset "approximations" begin
+        include(joinpath("approximations", "nystrom.jl"))
+    end
```
theogf: Replace each list of include by 
```
for f in readdir(folder)
   include(joinpath(folder,f))
end
```
as in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/56/files#diff-7a2da5c915dbe72497c7dfda6c7fcb8b

willtebbutt: Ah, yeah, I specifically don't want to do that. See [here](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/63/files#diff-fce720c43af3c52c862fd7451c7374b8R40)

theogf: Looks like I stopped reading after point 5 :sweat_smile:

willtebbutt:
Ah, interesting -- I had not come across `SafeTestSets` before. I can see how they would be helpful. What does the workflow look like if I want to have some utility functionality floating around? Say I have a file called `test_utils` located in `test`, which provides some utility functionality that I want to use in, say, `test/kernels/exponential`. Does putting everything inside its own module not complicate that procedure somewhat?

As regards testing individual files, my preferred workflow is simply to comment out tests that you don't want to run during development.

The benefit of having all of the `using`s in the same file is that it prevents you from accidentally depending on something `using`-ed in another test file which (which is really easy to do and happens all of the time), if you comment it out / don't run it, causes annoying errors.

devmotion:
> The benefit of having all of the usings in the same file is that it prevents you from accidentally depending on something using-ed in another test file which (which is really easy to do and happens all of the time), if you comment it out / don't run it, causes annoying errors.

That's exactly why I like SafeTestsets - you can keep the `using` statements in the individual files (so you can just run them, include them in the REPL, Juno etc.) if you wrap them inside a `@safetestset` statement.

> Say I have a file called test_utils located in test, which provides some utility functionality that I want to use in, say, test/kernels/exponential. Does putting everything inside its own module not complicate that procedure somewhat? (I'm imagining that you would include("test_utils.jl") directly in runtests.jl)

If you want to be able to run each test file separately, you have to add `include("test_utils.jl")` to each test file that uses the utility functionality. If you don't care about this and rather load the utilities in runtests.jl, I guess you have to import that functionality in each test file (but I think that's a bit weird, I've never done this).

willtebbutt:
@devmotion I think you've got some really good points. Would you be happy for us to merge this, and you to open another PR with the proposed approach? I would like to get this in so that I can push on some other stuff concurrently.

devmotion:
Sure, go ahead!

PR #66, opened 2020-03-27
Corrected docs and improved formatting

theogf:
Corrected obsolete docs and added print functions for a lot of kernels

### DIFF
--- a/src/kernels/exponential.jl
+++ b/src/kernels/exponential.jl
@@ -46,9 +46,9 @@ const LaplacianKernel = ExponentialKernel
 
 The γ-exponential kernel is an isotropic Mercer kernel given by the formula:
 ```
-    κ(x,y) = exp(-ρ^(2γ)‖x-y‖^(2γ))
+    κ(x,y) = exp(-‖x-y‖^(2γ))
 ```
-Where `γ>0`, (the keyword `γ` can be replaced by `gamma`)
+Where `γ > 0`, (the keyword `γ` can be replaced by `gamma`)
 For `γ = 1`, see `SqExponentialKernel` and `γ = 0.5`, see `ExponentialKernel`
 """
 struct GammaExponentialKernel{Tγ<:Real} <: BaseKernel
--- a/src/kernels/exponential.jl
+++ b/src/kernels/exponential.jl
@@ -1,10 +1,11 @@
 """
-`SqExponentialKernel()`
+    SqExponentialKernel()
 
-The squared exponential kernel is an isotropic Mercer kernel given by the formula:
+The squared exponential kernel is a Mercer kernel given by the formula:
 ```
     κ(x,y) = exp(-‖x-y‖²)
 ```
+Can also be called via `SEKernel`, `GaussianKernel` or `SEKernel`.
 See also [`ExponentialKernel`](@ref) for a
 related form of the kernel or [`GammaExponentialKernel`](@ref) for a generalization.
 """
@@ -12,7 +13,6 @@ struct SqExponentialKernel <: BaseKernel end
 
 kappa(κ::SqExponentialKernel, d²::Real) = exp(-d²)
 iskroncompatible(::SqExponentialKernel) = true
-
 metric(::SqExponentialKernel) = SqEuclidean()
 
 Base.show(io::IO,::SqExponentialKernel) = print(io,"Squared Exponential Kernel")
@@ -23,10 +23,11 @@ const GaussianKernel = SqExponentialKernel
 const SEKernel = SqExponentialKernel
 
 """
-`ExponentialKernel([ρ=1.0])`
-The exponential kernel is an isotropic Mercer kernel given by the formula:
+    ExponentialKernel()
+
+The exponential kernel is a Mercer kernel given by the formula:
 ```
-    κ(x,y) = exp(-ρ‖x-y‖)
+    κ(x,y) = exp(-‖x-y‖)
 ```
 """
 struct ExponentialKernel <: BaseKernel end
@@ -35,21 +36,23 @@ kappa(κ::ExponentialKernel, d::Real) = exp(-d)
 iskroncompatible(::ExponentialKernel) = true
 metric(::ExponentialKernel) = Euclidean()
 
-Base.show(io::IO,::ExponentialKernel) = print(io,"Exponential Kernel")
+Base.show(io::IO, ::ExponentialKernel) = print(io, "Exponential Kernel")
 
 ## Alias ##
 const LaplacianKernel = ExponentialKernel
 
 """
-`GammaExponentialKernel([ρ=1.0, [γ=2.0]])`
+    GammaExponentialKernel(; γ = 2.0)
 The γ-exponential kernel is an isotropic Mercer kernel given by the formula:
 ```
     κ(x,y) = exp(-ρ^(2γ)‖x-y‖^(2γ))
 ```
+Where `γ>0`, (the keyword `γ` can be replaced by `gamma`)
+For `γ = 1`, see `SqExponentialKernel` and `γ = 0.5`, see `ExponentialKernel`
 """
 struct GammaExponentialKernel{Tγ<:Real} <: BaseKernel
     γ::Vector{Tγ}
-    function GammaExponentialKernel(;gamma::T=2.0, γ::T=gamma) where {T<:Real}
+    function GammaExponentialKernel(; gamma::T=2.0, γ::T=gamma) where {T<:Real}
         @check_args(GammaExponentialKernel, γ, γ >= zero(T), "γ > 0")
         return new{T}([γ])
     end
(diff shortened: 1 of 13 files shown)


src/kernels/exponentiated.jl
```diff
@@ -1,14 +1,15 @@
 """
-`ExponentiatedKernel([ρ=1])`
+    ExponentiatedKernel()`
```
devmotion:
```suggestion
    ExponentiatedKernel()
```
I think, we should also consistently use the constructor in the doc strings (e.g., that's not the case for `CosineKernel`).

src/kernels/exponential.jl
```diff
@@ -35,21 +36,23 @@ kappa(κ::ExponentialKernel, d::Real) = exp(-d)
...
 
-Base.show(io::IO,::ExponentialKernel) = print(io,"Exponential Kernel")
+Base.show(io::IO, ::ExponentialKernel) = print(io, "Exponential Kernel")
 
 ## Alias ##
 const LaplacianKernel = ExponentialKernel
 
 """
-`GammaExponentialKernel([ρ=1.0, [γ=2.0]])`
+    GammaExponentialKernel(; γ = 2.0)
```
devmotion:
```suggestion
    GammaExponentialKernel(; γ = 2.0)

```
I assume the default is `2.0` and not `2` due to possible optimization of the hyperparameter?

src/kernels/exponential.jl
```diff
@@ -35,21 +36,23 @@ kappa(κ::ExponentialKernel, d::Real) = exp(-d)
...
 The γ-exponential kernel is an isotropic Mercer kernel given by the formula:
 ```
     κ(x,y) = exp(-ρ^(2γ)‖x-y‖^(2γ))
```
devmotion:
```suggestion
    κ(x,y) = exp(-‖x-y‖^(2γ))
```

src/kernels/exponential.jl
devmotion: BTW I guess 2 is a standard choice in literature?

src/kernels/kernelproduct.jl
```diff
@@ -1,13 +1,13 @@
 """
-`KernelProduct(kernels::Array{Kernel})`
-Create a multiplication of kernels.
-One can also use the operator `*`
+    KernelProduct(kernels::Array{Kernel})
```
devmotion:
```suggestion
    KernelProduct(kernels::Array{Kernel})

```

src/kernels/matern.jl
```diff
@@ -1,6 +1,6 @@
 """
-`MaternKernel([ρ=1.0,[ν=1.0]])`
-The matern kernel is an isotropic Mercer kernel given by the formula:
+    MaternKernel(; ν=1.0)
```
devmotion:
```suggestion
    MaternKernel(; ν=1.0)

```

src/kernels/polynomial.jl
```diff
@@ -1,8 +1,8 @@
 """
-`LinearKernel([ρ=1.0,[c=0.0]])`
+    LinearKernel(; c=0.0)
```
devmotion:
```suggestion
    LinearKernel(; c=0.0)

```

devmotion: BTW do we actually need the `c` here? Couldn't it be added with a constant kernel anyways?

src/kernels/polynomial.jl
```diff
@@ -14,16 +14,17 @@ struct LinearKernel{Tc<:Real} <: BaseKernel
...
 
 kappa(κ::LinearKernel, xᵀy::Real) = xᵀy + first(κ.c)
-
 metric(::LinearKernel) = DotProduct()
 
+Base.show(io::IO, κ::LinearKernel) = print(io, "Linear Kernel (c = $(first(κ.c)))")
+
 """
-`PolynomialKernel([ρ=1.0[,d=2.0[,c=0.0]]])`
+    PolynomialKernel(; d = 2.0, c = 0.0)
```
devmotion:
```suggestion
    PolynomialKernel(; d = 2.0, c = 0.0)

```

devmotion: Same here, do we need the `c` and couldn't the exponentiation be implemented using a transform of a linear kernel?

src/kernels/rationalquad.jl
```diff
@@ -1,6 +1,6 @@
 """
-RationalQuadraticKernel([ρ=1.0[,α=2.0]])
-The rational-quadratic kernel is an isotropic Mercer kernel given by the formula:
+    RationalQuadraticKernel(; α = 2.0)
```
devmotion:
```suggestion
    RationalQuadraticKernel(; α = 2.0)

```

src/kernels/scaledkernel.jl
```diff
@@ -1,5 +1,5 @@
 """
-`ScaledKernel(k::Kernel,σ²::Real)`
+    ScaledKernel(k::Kernel, σ²::Real)
```
devmotion:
```suggestion
    ScaledKernel(k::Kernel, σ²::Real)

```

src/kernels/transformedkernel.jl
```diff
@@ -1,6 +1,7 @@
 """
-`TransformedKernel(k::Kernel,t::Transform)`
+    TransformedKernel(k::Kernel,t::Transform)
```
devmotion:
```suggestion
    TransformedKernel(k::Kernel,t::Transform)

```

src/kernels/matern.jl
```diff
@@ -26,28 +26,32 @@ end
 
 metric(::MaternKernel) = Euclidean()
 
+Base.show(io::IO, κ::MaternKernel) = print(io, "Matern Kernel (ν = $(first(κ.ν)))")
+
 """
-`Matern32Kernel([ρ=1.0])`
-The matern 3/2 kernel is an isotropic Mercer kernel given by the formula:
+    Matern32Kernel()
```
devmotion:
```suggestion
    Matern32Kernel()

```

src/kernels/matern.jl
```diff
@@ -26,28 +26,32 @@ end
...
 kappa(κ::Matern32Kernel, d::Real) = (1 + sqrt(3) * d) * exp(-sqrt(3) * d)
-
 metric(::Matern32Kernel) = Euclidean()
 
+Base.show(io::IO, ::Matern32Kernel) = print(io, "Matern 3/2 Kernel")
+
 """
-`Matern52Kernel([ρ=1.0])`
-The matern 5/2 kernel is an isotropic Mercer kernel given by the formula:
+    Matern52Kernel()
```
devmotion:
```suggestion
    Matern52Kernel()

```

src/kernels/kernelsum.jl
```diff
@@ -1,14 +1,15 @@
 """
-`KernelSum(kernels::Array{Kernel};weights::Array{Real}=ones(length(kernels)))`
-Create a positive weighted sum of kernels.
+    KernelSum(kernels::Array{Kernel}; weights::Array{Real}=ones(length(kernels)))
```
devmotion: IMO we could/should remove the weights since a weighted kernel is just a `ScaledKernel`.

devmotion: Additionally, I have been thinking about restricting the KernelSum and the KernelProduct to only two kernels since
1. that would easily allow to obtain a concretely typed fields (I don't like the fact that both types use abstract fields),
2. that would make it trivial to check if the metric of both kernels (if existent) is the same and hence allow to optimize computations in such cases by only evaluating the metric once.
I guess the same could be achieved by considering tuples of kernels and checking all there metrics but it doesn't feel as clean.

Do you want me to open an issue for it?

src/kernels/exponential.jl
theogf: What do you mean by default is `2.0` and not `2`? I set the default as 2.0 a bit randomly

src/kernels/polynomial.jl
theogf: That's how it's usually defined in literature. I don't think it's such an overlap with the Constant kernel

src/kernels/polynomial.jl
theogf: Here we do need the `c` since it's going to be involved in the polynome. That's also how it is usually defined

src/kernels/kernelsum.jl
theogf: Regarding 1. : that's a good point, that would avoid this weird thing I have to do with sums of scaled kernels.
For 2. I am not sure how this would go in terms of performance. An issue would be very welcome to explore the options

src/kernels/exponential.jl
devmotion: I mean, why using a floating point number instead of an integer as default? My assumption was that you can't really train an integer, and hence the initialization as a (vector of) floating point number might be more suitable.

theogf: Ah yes, I get it now. \gamma belong to the positive reals. Hence the initialization as a Float

theogf:
Wuhu I finally fixed the docs transitions!

src/kernels/polynomial.jl
devmotion: I see :slightly_smiling_face:

theogf:
Will add the tests once #63 is merged

PR #69, opened 2020-03-28
Update Travis configuration

devmotion:
This PR should ensure that each commit is only tested once, that coverage analysis is only performed with Julia 1.4, and the documentation is built with Julia 1.4.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 ## Documentation: http://docs.travis-ci.com/user/languages/julia/
 language: julia
+branches:
+  only:
+    - master
 os:
   - linux
   - osx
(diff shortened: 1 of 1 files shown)


devmotion:
Ah it seems another problem is that the PRs are tested both on travis-ci.org and travis-ci.com. I guess the ones running on travis-ci.org should be disabled.

theogf:
I am not sure I know the difference between travis-ci.org and travis-ci.com

devmotion:
Basically, travis-ci.org is the legacy service and travis-ci.com is the new service that one can use with Travis GitHub App (e.g., on https://docs.travis-ci.com/user/migrate/open-source-on-travis-ci-com/ you can find more information). It's definitely not useful to run tests with both services, so I guess one could just use travis-ci.com.

theogf:
Ok i think I deactivated travis-ci.org. Let's see if it works

theogf:
Ok it seems to be fixed

PR #71, opened 2020-03-29
Add alias for SelectTransform as Mask

### LINKED ISSUE #47: Rename SelectTransform by Mask
Just realised that `Mask` was the most used name for this.

ritesh99rakesh:
Fixes #47 

I have added alias for `SelectTransform` as `Mask` and also added tests for the same.

Please review.


theogf:
Can you fix the PR so the tests run ? (missing export I believe)
Also I don't know if a `mask` function would make sense :
```
mask(kernel, [2, 3, 4]) == transform(kernel, SelectTransform([2, 3, 4])
```
What do you think @willtebbutt , @devmotion  ?

willtebbutt:
I don't want us to be inconsistent. If we have a `mask` function, then we should have similar functions for the other transformations. Since that's not the case currently, we should go with the type and make providing aliases for transform a separate PR.

Also, @ritesh99rakesh I think you'll need to rebase onto the current master because the test suite has changed.

theogf:
Should I close this as it seems it's been given up on ?

PR #75, opened 2020-03-31
Use SafeTestsets?

devmotion:
This PR is a follow-up to the discussion in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/63. It shows how a test setup with SafeTestsets could look like.

Additionally, I fixed and enabled the tests of the Mahalanobis kernel.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,16 +1,6 @@
-using Distances
-using FiniteDifferences
-using Flux
-using KernelFunctions
-using Kronecker
-using LinearAlgebra
-using PDMats
-using Random
-using SpecialFunctions
-using Test
-using Zygote
+using SafeTestsets
 
-using KernelFunctions: metric
+using Test
 
 # Writing tests:
 # 1. The file structure of the test should match precisely the file structure of src.
@@ -18,16 +8,15 @@ using KernelFunctions: metric
 #   This makes it trivially easy for someone to find the tests associated with a particular
 #   src file.
 # 2. A consequence of 1 is that there should be exactly 1 test file per src file.
-# 3. A test file called foo.jl should have the structure:
-#   @testset "foo" begin
-#       code
-#   end
+# 3. Use whatever testset structure seems appropriate within a given file. eg. if multiple
+#   types / functions are defined in a particular source file, you might want multiple
+#   testsets in the test file.
+# 4. Each directory should have its own testset, in which each test file called `foo.jl` is
+#   `include`d in a `@safetestset` called "foo":
+#   @safetestset "foo" begin include(joinpath(..., "foo.jl")) end
 #
-#   Note that the testset is called `foo`, not `foo.jl`. Use whatever testset structure
-#   seems appropriate within a given file. eg. if multiple types / functions are defined in
-#   a particular source file, you might want multiple testsets in the test file.
-# 4. Each directory should have its own testset, in which each test file is `include`d.
-# 5. Each test file should create its own state, and shouldn't rely on state defined in
+#   Note that the testset is called `foo`, not `foo.jl`.
+# 5. Each test file creates its own state, and must not rely on state defined in
 #   other test files. e.g. don't define a matrix used by all of the files in kernels. If
 #   two test files are similar enough to share state, perhaps the corresponding source code
 #   should be in the same file.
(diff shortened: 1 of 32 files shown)


test/runtests.jl
```diff
@@ -36,63 +25,91 @@ using KernelFunctions: metric
...
-        include(joinpath("kernels", "exponentiated.jl"))
-        include(joinpath("kernels", "fbm.jl"))
-        include(joinpath("kernels", "kernelproduct.jl"))
-        include(joinpath("kernels", "kernelsum.jl"))
-        include(joinpath("kernels", "matern.jl"))
-        include(joinpath("kernels", "polynomial.jl"))
-        include(joinpath("kernels", "rationalquad.jl"))
-        include(joinpath("kernels", "scaledkernel.jl"))
-        include(joinpath("kernels", "transformedkernel.jl"))
+        @safetestset "constant" begin include(joinpath("kernels", "constant.jl")) end
```
theogf: I guess the `begin` `end` cannot be removed?

devmotion: No, since it ends up being just a `@testset` that is wrapped inside its own module and `@testset` needs a begin/end block or a for loop:
```julia
  LoadError: Expected begin/end block or for loop as argument to @testset
```

test/runtests.jl
```diff
@@ -1,33 +1,22 @@
...
-# 3. A test file called foo.jl should have the structure:
-#   @testset "foo" begin
-#       code
-#   end
+# 3. Use whatever testset structure seems appropriate within a given file. eg. if multiple
+#   types / functions are defined in a particular source file, you might want multiple
+#   testsets in the test file.
+# 4. Each directory should have its own testset, in which each test file called `foo.jl` is
+#   `include`d in a `@safetestset` called "foo":
+#   @safetestset "foo" begin include(joinpath(..., "foo.jl")) end
```
willtebbutt, line 16: Is there a particular reason for this, rather than putting the `@safetestset` inside the test file it's targetting?

devmotion: It just felt more natural to me. I thought, the main motivation for using `@safetestset` is not to mix imports and definitions of different test files, and hence it is only/mostly relevant when including different test files such as in `runtests.jl`.

willtebbutt: Ah I see.

> It just felt more natural to me.

I imagine this is a matter of preference. Personally I think I would prefer to have a new `@safetestset` in each test file so that when you run it by itself, you get a new testset. So the test files would be of the form

```julia
@safetestset "foo" begin
    using KernelFunctions
    using Test
    using Random

    @testset "some tests" begin

    end

    @testset "some more tests" begin

    end
end
```
This would tidy up the `runtests.jl` file a bit, and make the results of running a particular test suite nicely encapsulated if I run `include("test/mytests.jl")`. If you don't include stuff inside a `@safetestset` in each file you'll add state to your global state if running from the REPL.

theogf: Probably it's a bad practice but I actually like to have a global state when some debugging is needed!

willtebbutt: Sure, but they're absolutely not something that you want when you're running your test sets. I can get on board with producing some global state to figure out what's going on with a particular thing, but then I want to be able to re-run my tests and know that I'm not going to get different results if I restart my repl session.

theogf: > I definitely think you want to be able to run include("test/some_tests.jl") and know that what you've been doing at the repl isn't going to effect the results.

Ok yeah that's fair

>  you can just copy + paste some code from the test set to the repl,

I am really not a big fan of the REPL copy paste

theogf: Anyway I agree your point is stronger than my lazy practices :smile:

willtebbutt: > I am really not a big fan of the REPL copy paste

😆 that's totally fair, can definitely get on board with that.

devmotion: I agree, it seems reasonable to expect that `include`ing the test file should not affect the global state. However, it seems the main advantage of `@safetestset` over just having separate `module`s in every test file would be that the names of the modules are generated with `gensym`. I'm not sure if that justifies the use of `@safetestset`, in particular since one has to import SafeTestsets before including such a test file whereas including a test file that defines its own module would just work. So it might actually be more reasonable to not use SafeTestsets?

willtebbutt: I can't say that I'm overly concerned at the idea of having to load `SafeTestsets` to be able to run the files -- that wouldn't be a deal breaker from a workflow perspective I don't think. I'm easy on this one -- if people would prefer to use `SafeTestsets` and are happy to load it before including particular files, that's fine by me. I'm also completely happy with our current test set up.

We could always revisit this is in the future if we feel it's worthwhile.

devmotion: I'll close this PR for now since I'm also happy with the current setup. And as you say, we can always revisit this in the future.

PR #76, opened 2020-04-01
Add Piecewise Polynomial Kernel

sharanry:
Issue #44 

*Piecewise Polynomial Kernel*

Piecewise Polynomial covariance function with compact support, v = 0,1,2,3.
The kernel functions are 2v times continuously differentiable and the corresponding
processes are hence v times  mean-square differentiable. The kernel function is:
```math
    κ(x,y) = max(1-r,0)^(j+v) * f(r,j) with j = floor(D/2)+v+1
```
where r is the Mahalanobis distance sqrt(maha(x,z)).

**References**
https://github.com/alshedivat/gpml/blob/master/cov/covPP.m
www.gaussianprocess.org/gpml/code/matlab/doc/manual.pdf pg44

**Issues**
Currently the dimension `D` of the inputs x and y is not being taken properly to calculate `j`.

### DIFF
--- a/src/kernels/piecewisepolynomial.jl
+++ b/src/kernels/piecewisepolynomial.jl
@@ -58,17 +58,17 @@ function kernelmatrix(
     X::AbstractMatrix;
     obsdim::Int = defaultobs
 ) where {V}
-    j = div(size(X, 3-obsdim), 2) + V + 1
+    j = div(size(X, feature_dim(obsdim)), 2) + V + 1
     return map(r->_piecewisepolynomial(κ,r,j),pairwise(metric(κ),X,dims=obsdim))
 end
 
 function _kernelmatrix(κ::PiecewisePolynomialKernel{V},X,Y,obsdim) where {V}
-    j = div(size(X, 3-obsdim), 2) + V + 1
+    j = div(size(X, feature_dim(obsdim)), 2) + V + 1
     return map(r->_piecewisepolynomial(κ,r,j),pairwise(metric(κ),X,Y,dims=obsdim))
 end
 
 function kernelmatrix!(
-    K::AbstractMatrix,
+    K::AbstractMatrix
     κ::PiecewisePolynomialKernel{V},
     X::AbstractMatrix;
     obsdim::Int = defaultobs
@@ -77,7 +77,7 @@ function kernelmatrix!(
     if !check_dims(K,X,X,feature_dim(obsdim),obsdim)
         throw(DimensionMismatch("Dimensions of the target array K $(size(K)) are not consistent with X $(size(X))"))
     end
-    j = div(size(X, 3-obsdim), 2) + V + 1
+    j = div(size(X, feature_dim(obsdim)), 2) + V + 1
     map!(r->_piecewisepolynomial(κ,r,j),K,pairwise(metric(κ),X,dims=obsdim))
 end
 
@@ -92,7 +92,7 @@ function kernelmatrix!(
     if !check_dims(K,X,Y,feature_dim(obsdim),obsdim)
         throw(DimensionMismatch("Dimensions $(size(K)) of the target array K are not consistent with X ($(size(X))) and Y ($(size(Y)))"))
     end
-    j = div(size(X, 3-obsdim), 2) + V + 1
+    j = div(size(X, feature_dim(obsdim)), 2) + V + 1
     map!(r->_piecewisepolynomial(κ,r,j),K,pairwise(metric(κ),X,Y,dims=obsdim))
 end
 
--- a/src/kernels/piecewisepolynomial.jl
+++ b/src/kernels/piecewisepolynomial.jl
@@ -62,7 +62,7 @@ function kernelmatrix(
     return map(r->_piecewisepolynomial(κ, r, j), pairwise(metric(κ), X; dims=obsdim))
 end
 
-function _kernelmatrix(κ::PiecewisePolynomialKernel{V}, X, Y; obsdim) where {V}
+function _kernelmatrix(κ::PiecewisePolynomialKernel{V}, X, Y, obsdim) where {V}
     j = div(size(X, feature_dim(obsdim)), 2) + V + 1
     return map(r->_piecewisepolynomial(κ, r, j), pairwise(metric(κ), X, Y; dims=obsdim))
 end
--- a/src/kernels/piecewisepolynomial.jl
+++ b/src/kernels/piecewisepolynomial.jl
@@ -5,7 +5,7 @@ Piecewise Polynomial covariance function with compact support, V = 0,1,2,3.
 The kernel functions are 2v times continuously differentiable and the corresponding
 processes are hence v times  mean-square differentiable. The kernel function is:
 ```math
-    κ(x,y) = max(1-r,0)^(j+V) * f(r,j) with j = floor(D/2)+V+1
+    κ(x, y) = max(1 - r, 0)^(j + V) * f(r, j) with j = floor(D / 2) + V + 1
 ```
 where `r` is the Mahalanobis distance mahalanobis(x,y) with `maha` as the metric.
 
--- a/src/kernels/piecewisepolynomial.jl
+++ b/src/kernels/piecewisepolynomial.jl
@@ -100,7 +100,6 @@ function kernelmatrix!(
     map!(r->_piecewisepolynomial(κ,r,j),K,pairwise(metric(κ),X,Y,dims=obsdim))
 end
 
-
 metric(κ::PiecewisePolynomialKernel) = Mahalanobis(κ.maha)
 
 Base.show(io::IO, κ::PiecewisePolynomialKernel{V}) where V = print(io, "Piecewise Polynomial Kernel (v = $(V), size(maha) = $(size(κ.maha))")
--- a/src/kernels/piecewisepolynomial.jl
+++ b/src/kernels/piecewisepolynomial.jl
@@ -1,3 +1,5 @@
+abstract type V{Integer} end
+
 """
     PiecewisePolynomialKernel(v::Type, maha::AbstractMatrix)
 
@@ -10,13 +12,13 @@ processes are hence v times  mean-square differentiable. The kernel function is:
 where r is the Mahalanobis distance mahalanobis(x,y) with `maha` as the metric.
 
 """
-struct PiecewisePolynomialKernel{T1<:T1, T2<:Real, A<:AbstractMatrix{T2}} <: BaseKernel
-    v::T1
+struct PiecewisePolynomialKernel{V, A<:AbstractMatrix{<:Real}} <: BaseKernel
+    v::Integer
     maha::A
-    function PiecewisePolynomialKernel(v::T1, maha::AbstractMatrix{T2}) where {T1<:Integer,T2<:Real}
+    function PiecewisePolynomialKernel(v, maha::AbstractMatrix{<:Real})
         if v ∉ [0,1,2,3] error("Invalid paramter v=$(v). Should be 0,1,2 or 3.") end
         LinearAlgebra.checksquare(maha)
-        new{T1,T2,typeof(maha)}(v,maha)
+        new{V{v},typeof(maha)}(v,maha)
     end
 end
 
--- a/src/kernels/piecewisepolynomial.jl
+++ b/src/kernels/piecewisepolynomial.jl
@@ -13,12 +13,11 @@ where r is the Mahalanobis distance mahalanobis(x,y) with `maha` as the metric.
 
 """
 struct PiecewisePolynomialKernel{V, A<:AbstractMatrix{<:Real}} <: BaseKernel
-    v::Integer
     maha::A
-    function PiecewisePolynomialKernel(v, maha::AbstractMatrix{<:Real})
-        if v ∉ [0,1,2,3] error("Invalid paramter v=$(v). Should be 0,1,2 or 3.") end
+    function PiecewisePolynomialKernel{V}(maha::AbstractMatrix{<:Real}) where V
+        V in (0, 1, 2, 3) || error("Invalid paramter v=$(V). Should be 0, 1, 2 or 3.")
         LinearAlgebra.checksquare(maha)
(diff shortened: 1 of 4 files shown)


src/kernels/piecewisepolynomial.jl
```diff
@@ -0,0 +1,43 @@
+"""
+    PiecewisePolynomialKernel(P::AbstractMatrix)
```
devmotion, line 2: That's not the actual constructor it seems?

sharanry: Yeah! For got to change that. Thanks!

src/kernels/piecewisepolynomial.jl
```diff
@@ -0,0 +1,43 @@
...
+    κ(x,y) = max(1-r,0)^(j+v) * f(r,j) with j = floor(D/2)+v+1
+```
+where r is the Mahalanobis distance sqrt(maha(x,z)).
+
+"""
+struct PiecewisePolynomialKernel{T1<:Integer, T2<:Real, A<:AbstractMatrix{T2}} <: BaseKernel
+    v::T1
+    maha::A
+    function PiecewisePolynomialKernel(v::T1, maha::AbstractMatrix{T2}) where {T1<:Integer,T2<:Real}
+        if v ∉ [0,1,2,3] error("Invalid paramter v=$(v). Should be 0,1,2 or 3.") end
```
devmotion, line 17: Maybe `v` could be a type parameter? Then the branches in the function evaluations below would be optimized for the given value of `v` and be compiled away (i.e., instead of runtime checks the branching would happen at compile time).

src/kernels/piecewisepolynomial.jl
```diff
@@ -0,0 +1,43 @@
...
+        return 1 + (j+2)*r +   (j^2+ 4*j+ 3)/ 3*r.^2
+    elseif κ.v==3
+        return 1 + (j+3)*r +   (6*j^2+36*j+45)/15*r.^2 + (j^3 + 9*j^2 + 23*j + 15)/15 * r.^3
+    else
+        error("Invalid paramter v=$(v). Should be 0,1,2 or 3.")
+    end
+end
+
+function kappa(κ::PiecewisePolynomialKernel, r::T) where {T<:Real} 
+    j = floor(size(r,2)/2 + κ.v + 1)
```
devmotion, line 38: I'd recommend using
```suggestion
    j = div(size(r, 2), 2) + κ.v + 1
```
instead.

src/kernels/piecewisepolynomial.jl
sharanry: So are you suggesting we create types like `P0`, `P1`, `P2` and `P3` which is what we expect as a parameter `v`?

devmotion: No, I suggest defining
```suggestion
struct PiecewisePolynomialKernel{V,A<:AbstractMatrix{<:Real}} <: BaseKernel
```
where `V` will be 0, 1, 2, or 3.

sharanry: Oh okay. Makes sense.

sharanry:
```julia
struct PiecewisePolynomialKernel{V, T2<:Real, A<:AbstractMatrix{T2}} <: BaseKernel
    v::V
    maha::A
    function PiecewisePolynomialKernel(v::T1, maha::AbstractMatrix{T2}) where {T1<:Integer,T2<:Real}
        if v ∉ [0,1,2,3] error("Invalid paramter v=$(v). Should be 0,1,2 or 3.") end
        LinearAlgebra.checksquare(maha)
        new{T1,T2,typeof(maha)}(v,maha)
    end
end
```
It this what you mean?

src/kernels/piecewisepolynomial.jl
```diff
@@ -1,3 +1,5 @@
+abstract type V{Integer} end
+
```
devmotion, line 2: No, `V` should just be a type parameter not a separate type. In the same way as you have integers as type parameters indicating the dimension in the types `Array{Float64,1}` and `Array{Float64,2}`.

src/kernels/piecewisepolynomial.jl
```diff
@@ -10,13 +12,13 @@ processes are hence v times  mean-square differentiable. The kernel function is:
 where r is the Mahalanobis distance mahalanobis(x,y) with `maha` as the metric.
 
 """
-struct PiecewisePolynomialKernel{T1<:T1, T2<:Real, A<:AbstractMatrix{T2}} <: BaseKernel
-    v::T1
+struct PiecewisePolynomialKernel{V, A<:AbstractMatrix{<:Real}} <: BaseKernel
+    v::Integer
```
devmotion, line 16: This is not needed.
```suggestion
```

src/kernels/piecewisepolynomial.jl
```diff
@@ -10,13 +12,13 @@ processes are hence v times  mean-square differentiable. The kernel function is:
...
     maha::A
-    function PiecewisePolynomialKernel(v::T1, maha::AbstractMatrix{T2}) where {T1<:Integer,T2<:Real}
+    function PiecewisePolynomialKernel(v, maha::AbstractMatrix{<:Real})
```
devmotion, line 18: This would be changed to
```suggestion
    function PiecewisePolynomialKernel{V}(maha::AbstractMatrix{<:Real}) where V
```
If a user should be allowed to define a kernel with `PiecewisePolynomialKernel(v, maha)` then you should define an outer constructor
```julia
PiecewisePolynomialKernel(v::Integer, maha::AbstractMatrix{<:Real}) = PiecewisePolynomialKernel{v}(maha)
```
However, that won't be type stable.

src/kernels/piecewisepolynomial.jl
```diff
@@ -10,13 +12,13 @@ processes are hence v times  mean-square differentiable. The kernel function is:
...
         if v ∉ [0,1,2,3] error("Invalid paramter v=$(v). Should be 0,1,2 or 3.") end
         LinearAlgebra.checksquare(maha)
-        new{T1,T2,typeof(maha)}(v,maha)
+        new{V{v},typeof(maha)}(v,maha)
```
devmotion, line 21: This line would just be changed to
```suggestion
        new{V,typeof(maha)}(v,maha)
```

src/kernels/piecewisepolynomial.jl
```diff
@@ -10,13 +12,13 @@ processes are hence v times  mean-square differentiable. The kernel function is:
...
 
 """
-struct PiecewisePolynomialKernel{T1<:T1, T2<:Real, A<:AbstractMatrix{T2}} <: BaseKernel
-    v::T1
+struct PiecewisePolynomialKernel{V, A<:AbstractMatrix{<:Real}} <: BaseKernel
+    v::Integer
     maha::A
-    function PiecewisePolynomialKernel(v::T1, maha::AbstractMatrix{T2}) where {T1<:Integer,T2<:Real}
+    function PiecewisePolynomialKernel(v, maha::AbstractMatrix{<:Real})
         if v ∉ [0,1,2,3] error("Invalid paramter v=$(v). Should be 0,1,2 or 3.") end
```
devmotion, line 19:
```suggestion
        V in (0, 1, 2, 3) || error("Invalid paramter v=$(V). Should be 0, 1, 2 or 3.")
```

src/kernels/piecewisepolynomial.jl
```diff
@@ -19,6 +19,16 @@ struct PiecewisePolynomialKernel{V, A<:AbstractMatrix{<:Real}} <: BaseKernel
...
+function Base.getproperty(κ::PiecewisePolynomialKernel, s::Symbol)
+    if s == :maha
+        return getfield(κ,s)
+    elseif s== :v
+        return typeof(κ).parameters[1]
+    else
+        error("Invalid Property")
+    end
+end
+
```
devmotion, line 31: This should not be needed since I guess you should always be able to dispatch on `v` if needed. E.g., you would write
```
Base.show(io::IO, κ::PiecewisePolynomialKernel{V}) where V = print(io, "Piecewise Polynomial Kernel (v = $V, size(maha) = $(size(κ.maha))")
```

sharanry: We use v for calculations internally. Please see `kappa`.

sharanry: So I guess we can follow a similar trick for `kappa` too.

src/kernels/piecewisepolynomial.jl
```diff
@@ -19,34 +19,24 @@ struct PiecewisePolynomialKernel{V, A<:AbstractMatrix{<:Real}} <: BaseKernel
...
+    elseif V==2
         return 1 + (j+2)*r +   (j^2+ 4*j+ 3)/ 3*r.^2
-    elseif κ.v==3
+    elseif V==3
         return 1 + (j+3)*r +   (6*j^2+36*j+45)/15*r.^2 + (j^3 + 9*j^2 + 23*j + 15)/15 * r.^3
     else
-        error("Invalid paramter v=$(v). Should be 0,1,2 or 3.")
+        error("Invalid paramter v=$(V). Should be 0,1,2 or 3.")
     end
 end
```
theogf: For this function you could directly use multiple dispatch on V:
f`unction _f(\kappa::PieceWise{1}, r, j)`
etc`

src/kernels/piecewisepolynomial.jl
```diff
@@ -0,0 +1,102 @@
...
+    @assert length(x) == length(y) "x and y don't have the same dimension!"
+    kappa(κ,x,y)
+end
+
+function kernelmatrix(
+    κ::PiecewisePolynomialKernel{V},
+    X::AbstractMatrix;
+    obsdim::Int = defaultobs
+) where {V}
+    j = div(size(X, 3-obsdim), 2) + V + 1
```
theogf, line 61: Instead of `3-obsdim` you can call `feature_dim(obsdim)`

src/kernels/piecewisepolynomial.jl
```diff
@@ -0,0 +1,102 @@
...
+processes are hence v times  mean-square differentiable. The kernel function is:
+```math
+    κ(x,y) = max(1-r,0)^(j+V) * f(r,j) with j = floor(D/2)+V+1
+```
+where r is the Mahalanobis distance mahalanobis(x,y) with `maha` as the metric.
+
+"""
+struct PiecewisePolynomialKernel{V, A<:AbstractMatrix{<:Real}} <: BaseKernel
+    maha::A
+    function PiecewisePolynomialKernel{V}(maha::AbstractMatrix{<:Real}) where V
```
theogf, line 15: Could you also create an external constructor `PiecewisePolynomialKernel(; v = 1, maha=smthg)`

src/kernels/piecewisepolynomial.jl
```diff
@@ -19,6 +19,13 @@ struct PiecewisePolynomialKernel{V, A<:AbstractMatrix{<:Real}} <: BaseKernel
     end
 end
 
+function PiecewisePolynomialKernel(;v::Integer=0, maha::AbstractMatrix{<:Real})
+    if maha==nothing
+        error("Please specify metric `maha` for Mahalanobis distance measure.")
+    end
```
devmotion: `maha` can't be `nothing` since you specified in the function signature that it has to be a matrix.

devmotion:
```suggestion
```

sharanry: Just committed that. You might not have refreshed the page.

theogf:
@sharanry Note that if you use Juno, you can automatically format your code as @willtebbutt by using  the shortcut ctrl+J -> ctrl+F

sharanry:
@theogf Thanks for the suggestion! I unfortunately use emacs or VS Code. I will check if there is any equivalent linter.

willtebbutt:
> @sharanry Note that if you use Juno, you can automatically format your code as @willtebbutt by using the shortcut ctrl+J -> ctrl+F

I was not aware of this... do you know what it hooks into?

src/kernels/piecewisepolynomial.jl
```diff
@@ -0,0 +1,118 @@
...
+        return 1 + (j + 1) * r
+    elseif V==2
+        return 1 + (j + 2) * r + (j^2 + 4 * j + 3) / 3 * r.^2
+    elseif V==3
+        return 1 + (j + 3) * r + (6 * j^2 + 36j + 45) / 15 * r.^2 +
+            (j^3 + 9 * j^2 + 23j + 15) / 15 * r.^3
+    else
+        error("Invalid paramter v=$(V). Should be 0,1,2 or 3.")
+    end
+end
```
willtebbutt, line 39: One other suggestion, it's probably more idiomatic to implement this by dispatching on the type of `V`. Something like
```julia
_f(κ::PiecewisePolynomialKernel{1}, r, j) = 1
_f(κ::PiecewisePolynomialKernel{2}, r, j) = 1 + (j + 1) * r
```
etc

theogf:
@willtebbutt Yes it is based on https://github.com/domluna/JuliaFormatter.jl

sharanry:
> @willtebbutt Yes it is based on https://github.com/domluna/JuliaFormatter.jl
@theogf I guess we could format the whole of current codebase using this?
```julia
julia> using JuliaFormatter

# Recursively formats all Julia files in the current directory
julia> format(".")
```

willtebbutt:
Hmmm yeah, I'm wary of doing that without first understanding the style guide that it's implicitly imposing. I'm a big fan of Invenia's BlueStyle, and it's what I try to follow in my own packages and would like us to follow here.

theogf:
Moreover that would break all the version history of the files at once...

willtebbutt:
> Moreover that would break all the version history of the files at once...

Haha true. It's something that we can do incrementally in any case. We should definitely start being strict about style now that there are quite a few of us working on this though.

willtebbutt:
@sharanry are you happy with this now? I'm happy to merge if you are.

sharanry:
@willtebbutt LGTM  Thanks!

PR #77, opened 2020-04-07
Add i-times integrated Wiener Kernel

sharanry:
Issue #44 

i-times integrated Wiener process kernel function given by
```julia
    κ(x,y) =  kᵢ(x,y)
```

For i=-1, this is just the white noise covariance, see WhiteKernel.
For i= 0, this is the Wiener process covariance,
for i= 1, this is the integrated Wiener process covariance (velocity),
for i= 2, this is the twice-integrated Wiener process covariance (accel.),
for i= 3, this is the thrice-integrated Wiener process covariance. where `kᵢ` is given by

```julia
    k₋₁(x,y) =  δ(x,y)
    i >= 0, kᵢ(x,y) = 1/ai * min(x,y)^(2i + 1) + bi * min(x,y)^(i+1) * |x-y| * ri(x,y),
    with the coefficients ai, bi and the residual ri(x,y) defined as follows:
        i = 0, ai =   1, bi = 0
        i = 1, ai =   3, bi = 1/  2, ri(x,y) = 1
        i = 2, ai =  20, bi = 1/ 12, ri(x,y) = x + y - 1/2 * min(x,y)
        i = 3, ai = 252, bi = 1/720, ri(x,y) = 5 * max(x,y)² + 2xz + 3min(x,y)²
```

**References:**
 - See the paper *Probabilistic ODE Solvers with Runge-Kutta Means* by Schober, Duvenaud and Hennig, NIPS, 2014, for more details.
 - http://www.gaussianprocess.org/gpml/code/matlab/doc/manual.pdf pg44
 - https://github.com/alshedivat/gpml/blob/master/cov/covW.m


theogf:
Maybe that would make sense to make separate kernels for each `i`, like for the Matern Kernel

sharanry:
@theogf Considering we are using parameterized types right now, would separate kernels make much of a difference in terms of performance? Currently, IMO even the usage seems straight forward.

src/kernels/wiener.jl
```diff
@@ -0,0 +1,158 @@
...
+    elseif I==1
+        return 1/3   * minXY^(2I + 1) + 1/2   * minXY^(I+1) * euclidean(x,y)
+    elseif I==2
+        return 1/20  * minXY^(2I + 1) + 1/12  * minXY^(I+1) * euclidean(x,y) * (X + Y - 1/2 * minXY)
+    elseif I==3
+        return 1/252 * minXY^(2I + 1) + 1/720 * minXY^(I+1) * euclidean(x,y) * (5*max(X,Y)^2 + 2*X*Y + 3 * minXY^2)
+    else
+        error("Invalid I")
+    end
+end
```
willtebbutt, line 58: Maybe re-write this to use dispatch, as in the other PR?

sharanry: On it. Thanks!

sharanry:
@willtebbutt Do you think this is ready to be merged?

src/basekernels/wiener.jl
```diff
@@ -0,0 +1,170 @@
...
+        i = 3, ai = 252, bi = 1/720, ri(x, y) = 5 * max(x, y)² + 2 * x * z + 3 * min(x, y)²
+```
+
+**References:**\\
+See the paper *Probabilistic ODE Solvers with Runge-Kutta Means* by Schober, Duvenaud and Hennig, NIPS, 2014, for more details.
+
+"""
+struct WienerKernel{I} <: BaseKernel
+    function WienerKernel{I}() where I
+        I in (-1, 0, 1, 2, 3) || error("Invalid paramter i=$(I). Should be -1, 0, 1, 2 or 3.")
```
theogf, line 31: You can replace this by an `@assert` no? Also there is a typo on "Invalid parameter"

theogf: `error()` will only send an error message but not throw an error

src/basekernels/wiener.jl
```diff
@@ -0,0 +1,170 @@
...
+    k₋₁(x, y) =  δ(x, y)
+    i >= 0, kᵢ(x, y) = 1 / ai * min(x, y)^(2i + 1) + bi * min(x, y)^(i + 1) * |x - y| * ri(x, y),
+    with the coefficients ai, bi and the residual ri(x, y) defined as follows:
+        i = 0, ai =   1, bi = 0
+        i = 1, ai =   3, bi = 1/  2, ri(x, y) = 1
+        i = 2, ai =  20, bi = 1/ 12, ri(x, y) = x + y - 1 / 2 * min(x, y)
+        i = 3, ai = 252, bi = 1/720, ri(x, y) = 5 * max(x, y)² + 2 * x * z + 3 * min(x, y)²
+```
+
+**References:**\\
```
devmotion, line 25:
```suggestion
# References:

```

src/basekernels/wiener.jl
```diff
@@ -0,0 +1,170 @@
+"""
+    WienerKernel{i}()
+
+i-times integrated Wiener process kernel function given by
+```julia
+    κ(x, y) =  kᵢ(x, y)
+```
```
devmotion, line 7: This is not really useful, isn't it?

src/basekernels/wiener.jl
devmotion: This should be fixed in other parts of the docstring as well.

src/basekernels/wiener.jl
sharanry: Not really. I was trying to keep it uniform w.r.t other docstrings.

src/basekernels/wiener.jl
```diff
@@ -0,0 +1,170 @@
...
+    Y = sqrt(sum(abs2.(y)))
+    minXY = min(X, Y)
+    return 1 / 252 * minXY^7 + 1 / 720 * minXY^4 * euclidean(x, y) *
+        ( 5 * max(X, Y)^2 + 2 * X * Y + 3 * minXY^2 )
+end
+
+function kappa(κ::WienerKernel, x, y)
+    return _wiener(κ, x, y)
+end
+
```
devmotion, line 74: We removed `kappa` with three arguments, and switched to implementing only `kappa(kernel, d)` where `d` is the result of a binary operation (often a distance) of `x` and `y`. For all other kernels, that don't follow this pattern, just implement `(k::MyKernelType)(x, y)`, where `x` and `y` is one data point. Kernel matrices are constructed only with `kernelmatrix` and `kernelmatrix!`.
So for this kernel we should just not define `kappa`:
```suggestion
```

src/basekernels/wiener.jl
```diff
@@ -0,0 +1,169 @@
...
+function _kernel(
+    κ::WienerKernel,
+    x::AbstractVector,
+    y::AbstractVector;
+    obsdim::Int = defaultobs
+)
+    @assert length(x) == length(y) "x and y don't have the same dimension!"
+    return kappa(κ, x, y)
+end
```
devmotion, line 85: `_kernel` was removed completely:
```suggestion
```

src/basekernels/wiener.jl
```diff
@@ -0,0 +1,169 @@
...
+
+function kernelmatrix(
+    κ::WienerKernel,
+    X::AbstractMatrix;
+    obsdim::Int = defaultobs
+)
+    return kernelmatrix(κ, X, X; obsdim=obsdim)
+end
+
+Base.show(io::IO, κ::WienerKernel{I}) where I = print(io, "Wiener Kernel $(I)-times integrated")
```
devmotion, line 169:
```suggestion
Base.show(io::IO, κ::WienerKernel{I}) where I = print(io, "Wiener Kernel ", I, "-times integrated")
```

src/basekernels/wiener.jl
```diff
@@ -0,0 +1,169 @@
...
+    WienerKernel{i}()
+
+i-times integrated Wiener process kernel function.
+
+For i=-1, this is just the white noise covariance, see WhiteKernel.\\
+For i= 0, this is the Wiener process covariance,\\
+For i= 1, this is the integrated Wiener process covariance (velocity),\\
+For i= 2, this is the twice-integrated Wiener process covariance (accel.),\\
+For i= 3, this is the thrice-integrated Wiener process covariance,\\
+
```
devmotion, line 11: There's no need to include `\\` here.

src/basekernels/wiener.jl
```diff
@@ -0,0 +1,169 @@
+"""
+    WienerKernel{i}()
+
+i-times integrated Wiener process kernel function.
+
+For i=-1, this is just the white noise covariance, see WhiteKernel.\\
```
devmotion, line 6:
```suggestion
For i=-1, this is just the white noise covariance, see [WhiteKernel](@ref).
```

src/basekernels/wiener.jl
```diff
@@ -0,0 +1,169 @@
...
+For i= 0, this is the Wiener process covariance,\\
+For i= 1, this is the integrated Wiener process covariance (velocity),\\
+For i= 2, this is the twice-integrated Wiener process covariance (accel.),\\
+For i= 3, this is the thrice-integrated Wiener process covariance,\\
+
+where `κᵢ` is given by
+
+```julia
```
devmotion, line 14: This is not actually a Julia source code block. Mathematical expressions should be put in ````math` blocks, and text should be separated.

src/basekernels/wiener.jl
```diff
@@ -0,0 +1,169 @@
...
+end
+
+function kernelmatrix(
+    κ::WienerKernel,
+    X::AbstractMatrix;
+    obsdim::Int = defaultobs
+)
+    return kernelmatrix(κ, X, X; obsdim=obsdim)
+end
+
```
devmotion, line 168: I assume (untested) that the implementation of `kernelmatrix` and `kernelmatrix!` can be removed completely. It should be covered by the generic implementation that now is able to deal with kernels that do not use any metrics as well. It should just make use of the definition of `(k::MyKernelType)(x, y)` where `x` and `y` are data points.

devmotion:
```suggestion
```

src/basekernels/wiener.jl
devmotion: As a reference: https://docs.julialang.org/en/v1/manual/performance-tips/#Avoid-string-interpolation-for-I/O-1

src/basekernels/wiener.jl
sharanry: Problem with using `math` is that the expressions do not have any syntax highlighting and becomes very unreadable when we display this in REPL using `?WienerKernel`. Used `julia` just for the sake better readability of the expressions due to color coding.

devmotion: IMO we should use `math` it is latex/math-like syntax and `julia` if it is proper Julia code.

src/basekernels/wiener.jl
sharanry: Some tests seem to fail when I remove them. I am looking into it.

src/basekernels/wiener.jl
sharanry: Everything come on the same line when I remove them. :/
sc

src/basekernels/wiener.jl
```diff
@@ -0,0 +1,155 @@
...
+function _wiener(κ::WienerKernel{3}, x, y)
+    X = sqrt(sum(abs2.(x)))
+    Y = sqrt(sum(abs2.(y)))
+    minXY = min(X, Y)
+    return 1 / 252 * minXY^7 + 1 / 720 * minXY^4 * euclidean(x, y) *
+        ( 5 * max(X, Y)^2 + 2 * X * Y + 3 * minXY^2 )
+end
+
+(κ::WienerKernel)(x::Real, y::Real) = wiener(κ, x, y)
```
devmotion, line 71: IMO this can be simplified to
```suggestion
function (::WienerKernel{0})(x, y)
    X = sqrt(sum(abs2, x))
    Y = sqrt(sum(abs2, y))
    return min(X, Y)
end

function (::WienerKernel{1})(x, y)
    X = sqrt(sum(abs2, x))
    Y = sqrt(sum(abs2, y))
    minXY = min(X, Y)
    return 1 / 3 * minXY^3 + 1 / 2 * minXY^2 * euclidean(x, y)
end

function (::WienerKernel{2})(x, y)
    X = sqrt(sum(abs2, x))
    Y = sqrt(sum(abs2, y))
    minXY = min(X, Y)
    return 1 / 20 * minXY^5 + 1 / 12 * minXY^3 * euclidean(x, y) *
        ( X + Y - 1 / 2 * minXY )
end

function (::WienerKernel{3})(x, y)
    X = sqrt(sum(abs2, x))
    Y = sqrt(sum(abs2, y))
    minXY = min(X, Y)
    return 1 / 252 * minXY^7 + 1 / 720 * minXY^4 * euclidean(x, y) *
        ( 5 * max(X, Y)^2 + 2 * X * Y + 3 * minXY^2 )
end
```
Maybe that also helps in removing the `kernelmatrix` implementations.

src/basekernels/wiener.jl
devmotion: You have to include empty lines in the markdown source code if you want to have new lines. However, IMO it would be better to use a list than trying to separate things by line breaks.

src/basekernels/wiener.jl
```diff
@@ -0,0 +1,71 @@
...
+
+```math
+    κ₋₁(x, y) =  δ(x, y)
+    i >= 0, κᵢ(x, y) = 1 / ai * min(x, y)^(2i + 1) + bi * min(x, y)^(i + 1) * |x - y| * ri(x, y),
+    with the coefficients ai, bi and the residual ri(x, y) defined as follows:
+        i = 0, ai =   1, bi = 0
+        i = 1, ai =   3, bi = 1/  2, ri(x, y) = 1
+        i = 2, ai =  20, bi = 1/ 12, ri(x, y) = x + y - 1 / 2 * min(x, y)
+        i = 3, ai = 252, bi = 1/720, ri(x, y) = 5 * max(x, y)² + 2 * x * z + 3 * min(x, y)²
+```
```
devmotion, line 22: I suggest to use something like (I wasn't able to show the formatting on Github without the incorrect backticks, they should be changed)
```
\``` math
    κ₋₁(x, y) =  δ(x, y)
\```
and for ``i ≥ 0``
\```math
    κᵢ(x, y) = min(x, y)^(2 * i + 1) / aᵢ + bᵢ * min(x, y)^(i + 1) * |x - y| * rᵢ(x, y),
\```
with the coefficients ``aᵢ``, ``bᵢ`` and the residual ``rᵢ(x, y)`` defined as follows:
\```math
a₀ = 1, b₀ = 0,
a₁ = 3, b₁ = 1/2, r₁(x, y) = 1,
a₂ = 20, b₂ = 1/12, r₂(x, y) = x + y - min(x, y) / 2,
a₃ = 252, b₃ = 1/720, r₃(x, y) = 5 * max(x, y)² + 2 * x * z + 3 * min(x, y)²
\```
```
to make it easier to read. Note that the definition of r₀ is missing.

sharanry: Thanks. This looks much better!

sharanry: Also, `r₀` needn't be defined as b₀ is 0. The second term is reduced to 0 because of this.

devmotion: Ah true :+1: Since the kernel is so simle for i = 0, maybe it would make sense to write it down explicitly, just like for i = -1?

src/basekernels/wiener.jl
```diff
@@ -0,0 +1,77 @@
...
+
+function (::WienerKernel{3})(x, y)
+    X = sqrt(sum(abs2, x))
+    Y = sqrt(sum(abs2, y))
+    minXY = min(X, Y)
+    return 1 / 252 * minXY^7 + 1 / 720 * minXY^4 * euclidean(x, y) *
+        ( 5 * max(X, Y)^2 + 2 * X * Y + 3 * minXY^2 )
+end
+
+Base.show(io::IO, κ::WienerKernel{I}) where I = print(io, "Wiener Kernel ", I, "-times integrated")
```
devmotion, line 77: Maybe
```suggestion
Base.show(io::IO, ::WienerKernel{I}) where I = print(io, I, "-times integrated Wiener kernel")
```
is a bit easier to read.

src/basekernels/wiener.jl
```diff
@@ -0,0 +1,77 @@
...
+    κᵢ(x, y) = 1 / aᵢ * min(x, y)^(2i + 1) + bᵢ * min(x, y)^(i + 1) * |x - y| * rᵢ(x, y),
+\```
+    with the coefficients `aᵢ`, `bᵢ` and the residual `rᵢ(x, y)` defined as follows:
+\```math
+    a₀ = 1, b₀ = 0,
+    a₁ = 3, b₁ = 1/2, r₁(x, y) = 1,
+    a₂ = 20, b₂ = 1/12, r₂(x, y) = x + y - min(x, y) / 2,
+    a₃ = 252, b₃ = 1/720, r₃(x, y) = 5 * max(x, y)² + 2 * x * z + 3 * min(x, y)²
+
+\```
```
devmotion, line 28: Ah I think my comment was a bit confusing, I meant you should use ` ``` ` instead of ` \``` `. The ` `` ` were actually intended, they should be used for inline math.

test/basekernels/wiener.jl
```diff
@@ -0,0 +1,40 @@
...
+    @test typeof(k3) <: WienerKernel{3}
+
+    @test_throws AssertionError WienerKernel(i=4)
+    @test_throws AssertionError WienerKernel(i=-2)
+
+    # kernelmatrix tests
+    m1 = rand(3,4)
+    m2 = rand(3,4)
+    @test kernelmatrix(k0, m1, m1) ≈ kernelmatrix(k0, m1) atol=1e-5
+    @test_broken kernelmatrix(k0, m1, m2) ≈ k0(m1, m2) atol=1e-5
```
devmotion, line 24: `k0(m1, m2)` is not supposed to work, kernels can only be applied to single data points (i.e., vectors of real numbers or real numbers but not matrices).
```suggestion
```

devmotion: Can you add a test for `kernel(x, y)` where `x` and `y` are vectors? And maybe test different `i`?

sharanry: > k0(m1, m2) is not supposed to work, kernels can only be applied to single data points (i.e., vectors of real numbers or real numbers but not matrices).

Previously, `(k::SomeKernel)(x, y) = kernelmatrix(k, x, y) `, so thats when this test was written.
Shouldn't `k(m1,m2)` at least give an error? It return a real number right now, which is very misleading.

devmotion: > Shouldn't k(m1,m2) at least give an error? It return a real number right now, which is very misleading.

IMO it returns the correct and reasonable answer in this case: with the current implementation, it returns the kernel evaluated on the two data points `m1` and `m2`, interpreting `m1` and `m2` as two reshaped vectors. Actually the current implementation would work for arrays of arbitrary dimension. Errors could only be defined for every kernel separately by constraining the arguments for `k(x, y)` (e.g., to only `Real` or `AbstractVector{<:Real}`). However, IMO one should usually avoid this since otherwise users can't use arbitrary number or array types or even just generic objects for which `sum(abs2, x)` and `euclidean(x, y)` is defined.

test/basekernels/wiener.jl
```diff
@@ -1,4 +1,6 @@
 @testset "wiener" begin
+    using Distances
+
```
devmotion, line 3: Packages should only be imported in `tests/runtests.jl` (which already loads Distances):
```suggestion
```

sharanry: Thanks! Forgot to remove it :sweat_smile:

sharanry:
@devmotion @theogf @willtebbutt Can this be merged?

src/basekernels/wiener.jl
```diff
@@ -0,0 +1,77 @@
...
+    with the coefficients ``aᵢ``, ``bᵢ`` and the residual ``rᵢ(x, y)`` defined as follows:
+```math
+    a₁ = 3, b₁ = 1/2, r₁(x, y) = 1,
+    a₂ = 20, b₂ = 1/12, r₂(x, y) = x + y - min(x, y) / 2,
+    a₃ = 252, b₃ = 1/720, r₃(x, y) = 5 * max(x, y)² + 2 * x * z + 3 * min(x, y)²
+
+```
+
+# References:
+See the paper *Probabilistic ODE Solvers with Runge-Kutta Means* by Schober, Duvenaud and Hennig, NIPS, 2014, for more details.
```
willtebbutt, line 31: Line length :)

PR #79, opened 2020-04-10
Add `kappa` to stationary Gabor kernel

sharanry:
Since Gabor kernel is a Stationary and and Isotropic kernel, I thought it would be a good idea to add `kappa` of the form `kappa(κ::GaborKernel, d::Real)`. Missed it earlier in #52 .


theogf:
Could you just change the `basekernel` path? Then we can merge!

src/kernels/gabor.jl
```diff
@@ -76,3 +80,6 @@ function kerneldiagmatrix(
     obsdim::Int=defaultobs) #TODO Add test
     kerneldiagmatrix(κ.kernel, X; obsdim=obsdim)
 end
+
+Base.show(io::IO, κ::GaborKernel) = print(io, "Gabor Kernel (ell = $(κ.ell), p = $(κ.p))")
+
```
devmotion, line 85:
```suggestion
Base.show(io::IO, κ::GaborKernel) = print(io, "Gabor Kernel (ell = ", κ.ell, ", p = ", κ.p, ")")
```

theogf: I corrected it in #85

test/kernels/gabor.jl
```diff
@@ -5,10 +5,12 @@
     k = GaborKernel(ell=ell, p=p)
     @test k.ell ≈ ell atol=1e-5
     @test k.p ≈ p atol=1e-5
-    @test kappa(k,v1,v2) ≈ exp(-sqeuclidean(v1,v2) ./(k.ell.^2))*cospi(euclidean(v1,v2)./ k.p) atol=1e-5
-    @test kappa(k,v1,v2) ≈ kappa(transform(SqExponentialKernel(), 1/k.ell),v1,v2)*kappa(transform(CosineKernel(), 1/k.p), v1,v2) atol=1e-5
+    @test kappa(k, v1, v2) ≈ exp(-sqeuclidean(v1, v2) ./ (k.ell.^2))*cospi(euclidean(v1, v2) ./ k.p) atol=1e-5
+    @test kappa(k, v1, v2) ≈ kappa(transform(SqExponentialKernel(), 1 / k.ell),v1, v2) * kappa(transform(CosineKernel(), 1 / k.p), v1, v2) atol=1e-5
```
willtebbutt, line 9: Sorry, while you're here would you mind sorting out the length of these lines?

src/basekernels/gabor.jl
```diff
@@ -51,9 +51,13 @@ function Base.getproperty(k::GaborKernel, v::Symbol)
     end
 end
 
-Base.show(io::IO, κ::GaborKernel) = print(io, "Gabor Kernel (ell = ", κ.ell, ", p = ", κ.p, ")")
```
devmotion, line 54: This should not be removed.

src/basekernels/gabor.jl
```diff
@@ -76,3 +80,6 @@ function kerneldiagmatrix(
     obsdim::Int=defaultobs) #TODO Add test
     kerneldiagmatrix(κ.kernel, X; obsdim=obsdim)
 end
+
+Base.show(io::IO, κ::GaborKernel) = print(io, "Gabor Kernel (ell = $(κ.ell), p = $(κ.p))")
+
```
devmotion, line 85: This is not needed, @theogf already added it in another commit to the master branch (see my comment above).

sharanry: Yeah thanks! Adding it back.

sharanry:
@theogf @devmotion Is it ready to merge?

src/basekernels/gabor.jl
```diff
@@ -51,9 +51,13 @@ function Base.getproperty(k::GaborKernel, v::Symbol)
     end
 end
 
-Base.show(io::IO, κ::GaborKernel) = print(io, "Gabor Kernel (ell = ", κ.ell, ", p = ", κ.p, ")")
+function kappa(κ::GaborKernel, d::Real)
+    return kappa(κ.kernel.kernels[1], d^2) * kappa(κ.kernel.kernels[2], d)
+end
+
+kappa(κ::GaborKernel, x, y) = kappa(κ.kernel, x, y)
```
devmotion, line 58: Do we still need this? Can't we just use the generic implementation in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/89baf9c7ebaf813a11ec24f69f15c7e021d30943/src/generic.jl#L7 now that a metric is defined for the Gabor kernel?
```suggestion
```

sharanry: I was wondering why without this the test `kappa(k, x, y) == kappa(k, evaluate(KernelFunctions.metric(k), x, y))` was failing. There was an error in the `kappa(::GaborKernel, d::Real)`. 

Apparently, `kappa(k::TransformedKernel, d) = kappa(k.kernel, d)`, so it doesn't do the transformation before applying the kernel using distance measure. I think this is a bug. Not sure.

sharanry: Present fix is to modify the kappa to give the intended output of gabor kernel
```julia
function kappa(κ::GaborKernel, d::Real)
    return kappa(κ.kernel.kernels[1], (d/κ.ell)^2) * kappa(κ.kernel.kernels[2], d/κ.p)
end
```

devmotion: That's not something we want to do since `ell` and `p` do not have to be scalar valued. It seems it just doesn't make sense to define `kappa(::GaborKernel, ::Real)` and `metric(::GaborKernel)`, and we should just keep the old implementation of forwarding `kappa` to the underlying kernel? If we want to do any optimization based on the metric, it should be done in `KernelProd` (see the PR for its refactoring).

devmotion:
Unfortunately, IMO this PR should be closed since it is not guaranteed that parameters `p` and `ell` are scalar valued (see https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/79#discussion_r410408616).

willtebbutt:
To be honest, I'm a bit confused as to why we have this implemented as a kernel at all. It seems like something that would be best implemented by a helper function in terms of `Prod`. Am I missing something?

devmotion:
I agree, I'm still not fully convinced that we should have a separate type. We had a discussion about this when it was added (see https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/52#issuecomment-604640049 and the comments below). IIRC one argument for having a dedicated type was that otherwise the parameters of the Gabor kernel can't be accessed easily.

sharanry:
Yes accessing the Gabor's parameters was the only issue if it were just a simple function returning a `KernelProduct`. This would not affect making the kernel trainable in any way. If the parameters are not really necessary to be accessed, I can't seem to think of any other good reason to keep a separate type.

sharanry:
I don't think this PR has any value anymore. Gabor kernel is a `KernelProduct` now. `kappa` is still cannot be defined for it because it is not defined for `KernelProduct`. Should we consider refining it?

theogf:
I guess we should close this PR because kappa only make sense for `SimpleKernel`

Issue #43, opened 2020-03-11
obsdim


willtebbutt:
This relates to #9 and presumably #10 .

The `obsdim` keyword is a bit annoying as it has to be propagated around the entire codebase.

Arguably a better solution for vector-valued inputs is to introduce  `ColVecs <: AbstractVector` and `RowVecs <: AbstractVector` types, as done in Stheno, so that it's clear how things are being treated, and functions like `kernelmatrix` that, in the context of non-`BaseKernel`s don't need to know about this information, don't have to worry about it.

@theogf @devmotion any thoughts?

edit: this obviously generalises to any input type that you fancy from an API perspective.

theogf:
I tend to like `obsdim` because it is also used in `Distances.jl` and `MLDataUtils.jl` for example. To be honest when I see `RowVecs` or `ColVecs` I intuitively have no idea what is what...

willtebbutt:
That's fair I guess.

When you see a `RowVecs`, you should intuitively think "this is an efficiently-represented vector of row-vectors" and with `ColVecs` "this is an efficiently-represented vector of column-vectors". Is there something unclear about that? I would genuinely like to know if I'm the only one who think this is a really helpful way to think about this stuff...

devmotion:
I'm not a fan of carrying around `obsdim` everywhere. It tries to solve a problem that only exists since observations in matrixform can be viewed differently, so the main problem is not a conceptual/mathematical one (and does not even exist for other input spaces, e.g., when dealing with distributions) but only caused by how we store the inputs.

I'm wondering if one actually has to support both column- and row-vectors (e.g., it is really necessary to provide explicit support for row vectors in Distances, couldn't the user just provide the lazy transpose of the matrix of observations in this case?). Would it make sense to only use column vectors if one is mainly/only interested in operations based on full observations and not individual components of observations?

[...]

PR #81, opened 2020-04-15
Add general TensorProduct kernel

devmotion:
See https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/80#issuecomment-614003196.

As discussed in the outdated PR https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/56#issue-393484992, it is still unclear what type of arguments `kernelmatrix` and `kerneldiagmatrix` should take, or how to implement this in a general way to support both data types (i.e., collections of joint observations in all spaces (e.g., of type `Vector{Tuple{Vector{Float64},Vector{Int}}}`), or collections of collections of observations in each space (e.g., of type `Tuple{Matrix{Float64},Matrix{Int}}`).

### DIFF
--- a/src/kernels/tensorproduct.jl
+++ b/src/kernels/tensorproduct.jl
@@ -40,14 +40,16 @@ function kernelmatrix!(
 
     featuredim = feature_dim(obsdim)
     if !check_dims(K, X, X, featuredim, obsdim)
-        throw(DimensionMismatch("Dimensions of the target array K $(size(K)) are not consistent with X $(size(X))"))
+        throw(DimensionMismatch("Dimensions of the target array K $(size(K)) are not " *
+                                "consistent with X $(size(X))"))
     end
 
     size(X, featuredim) == length(kernel) ||
         error("number of kernels and groups of features are not consistent")
 
-    kernelmatrix!(K, kernel.kernels[1], selectdim(X, featuredim, 1))
-    for (k, Xi) in Iterators.drop(zip(kernel.kernels, eachslice(X; dims = featuredim)), 1)
+    kernels_and_input = zip(kernel.kernels, eachslice(X; dims = featuredim))
+    kernelmatrix!(K, first(kernels_and_input)...)
+    for (k, Xi) in Iterators.drop(kernels_and_input, 1)
         K .*= kernelmatrix(k, Xi)
     end
 
@@ -65,14 +67,15 @@ function kernelmatrix!(
 
     featuredim = feature_dim(obsdim)
     if !check_dims(K, X, Y, featuredim, obsdim)
-        throw(DimensionMismatch("Dimensions $(size(K)) of the target array K are not consistent with X ($(size(X))) and Y ($(size(Y)))"))
+        throw(DimensionMismatch("Dimensions $(size(K)) of the target array K are not " *
+                                "consistent with X ($(size(X))) and Y ($(size(Y)))"))
     end
 
     size(X, featuredim) == length(kernel) ||
         error("number of kernels and groups of features are not consistent")
 
     kernels_and_input = zip(
-        zip(kernel.kernels,
+        kernel.kernels,
         eachslice(X; dims = featuredim),
         eachslice(Y; dims = featuredim),
     )
@@ -95,7 +98,8 @@ function kernelmatrix(
 
     featuredim = feature_dim(obsdim)
     if !check_dims(X, X, featuredim, obsdim)
-        throw(DimensionMismatch("Dimensions of the target array K $(size(K)) are not consistent with X $(size(X))"))
+        throw(DimensionMismatch("Dimensions of the target array K $(size(K)) are not " *
+                                "consistent with X $(size(X))"))
     end
 
     size(X, featuredim) == length(kernel) ||
@@ -113,11 +117,12 @@ function kernelmatrix(
     Y::AbstractMatrix;
     obsdim::Int = defaultobs
 )
-    @assert obsdim ∈ (1, 2) || error("obsdim should be 1 or 2 (see docs of kernelmatrix))")
+    obsdim ∈ (1, 2) || error("obsdim should be 1 or 2 (see docs of kernelmatrix))")
 
     featuredim = feature_dim(obsdim)
     if !check_dims(X, Y, featuredim, obsdim)
-        throw(DimensionMismatch("Dimensions $(size(K)) of the target array K are not consistent with X ($(size(X))) and Y ($(size(Y)))"))
+        throw(DimensionMismatch("Dimensions $(size(K)) of the target array K are not " *
+                                "consistent with X ($(size(X))) and Y ($(size(Y)))"))
     end
 
     size(X, featuredim) == length(kernel) ||
@@ -141,15 +146,17 @@ function kerneldiagmatrix!(
 )
     obsdim ∈ (1, 2) || error("obsdim should be 1 or 2 (see docs of kernelmatrix))")
     if length(K) != size(X, obsdim)
-        throw(DimensionMismatch("Dimensions of the target array K $(size(K)) are not consistent with X $(size(X))"))
+        throw(DimensionMismatch("Dimensions of the target array K $(size(K)) are not " *
+                                "consistent with X $(size(X))"))
     end
 
     featuredim = feature_dim(obsdim)
     size(X, featuredim) == length(kernel) ||
         error("number of kernels and groups of features are not consistent")
 
-    kerneldiagmatrix!(K, kernel.kernels[1], selectdim(X, featuredim, 1))
-    for (k, Xi) in Iterators.drop(zip(kernel.kernels, eachslice(X; dims = featuredim)), 1)
+    kernels_and_input = zip(kernel.kernels, eachslice(X; dims = featuredim))
+    kerneldiagmatrix!(K, first(kernels_and_input)...)
+    for (k, Xi) in Iterators.drop(kernels_and_input, 1)
         K .*= kerneldiagmatrix(k, Xi)
     end
 
--- a/src/kernels/tensorproduct.jl
+++ b/src/kernels/tensorproduct.jl
@@ -25,7 +25,7 @@ function kernelmatrix!(
     K::AbstractMatrix,
     kernel::TensorProduct,
     X::AbstractMatrix;
-    obsdim::Int = defaultobs
+    obsdim::Int = defaultobs,
 )
     obsdim ∈ (1, 2) || "obsdim should be 1 or 2 (see docs of kernelmatrix))"
 
@@ -50,7 +50,7 @@ function kernelmatrix!(
     kernel::TensorProduct,
     X::AbstractMatrix,
     Y::AbstractMatrix;
-    obsdim::Int = defaultobs
+    obsdim::Int = defaultobs,
 )
     obsdim ∈ (1, 2) || error("obsdim should be 1 or 2 (see docs of kernelmatrix))")
 
@@ -62,11 +62,13 @@ function kernelmatrix!(
     size(X, featuredim) == length(kernel) ||
         error("number of kernels and groups of features are not consistent")
 
-    kernelmatrix!(K, kernel.kernels[1], selectdim(X, featuredim, 1),
-                  selectdim(Y, featuredim, 1))
-    for (k, Xi, Yi) in Iterators.drop(zip(kernel.kernels,
-                                          eachslice(X; dims = featuredim),
-                                          eachslice(Y; dims = featuredim)), 1)
+    kernels_and_input = zip(
+        zip(kernel.kernels,
+        eachslice(X; dims = featuredim),
+        eachslice(Y; dims = featuredim),
+    )
+    kernelmatrix!(K, first(kernels_and_inputs)...)
+    for (k, Xi, Yi) in Iterators.drop(kernels_and_inputs, 1)
         K .*= kernelmatrix(k, Xi, Yi)
     end
 
@@ -78,7 +80,7 @@ end
 function kernelmatrix(
     kernel::TensorProduct,
     X::AbstractMatrix;
-    obsdim::Int = defaultobs
+    obsdim::Int = defaultobs,
 )
     obsdim ∈ (1, 2) || error("obsdim should be 1 or 2 (see docs of kernelmatrix))")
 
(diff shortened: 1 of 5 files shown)


willtebbutt:
It would be good to get a basic implementation off the ground -- so it should be fine to make this a `Matrix` for now, inline with the rest of the package. In the future when we refactor the inputs types, as per the discussion in #43 , we could consider having some custom input type that plays nicely with this type of kernel. eg. an internal representation comprising a `Vector` or `Tuple` comprising `D` `Vector`s, that looks like a `Vector` of `N` `Vector`s from the outside.

theogf:
For multiple dimensions one can play with `KernelProduct` and `SelectTransform`

theogf:
Realising that now we have quite a collection of kernels which will only work with `kappa(k, x, y)` shouldn't we create generic `kernelmatrix` and `kerneldiagmatrix` for these kernels (we can create a trait), where we iterate over all columns/rows?

willtebbutt:
I would be up for that generally, but I think in this case we're best off implementing `kernelmatrix` and `kerneldiagmatrix` directly, since it can be done in terms of other calls to `kernelmatrix` and `kerneldiagmatrix`, and that is what is going to be most efficient in general.

theogf:
Ok I will take care of this tomorrow then. This could also be a good start for tackling the "my data is not a matrix" problem

devmotion:
> In an initial implementation, we could just assume 1 dimension per group.

So you suggest implementing `kernelmatrix` and `kerneldiagmatrix` for `Matrix`-like inputs, where columns (or rows, depending on `obsdim`) with `n` elements represent observations of `n` groups that are handled by `n` different kernels?

> I think in this case we're best off implementing kernelmatrix and kerneldiagmatrix directly, since it can be done in terms of other calls to kernelmatrix and kerneldiagmatrix

Intuitively, I would say in the setting described above it's more efficient to fill the matrix by evaluating `kappa(::TensorProduct, obs_i, obs_j)` for each pair of observations (exploiting symmetry as well, I guess) instead of multiplying together `n` different matrices?

willtebbutt:
> So you suggest implementing kernelmatrix and kerneldiagmatrix for Matrix-like inputs, where columns (or rows, depending on obsdim) with n elements represent observations of n groups that are handled by n different kernels?

Yeah, although I think in my notation I was thinking D groups of inputs with D different kernels.

> Intuitively, I would say in the setting described above it's more efficient to fill the matrix by evaluating kappa(::TensorProduct, obs_i, obs_j) for each pair of observations (exploiting symmetry as well, I guess) instead of multiplying together n different matrices?

I can see your point, but it's not generally the case that any one of the D kernel matrices will be most efficiently computed in this manner. For example, if one of the D groups has P-dimensional inputs and the Exponentiated Quadratic kernel, the most efficient thing is generally going to be to make a call out to Distances.jl to compute the matrix, then product it, since the naive implementation of the kernel matrix in that case is really bad.

I think the high level point is that any given kernel knows best how to execute `kernelmatrix` and `kerneldiagmatrix`, so if you don't use their implementation it's quite possible that you'll be stuck with sub-optimal behaviour.

devmotion:
I completely agree that a general implementation should just call `kernelmatrix` and `kerneldiagmatrix` of the individual kernels. But in the case of matrices as inputs `P` always has to be 1, hasn't it? I got confused since we were talking about implementing this special case with this very special structure, but at the same time the discussion of `kernelmatrix` and `kerneldiagmatrix` seemed to suggest that one should still use the more general but in this case maybe less efficient implementation.

willtebbutt:
That's a fair point, my previous argument does really only apply to P > 1.

I think that there are other reason though, including code complexity and ease of doing reverse-mode AD. The former is in the eye of the beholder of course, and the latter is hard to assess without having a decent suite of benchmarks.

devmotion:
I ended up using `mapreduce` for `kernelmatrix` and `kerneldiagmatrix`. I'm not completely happy with the implementation of `kernelmatrix!` and `kerneldiagmatrix!`, somehow the use of `Iterators.drop` feels unsatisfying. It's basically just a more explicit version of the implementation of the out-of-place variants in which in the initial step the output array is filled with `kernelmatrix!` and `kerneldiagmatrix!`.

I also had to add some methods for constructing kernel matrices (and their diagonal) from vector-valued inputs (it felt silly to reshape just the slices in tensorproduct.jl, IMO this should work for other kernels as well).

src/kernels/tensorproduct.jl
```diff
@@ -0,0 +1,175 @@
...
+function kernelmatrix(
+    kernel::TensorProduct,
+    X::AbstractMatrix;
+    obsdim::Int = defaultobs
+)
+    obsdim ∈ (1, 2) || error("obsdim should be 1 or 2 (see docs of kernelmatrix))")
+
+    featuredim = feature_dim(obsdim)
+    if !check_dims(X, X, featuredim, obsdim)
+        throw(DimensionMismatch("Dimensions of the target array K $(size(K)) are not consistent with X $(size(X))"))
```
willtebbutt, line 87: Not within 92 char lim. Please wrap string over multiple lines

src/kernels/tensorproduct.jl
```diff
@@ -0,0 +1,175 @@
...
+
+    return K
+end
+
+# mapreduce with multiple iterators requires Julia 1.2 or later.
+
+function kernelmatrix(
+    kernel::TensorProduct,
+    X::AbstractMatrix;
+    obsdim::Int = defaultobs
```
willtebbutt, line 81:
```suggestion
    obsdim::Int = defaultobs,
```

src/kernels/tensorproduct.jl
```diff
@@ -0,0 +1,175 @@
...
+
+    return K
+end
+
+function kernelmatrix!(
+    K::AbstractMatrix,
+    kernel::TensorProduct,
+    X::AbstractMatrix,
+    Y::AbstractMatrix;
+    obsdim::Int = defaultobs
```
willtebbutt, line 53:
```suggestion
    obsdim::Int = defaultobs,
```

src/kernels/tensorproduct.jl
```diff
@@ -0,0 +1,175 @@
...
+end
+
+# TODO: General implementation of `kernelmatrix` and `kerneldiagmatrix`
+# Default implementation assumes 1D observations
+
+function kernelmatrix!(
+    K::AbstractMatrix,
+    kernel::TensorProduct,
+    X::AbstractMatrix;
+    obsdim::Int = defaultobs
```
willtebbutt, line 28:
```suggestion
    obsdim::Int = defaultobs,
```

src/kernels/tensorproduct.jl
```diff
@@ -0,0 +1,175 @@
...
+    size(X, featuredim) == length(kernel) ||
+        error("number of kernels and groups of features are not consistent")
+
+    kernelmatrix!(K, kernel.kernels[1], selectdim(X, featuredim, 1),
+                  selectdim(Y, featuredim, 1))
+    for (k, Xi, Yi) in Iterators.drop(zip(kernel.kernels,
+                                          eachslice(X; dims = featuredim),
+                                          eachslice(Y; dims = featuredim)), 1)
+        K .*= kernelmatrix(k, Xi, Yi)
+    end
```
willtebbutt, line 71:
```suggestion
    kernels_and_input = zip(
        zip(kernel.kernels,
        eachslice(X; dims = featuredim),
        eachslice(Y; dims = featuredim),
    )
    kernelmatrix!(K, first(kernels_and_inputs)...)
    for (k, Xi, Yi) in Iterators.drop(kernels_and_inputs, 1)
        K .*= kernelmatrix(k, Xi, Yi)
    end
```

src/kernels/tensorproduct.jl
```diff
@@ -0,0 +1,175 @@
...
+
+    size(X, featuredim) == length(kernel) ||
+        error("number of kernels and groups of features are not consistent")
+
+    return mapreduce((x, y) -> x .* y,
+                     zip(kernel.kernels,
+                         eachslice(X; dims = featuredim),
+                         eachslice(Y; dims = featuredim))) do (k, Xi, Yi)
+        kernelmatrix(k, Xi, Yi)
+    end
```
willtebbutt, line 120:
```suggestion
    kernels_and_inputs = zip(
        kernel.kernels,
        eachslice(X; dims = featuredim),
        eachslice(Y; dims = featuredim),
    )
    return mapreduce((x, y) -> x .* y, kernels_and_inputs) do (k, Xi, Yi)
        kernelmatrix(k, Xi, Yi)
    end
```

src/kernels/tensorproduct.jl
```diff
@@ -0,0 +1,175 @@
...
+
+function kerneldiagmatrix!(
+    K::AbstractVector,
+    kernel::TensorProduct,
+    X::AbstractMatrix;
+    obsdim::Int = defaultobs
+)
+    obsdim ∈ (1, 2) || error("obsdim should be 1 or 2 (see docs of kernelmatrix))")
+    if length(K) != size(X, obsdim)
+        throw(DimensionMismatch("Dimensions of the target array K $(size(K)) are not consistent with X $(size(X))"))
```
willtebbutt, line 131: line length

src/kernels/tensorproduct.jl
```diff
@@ -0,0 +1,175 @@
...
+    obsdim ∈ (1,2) || error("obsdim should be 1 or 2 (see docs of kernelmatrix))")
+
+    featuredim = feature_dim(obsdim)
+    size(X, featuredim) == length(kernel) ||
+        error("number of kernels and groups of features are not consistent")
+
+    return mapreduce((x, y) -> x .* y,
+                     zip(kernel.kernels, eachslice(X; dims = featuredim))) do (k, Xi)
+         kerneldiagmatrix(k, Xi)
+    end
```
willtebbutt, line 160:
```suggestion
    kernels_and_inputs = zip(kernel.kernels, eachslice(X; dims = featuredim))
    return mapreduce((x, y) -> x .* y, kernels_and_inputs) do (k, Xi)
        kerneldiagmatrix(k, Xi)
    end
```

src/matrix/kernelmatrix.jl
```diff
@@ -1,11 +1,19 @@
...
 
 In-place version of [`kernelmatrix`](@ref) where pre-allocated matrix `K` will be overwritten with the kernel matrix.
 """
 kernelmatrix!
 
+function kernelmatrix!(
+    K::AbstractMatrix,
+    kernel::Kernel,
+    X::AbstractVector;
+    obsdim::Int = defaultobs
```
willtebbutt, line 13:
```suggestion
    obsdim::Int = defaultobs,
```

src/matrix/kernelmatrix.jl
```diff
@@ -23,6 +31,16 @@ end
 kernelmatrix!(K::AbstractMatrix, κ::TransformedKernel, X::AbstractMatrix; obsdim::Int = defaultobs) =
         kernelmatrix!(K, kernel(κ), apply(κ.transform, X, obsdim = obsdim), obsdim = obsdim)
 
+function kernelmatrix!(
+    K::AbstractMatrix,
+    kernel::Kernel,
+    X::AbstractVector,
+    Y::AbstractVector;
+    obsdim::Int = defaultobs
```
willtebbutt, line 39:
```suggestion
    obsdim::Int = defaultobs,
```

src/matrix/kernelmatrix.jl
```diff
@@ -88,6 +106,15 @@ end
 kernelmatrix(κ::TransformedKernel, X::AbstractMatrix; obsdim::Int = defaultobs) =
         kernelmatrix(kernel(κ), apply(κ.transform, X, obsdim = obsdim), obsdim = obsdim)
 
+function kernelmatrix(
+    kernel::Kernel,
+    X::AbstractVector{<:Real},
+    Y::AbstractVector{<:Real};
+    obsdim::Int = defaultobs
```
willtebbutt, line 113:
```suggestion
    obsdim::Int = defaultobs,
```

src/matrix/kernelmatrix.jl
```diff
@@ -107,12 +134,20 @@ kernelmatrix(κ::TransformedKernel, X::AbstractMatrix, Y::AbstractMatrix; obsdim
...
+    kerneldiagmatrix(κ::Kernel, X; obsdim::Int = 2)
 
 Calculate the diagonal matrix of `X` with respect to kernel `κ`
 `obsdim = 1` means the matrix `X` has size #samples x #dimension
 `obsdim = 2` means the matrix `X` has size #dimension x #samples
 """
+function kerneldiagmatrix(
+    kernel::Kernel,
+    X::AbstractVector;
+    obsdim::Int = defaultobs
```
willtebbutt, line 146:
```suggestion
    obsdim::Int = defaultobs,
```

src/matrix/kernelmatrix.jl
```diff
@@ -127,10 +162,19 @@ function kerneldiagmatrix(
...
-    kerneldiagmatrix!(K::AbstractVector,κ::Kernel, X::Matrix; obsdim::Int = 2)
+    kerneldiagmatrix!(K::AbstractVector, κ::Kernel, X; obsdim::Int = 2)
 
 In place version of [`kerneldiagmatrix`](@ref)
 """
+function kerneldiagmatrix!(
+    K::AbstractVector,
+    kernel::Kernel,
+    X::AbstractVector;
+    obsdim::Int = defaultobs
```
willtebbutt, line 173:
```suggestion
    obsdim::Int = defaultobs,
```

src/kernels/tensorproduct.jl
```diff
@@ -0,0 +1,175 @@
+"""
+    TensorProduct(kernels...)
+
+Create a tensor product of kernels.
```
willtebbutt, line 4: More information about what the kernel computes would be great here.

Also some docstrings showing how to construct + suggestions for when to use a vector vs a tuple.

src/kernels/tensorproduct.jl
devmotion: Yeah, I just copied this from `kernelmatrix.jl` (which is not great either) and missed that it doesn't follow the style guide.

src/kernels/tensorproduct.jl
willtebbutt: Thanks for addressing this

willtebbutt:
Happy for you to merge when you're happy @devmotion , nice work.

devmotion:
Just want to make sure that tests pass. Otherwise I'm happy.

devmotion:
OK, tests pass (test failures on Julia master are unrelated).

PR #83, opened 2020-04-16
[WIP] Rework on kernelmatrix to work with Vectors and more complex kernels

theogf:
I  added the notion of `SimpleKernel` for when `Distances.jl` pairwise can just be used.
While `BaseKernel` will just rely on `kappa(k, x, y)`
@willtebbutt Now I can see the point of `ColVecs`, is that okay if I steal your implementation from Stheno and rework it for our purpose?

[EDIT] This PR only aims at making the difference between `SimpleKernel` where `kernelmatrix` is based on `pairwise` from `Distances.jl`, `BaseKernel` where `kernelmatrix` relies on `kappa(k, x, y)` and `Kernel` which also contains composite kernels.
`ColVecs` and `RowVecs` are added separately in #84 and #89

### DIFF
--- a/src/KernelFunctions.jl
+++ b/src/KernelFunctions.jl
@@ -10,6 +10,7 @@ export duplicate, set! # Helpers
 
 export Kernel
 export ConstantKernel, WhiteKernel, EyeKernel, ZeroKernel
+export CosineKernel
 export SqExponentialKernel, RBFKernel, GaussianKernel, SEKernel
 export LaplacianKernel, ExponentialKernel, GammaExponentialKernel
 export ExponentiatedKernel
--- a/src/KernelFunctions.jl
+++ b/src/KernelFunctions.jl
@@ -41,6 +41,7 @@ Abstract type defining a slice-wise transformation on an input matrix
 abstract type Transform end
 abstract type Kernel end
 abstract type BaseKernel <: Kernel end
+abstract type SimpleKernel <: BaseKernel end
 
 include("utils.jl")
 include("distances/dotproduct.jl")
--- a/src/basekernels/constant.jl
+++ b/src/basekernels/constant.jl
@@ -7,7 +7,7 @@ Create a kernel that always returning zero
 ```
 The output type depends of `x` and `y`
 """
-struct ZeroKernel <: BaseKernel end
+struct ZeroKernel <: SimpleKernel end
 
 kappa(κ::ZeroKernel, d::T) where {T<:Real} = zero(T)
 
@@ -15,6 +15,7 @@ metric(::ZeroKernel) = Delta()
 
 Base.show(io::IO, ::ZeroKernel) = print(io, "Zero Kernel")
 
+issimple(ZeroKernel) = true
 
 """
     WhiteKernel()
@@ -24,7 +25,7 @@ Base.show(io::IO, ::ZeroKernel) = print(io, "Zero Kernel")
 ```
 Kernel function working as an equivalent to add white noise. Can also be called via `EyeKernel()`
 """
-struct WhiteKernel <: BaseKernel end
+struct WhiteKernel <: SimpleKernel end
 
 """
     EyeKernel()
@@ -48,7 +49,7 @@ Kernel function always returning a constant value `c`
     κ(x,y) = c
 ```
 """
-struct ConstantKernel{Tc<:Real} <: BaseKernel
+struct ConstantKernel{Tc<:Real} <: SimpleKernel
     c::Vector{Tc}
     function ConstantKernel(;c::T=1.0) where {T<:Real}
         new{T}([c])
--- a/src/basekernels/constant.jl
+++ b/src/basekernels/constant.jl
@@ -15,7 +15,6 @@ metric(::ZeroKernel) = Delta()
 
 Base.show(io::IO, ::ZeroKernel) = print(io, "Zero Kernel")
 
-issimple(ZeroKernel) = true
 
 """
     WhiteKernel()
--- a/src/basekernels/piecewisepolynomial.jl
+++ b/src/basekernels/piecewisepolynomial.jl
@@ -43,16 +43,6 @@ function kappa(
     return _piecewisepolynomial(κ, r, j)
 end
 
-function _kernel(
-    κ::PiecewisePolynomialKernel,
-    x::AbstractVector,
-    y::AbstractVector;
-    obsdim::Int = defaultobs,
-)
-    @assert length(x) == length(y) "x and y don't have the same dimension!"
-    return kappa(κ,x,y)
-end
-
 function kernelmatrix(
     κ::PiecewisePolynomialKernel{V},
     X::AbstractMatrix;
--- a/src/basekernels/piecewisepolynomial.jl
+++ b/src/basekernels/piecewisepolynomial.jl
@@ -52,6 +52,17 @@ function kernelmatrix(
     return map(r->_piecewisepolynomial(κ, r, j), pairwise(metric(κ), X; dims=obsdim))
 end
 
+function kernelmatrix(
+    κ::PiecewisePolynomialKernel{V},
+    X::AbstractMatrix,
+    Y::AbstractMatrix;
+    obsdim::Int = defaultobs
+) where {V}
+    j = div(size(X, feature_dim(obsdim)), 2) + V + 1
+    return map(r->_piecewisepolynomial(κ, r, j), pairwise(metric(κ), X, Y; dims=obsdim))
+end
+
+
 function _kernelmatrix(κ::PiecewisePolynomialKernel{V}, X, Y, obsdim) where {V}
     j = div(size(X, feature_dim(obsdim)), 2) + V + 1
     return map(r->_piecewisepolynomial(κ, r, j), pairwise(metric(κ), X, Y; dims=obsdim))
--- a/src/basekernels/piecewisepolynomial.jl
+++ b/src/basekernels/piecewisepolynomial.jl
@@ -12,7 +12,7 @@ where `r` is the Mahalanobis distance mahalanobis(x,y) with `maha` as the metric
 """
 struct PiecewisePolynomialKernel{V, A<:AbstractMatrix{<:Real}} <: SimpleKernel
     maha::A
-    j::Int64
+    j::Int
     function PiecewisePolynomialKernel{V}(maha::AbstractMatrix{<:Real}) where V
         V in (0, 1, 2, 3) || error("Invalid paramter v=$(V). Should be 0, 1, 2 or 3.")
         LinearAlgebra.checksquare(maha)
@@ -31,11 +31,7 @@ _f(κ::PiecewisePolynomialKernel{2}, r, j) = 1 + (j + 2) * r + (j^2 + 4 * j + 3)
 _f(κ::PiecewisePolynomialKernel{3}, r, j) = 1 + (j + 3) * r +
     (6 * j^2 + 36j + 45) / 15 * r.^2 + (j^3 + 9 * j^2 + 23j + 15) / 15 * r.^3
 
-function _piecewisepolynomial(κ::PiecewisePolynomialKernel{V}, r, j) where V
-    return max(1 - r, 0)^(j + V) * _f(κ, r, j)
-end
-
-kappa(κ::PiecewisePolynomialKernel, d::Real) = _piecewisepolynomial(κ, r, κ.j)
+kappa(κ::PiecewisePolynomialKernel{V}, r) where V = max(1 - r, 0)^(κ.j + V) * _f(κ, r, κ.j)
 
 metric(κ::PiecewisePolynomialKernel) = Mahalanobis(κ.maha)
 
--- a/src/basekernels/piecewisepolynomial.jl
+++ b/src/basekernels/piecewisepolynomial.jl
@@ -10,12 +10,14 @@ processes are hence v times  mean-square differentiable. The kernel function is:
 where `r` is the Mahalanobis distance mahalanobis(x,y) with `maha` as the metric.
 
 """
-struct PiecewisePolynomialKernel{V, A<:AbstractMatrix{<:Real}} <: BaseKernel
+struct PiecewisePolynomialKernel{V, A<:AbstractMatrix{<:Real}} <: SimpleKernel
     maha::A
+    j::Int64
     function PiecewisePolynomialKernel{V}(maha::AbstractMatrix{<:Real}) where V
         V in (0, 1, 2, 3) || error("Invalid paramter v=$(V). Should be 0, 1, 2 or 3.")
         LinearAlgebra.checksquare(maha)
-        return new{V,typeof(maha)}(maha)
+        j = div(size(maha, 1), 2) + V + 1
+        return new{V,typeof(maha)}(maha, j)
     end
 end
 
@@ -33,75 +35,7 @@ function _piecewisepolynomial(κ::PiecewisePolynomialKernel{V}, r, j) where V
     return max(1 - r, 0)^(j + V) * _f(κ, r, j)
 end
 
-function kappa(
-    κ::PiecewisePolynomialKernel{V},
-    x::AbstractVector{<:Real},
-    y::AbstractVector{<:Real},
-) where {V}
-    r = evaluate(metric(κ), x, y)
-    j = div(size(x, 2), 1) + V + 1
-    return _piecewisepolynomial(κ, r, j)
-end
-
-function kernelmatrix(
-    κ::PiecewisePolynomialKernel{V},
-    X::AbstractMatrix;
-    obsdim::Int = defaultobs
-) where {V}
-    j = div(size(X, feature_dim(obsdim)), 2) + V + 1
-    return map(r->_piecewisepolynomial(κ, r, j), pairwise(metric(κ), X; dims=obsdim))
-end
-
-function kernelmatrix(
-    κ::PiecewisePolynomialKernel{V},
-    X::AbstractMatrix,
-    Y::AbstractMatrix;
-    obsdim::Int = defaultobs
-) where {V}
-    j = div(size(X, feature_dim(obsdim)), 2) + V + 1
-    return map(r->_piecewisepolynomial(κ, r, j), pairwise(metric(κ), X, Y; dims=obsdim))
-end
-
-
-function _kernelmatrix(κ::PiecewisePolynomialKernel{V}, X, Y, obsdim) where {V}
-    j = div(size(X, feature_dim(obsdim)), 2) + V + 1
-    return map(r->_piecewisepolynomial(κ, r, j), pairwise(metric(κ), X, Y; dims=obsdim))
-end
-
-function kernelmatrix!(
-    K::AbstractMatrix,
-    κ::PiecewisePolynomialKernel{V},
-    X::AbstractMatrix;
-    obsdim::Int = defaultobs
-) where {V}
-    @assert obsdim ∈ [1,2] "obsdim should be 1 or 2 (see docs of kernelmatrix))"
-    if !check_dims(K, X, X, feature_dim(obsdim), obsdim)
-        throw(DimensionMismatch(
-            "Dimensions of the target array K $(size(K)) are not consistent with X " *
-            "$(size(X))",
-        ))
-    end
-    j = div(size(X, feature_dim(obsdim)), 2) + V + 1
-    return map!(r->_piecewisepolynomial(κ,r,j), K, pairwise(metric(κ), X; dims=obsdim))
-end
-
-function kernelmatrix!(
-    K::AbstractMatrix,
-    κ::PiecewisePolynomialKernel{V},
-    X::AbstractMatrix,
-    Y::AbstractMatrix;
-    obsdim::Int = defaultobs,
-) where {V}
-    @assert obsdim ∈ [1,2] "obsdim should be 1 or 2 (see docs of kernelmatrix))"
-    if !check_dims(K, X, Y, feature_dim(obsdim), obsdim)
-        throw(DimensionMismatch(
-            "Dimensions $(size(K)) of the target array K are not consistent with X " *
-            "($(size(X))) and Y ($(size(Y)))",
-        ))
-    end
-    j = div(size(X, feature_dim(obsdim)), 2) + V + 1
-    return map!(r->_piecewisepolynomial(κ,r,j), K, pairwise(metric(κ), X, Y; dims=obsdim))
-end
+kappa(κ::PiecewisePolynomialKernel, d::Real) = _piecewisepolynomial(κ, r, κ.j)
 
 metric(κ::PiecewisePolynomialKernel) = Mahalanobis(κ.maha)
 
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -22,7 +22,7 @@ end
 
 for k in concretetypes(Kernel, [])
     @eval begin
-        @inline (κ::$k)(x::AbstractVector{<:Real}, y::AbstractVector{<:Real}) = kappa(κ, x, y)
+        @inline (κ::$k)(x, y) = kappa(κ, x, y)
         @inline (κ::$k)(X::AbstractMatrix{T}, Y::AbstractMatrix{T}; obsdim::Integer=defaultobs) where {T} = kernelmatrix(κ, X, Y, obsdim=obsdim)
         @inline (κ::$k)(X::AbstractMatrix{T}; obsdim::Integer=defaultobs) where {T} = kernelmatrix(κ, X, obsdim=obsdim)
     end
--- a/src/kernels/kernelproduct.jl
+++ b/src/kernels/kernelproduct.jl
@@ -29,23 +29,26 @@ hadamard(x,y) = x.*y
 function kernelmatrix(
     κ::KernelProduct,
     X::AbstractMatrix;
-    obsdim::Int=defaultobs)
-    reduce(hadamard,kernelmatrix(κ.kernels[i],X,obsdim=obsdim) for i in 1:length(κ))
+    obsdim::Int=defaultobs,
+)
+    reduce(hadamard, kernelmatrix(κ.kernels[i], X, obsdim = obsdim) for i in 1:length(κ))
 end
 
 function kernelmatrix(
     κ::KernelProduct,
     X::AbstractMatrix,
     Y::AbstractMatrix;
-    obsdim::Int=defaultobs)
-    reduce(hadamard,_kernelmatrix(κ.kernels[i],X,Y,obsdim) for i in 1:length(κ))
+    obsdim::Int=defaultobs,
+)
+    reduce(hadamard, kernelmatrix(κ.kernels[i], X, Y, obsdim = obsdim) for i in 1:length(κ))
 end
 
 function kerneldiagmatrix(
     κ::KernelProduct,
     X::AbstractMatrix;
-    obsdim::Int=defaultobs) #TODO Add test
-    reduce(hadamard,kerneldiagmatrix(κ.kernels[i],X,obsdim=obsdim) for i in 1:length(κ))
+    obsdim::Int=defaultobs,
+) #TODO Add test
+    reduce(hadamard, kerneldiagmatrix(κ.kernels[i], X, obsdim = obsdim) for i in 1:length(κ))
 end
 
 function Base.show(io::IO, κ::KernelProduct)
--- a/src/kernels/scaledkernel.jl
+++ b/src/kernels/scaledkernel.jl
@@ -15,6 +15,8 @@ end
 
 kappa(k::ScaledKernel, x) = first(k.σ²) * kappa(k.kernel, x)
 
+kappa(k::ScaledKernel, x, y) = first(k.σ²) * kappa(k.kernel, x, y)
+
 metric(k::ScaledKernel) = metric(k.kernel)
 
 Base.:*(w::Real, k::Kernel) = ScaledKernel(k, w)
--- a/src/kernels/transformedkernel.jl
+++ b/src/kernels/transformedkernel.jl
@@ -39,3 +39,17 @@ function printshifted(io::IO, κ::TransformedKernel, shift::Int)
     printshifted(io, κ.kernel, shift)
     print(io,"\n" * ("\t" ^ (shift + 1)) * "- $(κ.transform)")
 end
+
+# Kernel matrix operations
+
+kernelmatrix!(K::AbstractMatrix, κ::TransformedKernel, X::AbstractMatrix; obsdim::Int = defaultobs) =
+    kernelmatrix!(K, kernel(κ), apply(κ.transform, X, obsdim = obsdim), obsdim = obsdim)
+
+kernelmatrix!(K::AbstractMatrix, κ::TransformedKernel, X::AbstractMatrix, Y::AbstractMatrix; obsdim::Int = defaultobs) =
+    kernelmatrix!(K, kernel(κ), apply(κ.transform, X, obsdim = obsdim), apply(κ.transform, Y, obsdim = obsdim), obsdim = obsdim)
+
+kernelmatrix(κ::TransformedKernel, X::AbstractMatrix; obsdim::Int = defaultobs) =
+    kernelmatrix(kernel(κ), apply(κ.transform, X, obsdim = obsdim), obsdim = obsdim)
+    
+kernelmatrix(κ::TransformedKernel, X::AbstractMatrix, Y::AbstractMatrix; obsdim::Int = defaultobs) =
+    kernelmatrix(kernel(κ), apply(κ.transform, X, obsdim = obsdim), apply(κ.transform, Y, obsdim = obsdim), obsdim = obsdim)
--- a/src/matrix/kernelmatrix.jl
+++ b/src/matrix/kernelmatrix.jl
@@ -12,7 +12,7 @@ function kernelmatrix!(
     κ::Kernel,
     X::AbstractVector{<:Real}
 )
-    kernelmatrix!(K, κ, reshape(X, 1, :), obsdim = 2)
+    kernelmatrix!(K, κ, ColVecs(reshape(X, 1, :)))
 end
 
 function kernelmatrix!(
@@ -47,7 +47,7 @@ function kernelmatrix!(
     κ::Kernel,
     X::AbstractVector
     )
-    if (size(K, 1) != size(K, 2)) || (length(X) != size(K, 1))
+    if !check_dims(K, X, X)
         throw(DimensionMismatch("Dimensions of the target array K $(size(K)) are not consistent with X $(size(X))"))
     end
     K .= κ.(X, X')
@@ -60,7 +60,7 @@ function kernelmatrix!(
     X::AbstractVector{<:Real},
     Y::AbstractVector{<:Real}
 )
-    kernelmatrix!(K, κ, reshape(X, 1, :), reshape(Y, 1, :), obsdim = 2)
+    kernelmatrix!(K, κ, ColVecs(reshape(X, 1, :)), ColVecs(reshape(Y, 1, :)))
 end
 
 function kernelmatrix!(
@@ -98,7 +98,7 @@ function kernelmatrix!(
     X::AbstractVector,
     Y::AbstractVector
     )
-    if (size(K, 1) != length(X)) || (size(K, 2) != length(Y))
+    if !check_dims(K, X, Y)
         throw(DimensionMismatch("Dimensions of the target array K $(size(K)) are not consistent with X $(size(X)) and Y $(size(Y))"))
     end
     K .= κ.(X, Y')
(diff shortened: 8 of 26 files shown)


devmotion:
I feel like this might be a good case for using traits instead of a `SimpleKernel` type.

theogf:
I really don't think traits are needed here. It would force users who wants to add "simple" kernels to always add a `issimple(x)` for their kernels whereas the inheritance comes in very naturally. A `SimpleKernel` is just a `BaseKernel` where we have an optimized way of computing the kernel matrix.

devmotion:
I guess users might want to user their own type structure for various reasons (e.g., kernels on special spaces that you might want to group but where not every kernel is a simple or base kernel), and it's just not possible for them to exploit the general implementation using `pairwise` in KernelFunctions if this requires `SimpleKernel`.

> It would force users who wants to add "simple" kernels to always add a issimple(x) for their kernels whereas the inheritance comes in very naturally.

You could get both by just defining
```julia
issimple(::Kernel) = false
issimple(::SimpleKernel) = true
```
or something similar with specific traits. So everyone that subtypes `SimpleKernel` would get it for free, and for custom types one could still hook into the pairwise computations by defining `issimple`.

willtebbutt:
> @willtebbutt Now I can see the point of ColVecs, is that okay if I steal your implementation from Stheno and rework it for our purpose?

Please do :)

If we go ahead with this change, I'm unclear what the differences between `SimpleKernel`, `BaseKernel`, and `Kernel` are going to be though.

At the minute, `BaseKernel` roughly means "anything that's not a composite kernel" and by default uses the map-kappa–over-metric implementation for ease of implementation. What will `SimpleKernel` add to this, and what will it mean for a `Kernel` to be a `BaseKernel` but not a `SimpleKernel`?

theogf:
`SimpleKernel` is using the `pairwise` strategy that we have from the beginning. The latest commit should be more explicit

willtebbutt:
Sorry, is it critical that we do the `ColVecs` implementation in this PR? Would be much easier to review as a separate PR.

willtebbutt:
> SimpleKernel is using the pairwise strategy that we have from the beginning. The latest commit should be more explicit

Yup, but then why do we have `BaseKernel`?

theogf:
Because KernelProduct etc are not BaseKernels

theogf:
> Sorry, is it critical that we do the `ColVecs` implementation in this PR? Would be much easier to review as a separate PR.

I can put an expensive `collect(eachcol(X))` with a note for now

willtebbutt:
> I can put an expensive collect(eachcol(X)) with a note for now

Okay, sorry, I think I've been getting myself confused by the purpose of this PR, but I still think it would be good to introduce the changes over the course of two PRs.

1. Change from `AbstractMatrix` to `AbstractVector`, including adding the `ColVecs` type.
2. Introduce the `SimpleKernel` abstraction.

AFAICT, the first is going to involve some fairly large changes to the codebase, so it would be good to do it as a single PR. Would you be okay with doing that @theogf ?

theogf:
The second one implies that we need to be able to treat the matrices as vectors of vectors, hence the need for `ColVecs`. But you are right the description of the PR is now wrong.

willtebbutt:
Yup, so if we do 1 then 2 everything should be okay?

theogf:
So as you mentioned in #84 and #43, time to tackle obsdim!
This what I aimed at with this PR by changing `kernelmatrix` (I will take the case of `kernelmatrix(k, X)` but of course this generalizes to `kernelmatrix(k, X, Y)`
 - X is a Vector of Real, we reshape X into a matrix and call `kernelmatrix` on it
 - X is a Matrix (of Real), we need the obsdim indicator to know if the observation are rowwise or colwise. I know that for me I always have my observation rowwise (obsdim = 1).
   - If `k` is a `SimpleKernel` we can call `pairwise` from `Distances.jl` and pass the `obsdim` argument
   - if `k` is a `BaseKernel` we transform the matrix in a vec of vec via `ColVecs`
- X is a Vector of Stuff, we iteratively call `kappa(k, x, y)` on each pair without passing by pairwise. No `obsdim` needed

willtebbutt:
I'm a little confused. I was expecting the following to happen:

- `obsdim` is removed (almost) everywhere
- All input types are changed from`AbstractMatrix`s to `AbstractVector`s
- (Maybe) helper functionality added at the top level to let people to continue to use the `obsdim` functionality if that's really what they prefer, to transform an input into either a `ColVecs` or `RowVecs` depending on the choice of `obsdim`, which can then be forwarded on to the rest of the package.

Why aren't we doing that?

theogf:
> All input types are changed fromAbstractMatrixs to AbstractVectors 

I don't understand the point of using a ColVecs wrapper when it's not necessary (for AbstractMatrix with SimpleKernels).

> obsdim is removed (almost) everywhere 

 If I understand you correctly, the problem there is with `obsdim` is that we have to pass it to all the `Transform` objects right? Then we could just agree that the convention for the Transform  is that we always have `obsdim = 2` (so we remove the need for it), and when `kernelmatrix` is called with `obsdim = 1`, we just send the transpose!

willtebbutt:
> If I understand you correctly, the problem there is with obsdim is that we have to pass it to all the Transform objects right?

No. The problem is that you ever have to pass an `obsdim`-like parameter to anything anywhere in the package (excluding `Distances`, but that's beyond our control). It makes everything harder to read and can be abstracted away completely! But we should take this discussion back to the relevant issue since we don't seem to have agreed on what the right solution is yet.

Since I've misunderstood, could you explain to me why you need `ColVecs` in this PR?

theogf:
For `BaseKernel` the only generic function possible to compute the function only directly on the objects (vectors). Therefore I needed to transform the input matrix into a vectors of vectors. Hence the need for `ColVecs`. 

But as you stated in #43 we can make a uniform interface. I will just finish this PR so we can move on and get rid of `obsdim` in the internals

theogf:
@willtebbutt @devmotion It is ready to be reviewed. All the substantial part is in `kernelmatrix.jl`
I am not super happy with `kernelmatrix!(K, k, x::AbstractVector)` written with `K .= k.(x, x')`but `broadcast!` does not improve the performance.

src/basekernels/constant.jl
```diff
@@ -7,14 +7,15 @@ Create a kernel that always returning zero
...
-struct ZeroKernel <: BaseKernel end
+struct ZeroKernel <: SimpleKernel end
 
 kappa(κ::ZeroKernel, d::T) where {T<:Real} = zero(T)
 
 metric(::ZeroKernel) = Delta()
 
 Base.show(io::IO, ::ZeroKernel) = print(io, "Zero Kernel")
 
+issimple(ZeroKernel) = true
```
devmotion, line 18: I guess this shouldn't be here?
```suggestion
```

src/generic.jl
```diff
@@ -22,7 +22,7 @@ end
 
 for k in concretetypes(Kernel, [])
     @eval begin
-        @inline (κ::$k)(x::AbstractVector{<:Real}, y::AbstractVector{<:Real}) = kappa(κ, x, y)
+        @inline (κ::$k)(x, y) = kappa(κ, x, y)
```
devmotion, line 25: I guess the different implementations of `kappa(k, x, y)` will be removed in a separate PR?

src/KernelFunctions.jl
```diff
@@ -43,6 +44,7 @@ Abstract type defining a slice-wise transformation on an input matrix
 abstract type Transform end
 abstract type Kernel end
 abstract type BaseKernel <: Kernel end
+abstract type SimpleKernel <: BaseKernel end
```
devmotion, line 47: Maybe `DistancesKernel` would highlight more clearly that for these kernels we make use of the `pairwise` functionality in Distances? SimpleKernel and BaseKernel both sound very similar and are both not very descriptive names IMO.

src/basekernels/piecewisepolynomial.jl
```diff
@@ -62,6 +52,17 @@ function kernelmatrix(
...
 end
 
+function kernelmatrix(
+    κ::PiecewisePolynomialKernel{V},
+    X::AbstractMatrix,
+    Y::AbstractMatrix;
+    obsdim::Int = defaultobs
+) where {V}
+    j = div(size(X, feature_dim(obsdim)), 2) + V + 1
+    return map(r->_piecewisepolynomial(κ, r, j), pairwise(metric(κ), X, Y; dims=obsdim))
```
devmotion, line 62: Why is PiecewisePolynomialKernel not a SimpleKernel? It seems it also uses pairwise?

src/basekernels/constant.jl
theogf: Oh thanks for noticing. It was from when I wanted to implement simple a trait.

src/generic.jl
theogf: Yes this is the plan. The proposition from #87 will be in a different PR

src/basekernels/piecewisepolynomial.jl
theogf: It needs to compute `j` which depends directly of `X`

src/KernelFunctions.jl
theogf: I am not sure if it would be better but I was thinking we could change it as : 
- `Kernel` -> `AbstractKernel`
- `BaseKernel` -> `Kernel`
- `SimpleKernel` -> `SimpleKernel`/`BaseKernel`

src/matrix/kernelmatrix.jl
```diff
@@ -6,76 +6,103 @@ In-place version of [`kernelmatrix`](@ref) where pre-allocated matrix `K` will b
...
 
+## Wrapper for vector of reals
 function kernelmatrix!(
     K::AbstractMatrix,
-    kernel::Kernel,
-    X::AbstractVector;
+    κ::Kernel,
+    X::AbstractVector{<:Real}
+)
+    kernelmatrix!(K, κ, reshape(X, 1, :), obsdim = 2)
```
devmotion, line 15: I thought we agreed not to do this? I guess more in line with the discussion we had would be to not import Distances.pairwise and define an internal method `pairwise` along the lines of
```julia
function pairwise(metric::PreMetric, X::AbstractMatrix; obsdim = defaultobs)
	return Distances.pairwise(metric, X; dims = obsdim)
end
function pairwise(metric::PreMetric, X::AbstractVector; obsdim = defaultobs)
	return Distances.pairwise(metric, reshape(X, 1, :); dims = 2)
end
```
and similarly for `pairwise(metric, X, Y, ...)` (and pairwise! maybe, not sure if that's useful?).

I guess then it would be easier to remove obsdim in one of the subsequent PRs and add support for ColVecs. It seems a bit weird to add this implementation with our plan in mind.

src/utils.jl
```diff
@@ -36,9 +37,11 @@ Base.getindex(D::ColVecs, i) = ColVecs(view(D.X, :, i))
 #     return T <: Real ? T : Float64
 # end
 
-check_dims(K,X,Y,featdim,obsdim) = check_dims(X,Y,featdim,obsdim) && (size(K) == (size(X,obsdim),size(Y,obsdim)))
+check_dims(K, X, Y, featdim, obsdim) =
+    check_dims(X, Y, featdim, obsdim) &&
+    (size(K) == (size(X, obsdim), size(Y, obsdim)))
```
devmotion, line 42:
```suggestion
function check_dims(K, X, Y, featdim, obsdim)
    return check_dims(X, Y, featdim, obsdim) && size(K) == (size(X, obsdim), size(Y, obsdim))
end
```

src/KernelFunctions.jl
devmotion: I guess `AbstractKernel` might be better and I guess `Kernel` would be fine instead of `BaseKernel` :+1: However, I think it would be nice to use something better than SimpleKernel or BaseKernel. At least I personally don't have an intuition for the names SimpleKernel or BaseKernel - is it referring to some mathematical property or form? That's why I think something more telling would be nice.

src/matrix/kernelmatrix.jl
theogf: Well right now it's not really an issue. If you want I can replace it by :
`kernelmatrix!(K, κ, ColVecs(reshape(X, 1, :)))`
This will be done anyway in the next PR.

src/utils.jl
theogf: I actually need to take proper care of this one for `ColVecs`

src/KernelFunctions.jl
theogf: Well the problem is that, to my knowledge, we make the separation only for computational reasons... So there is no clear name.

devmotion: Yes, that was the motivation for DistancesKernel since it might highlight the point that these kernels use Distances for computing the kernel matrix more explicitly. As mentioned before, an alternative would be to not encode this in the kernel type but just use a traits based system for separating kernels for computational reasons.

src/matrix/kernelmatrix.jl
devmotion: No, as far as I understand, the idea is to only use vectors so there is no reason to reshape and wrap it anymore?

src/utils.jl
devmotion: Oh my comment was just due to the style guide.

src/basekernels/piecewisepolynomial.jl
devmotion: But that feels quite inconsistent. The initial post in this PR states

> This PR only aims at making the difference between SimpleKernel where kernelmatrix is based on pairwise from Distances.jl, BaseKernel where kernelmatrix relies on kappa(k, x, y)

and for PiecewisePolynomialKernel kernelmatrix is apparently based on pairwise from Distances.jl, so it should be a SimpleKernel, shouldn't it? Otherwise it is even less clear how one would define a SimpleKernel, it seems? For this kernel we just apply a function in `map` that is dependent on the inputs, but we can just keep the special implementation even if it is a SimpleKernel.

src/matrix/kernelmatrix.jl
theogf: Oh now I understand! 
But it would raise the problem for `SimpleKernel` we would not use `pairwise` anymore. But I suppose that for 1D problem it will not make a big difference. 
It also raises the problem about using `(k::Kernel)(x,y)` for an abstract type. It's probably time to drop 1.0 ?

src/basekernels/piecewisepolynomial.jl
theogf: Oh actually `j` can be directly inferred from the `maha` object. So there is no need to compute it at every call.

src/basekernels/piecewisepolynomial.jl
```diff
@@ -10,12 +10,14 @@ processes are hence v times  mean-square differentiable. The kernel function is:
 where `r` is the Mahalanobis distance mahalanobis(x,y) with `maha` as the metric.
 
 """
-struct PiecewisePolynomialKernel{V, A<:AbstractMatrix{<:Real}} <: BaseKernel
+struct PiecewisePolynomialKernel{V, A<:AbstractMatrix{<:Real}} <: SimpleKernel
     maha::A
+    j::Int64
```
devmotion, line 15:
```suggestion
    j::Int
```
I'm also wondering if one could come up with a more descriptive name for `j`?

willtebbutt: Could we please leave this for a separate PR? This is already quite large.

src/basekernels/piecewisepolynomial.jl
```diff
@@ -33,75 +35,7 @@ function _piecewisepolynomial(κ::PiecewisePolynomialKernel{V}, r, j) where V
...
-    if !check_dims(K, X, Y, feature_dim(obsdim), obsdim)
-        throw(DimensionMismatch(
-            "Dimensions $(size(K)) of the target array K are not consistent with X " *
-            "($(size(X))) and Y ($(size(Y)))",
-        ))
-    end
-    j = div(size(X, feature_dim(obsdim)), 2) + V + 1
-    return map!(r->_piecewisepolynomial(κ,r,j), K, pairwise(metric(κ), X, Y; dims=obsdim))
-end
+kappa(κ::PiecewisePolynomialKernel, d::Real) = _piecewisepolynomial(κ, r, κ.j)
```
devmotion, line 38: One should just rename `_piecewisepolynomial` to `kappa` and remove the `j` argument (since it can be accessed by `kernel.j`). Same for `_f`.

src/kernels/transformedkernel.jl
```diff
@@ -39,3 +39,17 @@ function printshifted(io::IO, κ::TransformedKernel, shift::Int)
...
+    kernelmatrix!(K, kernel(κ), apply(κ.transform, X, obsdim = obsdim), obsdim = obsdim)
+
+kernelmatrix!(K::AbstractMatrix, κ::TransformedKernel, X::AbstractMatrix, Y::AbstractMatrix; obsdim::Int = defaultobs) =
+    kernelmatrix!(K, kernel(κ), apply(κ.transform, X, obsdim = obsdim), apply(κ.transform, Y, obsdim = obsdim), obsdim = obsdim)
+
+kernelmatrix(κ::TransformedKernel, X::AbstractMatrix; obsdim::Int = defaultobs) =
+    kernelmatrix(kernel(κ), apply(κ.transform, X, obsdim = obsdim), obsdim = obsdim)
+    
+kernelmatrix(κ::TransformedKernel, X::AbstractMatrix, Y::AbstractMatrix; obsdim::Int = defaultobs) =
+    kernelmatrix(kernel(κ), apply(κ.transform, X, obsdim = obsdim), apply(κ.transform, Y, obsdim = obsdim), obsdim = obsdim)
```
willtebbutt, line 55: Style (line length)

src/basekernels/piecewisepolynomial.jl
devmotion: Maybe rename `_f` to `_kappa`?

theogf: Agreed for `_piecewise...` but for `_f` it is actually easier to read it like this no?

src/matrix/kernelmatrix.jl
devmotion: Yes, we would - but the reshaping for Distances would only happen in the implementation of `pairwise` sketched above (since only Distances cares about it being a matrix instead of a vector).

src/basekernels/piecewisepolynomial.jl
devmotion: IMO we should at least make PiecewisePolynomialKernel a SimpleKernel (or whatever it is called), but all other optimizations might indeed be better to move to a separate PR.

src/basekernels/piecewisepolynomial.jl
willtebbutt: I agree with Theo. The `_f` function is sufficiently non-descript that I wouldn't want to call it `_kappa` as it doesn't tie up with what we usually define `kappa` to do.

src/basekernels/piecewisepolynomial.jl
willtebbutt: Agreed.

src/kernels/transformedkernel.jl
theogf: Maybe this can be done when `obsdim` is removed ?

src/basekernels/piecewisepolynomial.jl
theogf: The `Int64` -> `Int` is just formatting right?
For the name, I saw the denomination `extent` but it's not really helpful. Since it is an intermediary computed quantity it's probably better to leave it as it is.

src/kernels/transformedkernel.jl
willtebbutt: That's a fair point actually. Let's get this PR merged!

src/basekernels/piecewisepolynomial.jl
devmotion: No, that's not just formatting.

src/matrix/kernelmatrix.jl
theogf: :snail:

src/basekernels/piecewisepolynomial.jl
willtebbutt: To expand, whether or not `Int` is `Int64` or `Int32` depends whether your architecture is 32 bit or 64 bit. See [here](https://docs.julialang.org/en/v1/manual/integers-and-floating-point-numbers/#Integers-1) :)

src/basekernels/piecewisepolynomial.jl
devmotion: IMO if one removes the superfluous `j` input in `_f` it's reasonably close to what we usually define as `kappa`, it has the same input arguments and is only off by a constant factor. Anyways, it's clearly better to apply these changes in a separate PR (just noted these things while reviewing).

src/matrix/kernelmatrix.jl
```diff
@@ -8,74 +8,82 @@ kernelmatrix!
...
+    κ::SimpleKernel,
+    X::AbstractMatrix;
     obsdim::Int = defaultobs,
 )
-    return kernelmatrix!(K, kernel, reshape(X, 1, :); obsdim = 2)
+    @assert obsdim ∈ [1, 2] "obsdim should be 1 or 2 (see docs of `kernelmatrix`)"
+    if !check_dims(K, X, X, feature_dim(obsdim), obsdim)
+        throw(DimensionMismatch("Dimensions of the target array K $(size(K)) are not consistent with X $(size(X))"))
+    end
+    map!(x -> kappa(κ, x), K, pairwise(metric(κ), X, dims = obsdim))
```
devmotion, line 19: Again something for a separate PR (maybe) but why don't we use `pairwise!`?

src/matrix/kernelmatrix.jl
```diff
@@ -87,51 +95,52 @@ Calculate the kernel matrix of `X` (and `Y`) with respect to kernel `κ`.
...
+    X::AbstractMatrix,
+    Y::AbstractMatrix;
+    obsdim = defaultobs,
+)
+    @assert obsdim ∈ [1, 2] "obsdim should be 1 or 2 (see docs of kernelmatrix))"
+    if !check_dims(X, Y, feature_dim(obsdim))
         throw(DimensionMismatch("X $(size(X)) and Y $(size(Y)) do not have the same number of features on the dimension : $(feature_dim(obsdim))"))
     end
-    _kernelmatrix(κ,X,Y,obsdim)
+    _kernelmatrix(κ, X, Y, obsdim)
```
devmotion, line 130: `_kernelmatrix` appears only once, so just remove it?

src/matrix/kernelmatrix.jl
```diff
@@ -8,74 +8,82 @@ kernelmatrix!
...
+    κ::Kernel,
+    X::AbstractMatrix;
+    obsdim::Int = defaultobs
+)
+    @assert obsdim ∈ [1, 2] "obsdim should be 1 or 2 (see docs of `kernelmatrix`))"
+    if obsdim == 1
+        kernelmatrix!(K, κ, ColVecs(X'))
+    else
+        kernelmatrix!(K, κ, ColVecs(X))
+    end
```
devmotion, line 33: Now that we have `vec_of_vecs` and RowVecs this could be changed to
```suggestion
    kernelmatrix!(K, κ, vec_of_vecs(X; obsdim = obsdim))
```

src/matrix/kernelmatrix.jl
```diff
@@ -8,74 +8,82 @@ kernelmatrix!
...
+    X::AbstractMatrix,
+    Y::AbstractMatrix;
+    obsdim::Int = defaultobs
+)
+    @assert obsdim ∈ [1, 2] "obsdim should be 1 or 2 (see docs of `kernelmatrix`))"
+    if obsdim == 1
+        kernelmatrix!(K, κ, ColVecs(X'), ColVecs(Y'))
+    else
+        kernelmatrix!(K, κ, ColVecs(X), ColVecs(Y))
+    end
```
devmotion, line 73: Similarly, this could be written as
```suggestion
    kernelmatrix!(K, κ, vec_of_vecs(X; obsdim = obsdim), vec_of_vecs(Y; obsdim = obsdim))
```

src/matrix/kernelmatrix.jl
```diff
@@ -87,51 +95,52 @@ Calculate the kernel matrix of `X` (and `Y`) with respect to kernel `κ`.
...
-    obsdim::Int = defaultobs,
-)
-    return kernelmatrix(kernel, reshape(X, 1, :), reshape(Y, 1, :); obsdim = 2)
+function kernelmatrix(κ::Kernel, X::AbstractMatrix; obsdim::Int = defaultobs)
+    @assert obsdim ∈ [1, 2] "obsdim should be 1 or 2 (see docs of `kernelmatrix`))"
+    if obsdim == 1
+        kernelmatrix(κ, ColVecs(X'))
+    else
+        kernelmatrix(κ, ColVecs(X))
+    end
```
devmotion, line 117:
```suggestion
    return kernelmatrix(κ, vec_of_vecs(X; obsdim = obsdim))
```

theogf:
Oh yeah I could remove the `@assert` in kernelmatrix now!

src/utils.jl
```diff
@@ -10,7 +10,7 @@ macro check_args(K, param, cond, desc=string(cond))
 end
 
 function vec_of_vecs(X::AbstractMatrix; obsdim::Int = 2)
-    @assert obsdim ∈ (1, 2) "obsdim should be 1 or 2"
+    @assert obsdim ∈ (1, 2) "obsdim should be 1 or 2, see docs of kernelmatrix"
```
devmotion, line 13: IMO we should use `error` instead of `@assert` since the documentation says
```julia
 Warning
  │
  │  An assert might be disabled at various optimization levels. Assert should therefore only be used as a debugging tool
```
but that's again not part of this PR :smile:

theogf: Oh! I am definitely learning a lot on this PR :laughing:

src/matrix/kernelmatrix.jl
```diff
@@ -87,51 +86,39 @@ Calculate the kernel matrix of `X` (and `Y`) with respect to kernel `κ`.
...
 
-function kernelmatrix(
-        κ::Kernel,
-        X::AbstractVector{<:Real};
-        obsdim::Int=defaultobs
-        )
-        kernelmatrix(κ,reshape(X,1,:),obsdim=2)
+function kernelmatrix(κ::Kernel, X::AbstractVector)
+    kernelmatrix(κ, X, X) #TODO Can be optimized later
 end
```
willtebbutt, line 91:
```suggestion
kernelmatrix(κ::Kernel, X::AbstractVector) = kernelmatrix(κ, X, X)
```

src/matrix/kernelmatrix.jl
```diff
@@ -87,51 +86,39 @@ Calculate the kernel matrix of `X` (and `Y`) with respect to kernel `κ`.
...
 
-function kernelmatrix(
-        κ::Kernel,
-        X::AbstractMatrix;
-        obsdim::Int = defaultobs
-        )
-        K = map(x->kappa(κ,x),pairwise(metric(κ),X,dims=obsdim))
+function kernelmatrix(κ::Kernel, X::AbstractVector, Y::AbstractVector)
+    κ.(X, Y')
 end
```
willtebbutt, line 95:
```suggestion
kernelmatrix(κ::Kernel, X::AbstractVector, Y::AbstractVector) = κ.(X, Y')
```

src/matrix/kernelmatrix.jl
```diff
@@ -87,51 +86,39 @@ Calculate the kernel matrix of `X` (and `Y`) with respect to kernel `κ`.
...
 
-kernelmatrix(κ::TransformedKernel, X::AbstractMatrix; obsdim::Int = defaultobs) =
-        kernelmatrix(kernel(κ), apply(κ.transform, X, obsdim = obsdim), obsdim = obsdim)
+function kernelmatrix(κ::SimpleKernel, X::AbstractMatrix; obsdim::Int = defaultobs)
+    @assert obsdim ∈ [1, 2] "obsdim should be 1 or 2 (see docs of `kernelmatrix`))"
+    K = map(x -> kappa(κ, x), pairwise(metric(κ), X, dims = obsdim))
```
willtebbutt, line 99:
```suggestion
    return map(x -> kappa(κ, x), pairwise(metric(κ), X, dims = obsdim))
```

src/matrix/kernelmatrix.jl
```diff
@@ -87,51 +86,39 @@ Calculate the kernel matrix of `X` (and `Y`) with respect to kernel `κ`.
...
 
-function kernelmatrix(
-    kernel::Kernel,
-    X::AbstractVector{<:Real},
-    Y::AbstractVector{<:Real};
-    obsdim::Int = defaultobs,
-)
-    return kernelmatrix(kernel, reshape(X, 1, :), reshape(Y, 1, :); obsdim = 2)
+function kernelmatrix(κ::Kernel, X::AbstractMatrix; obsdim::Int = defaultobs)
+    kernelmatrix(κ, vec_of_vecs(X, obsdim = obsdim))
```
willtebbutt, line 103:
```suggestion
    return kernelmatrix(κ, vec_of_vecs(X, obsdim = obsdim))
```

src/matrix/kernelmatrix.jl
```diff
@@ -87,51 +86,39 @@ Calculate the kernel matrix of `X` (and `Y`) with respect to kernel `κ`.
...
+    X::AbstractMatrix,
+    Y::AbstractMatrix;
+    obsdim = defaultobs,
+)
+    @assert obsdim ∈ [1, 2] "obsdim should be 1 or 2 (see docs of kernelmatrix))"
+    if !check_dims(X, Y, feature_dim(obsdim))
         throw(DimensionMismatch("X $(size(X)) and Y $(size(Y)) do not have the same number of features on the dimension : $(feature_dim(obsdim))"))
     end
-    _kernelmatrix(κ,X,Y,obsdim)
+    map(x -> kappa(κ, x), pairwise(metric(κ), X, Y, dims = obsdim))
```
willtebbutt, line 116:
```suggestion
    return map(x -> kappa(κ, x), pairwise(metric(κ), X, Y, dims = obsdim))
```

src/matrix/kernelmatrix.jl
```diff
@@ -87,51 +86,39 @@ Calculate the kernel matrix of `X` (and `Y`) with respect to kernel `κ`.
...
 end
 
-@inline _kernelmatrix(κ::Kernel,X,Y,obsdim) = map(x->kappa(κ,x),pairwise(metric(κ),X,Y,dims=obsdim))
-
-kernelmatrix(κ::TransformedKernel, X::AbstractMatrix, Y::AbstractMatrix; obsdim::Int = defaultobs) =
-        kernelmatrix(kernel(κ), apply(κ.transform, X, obsdim = obsdim), apply(κ.transform, Y, obsdim = obsdim), obsdim = obsdim)
+function kernelmatrix(κ::Kernel, X::AbstractMatrix, Y::AbstractMatrix; obsdim::Int = defaultobs)
+    kernelmatrix(κ, vec_of_vecs(X, obsdim = obsdim), vec_of_vecs(Y, obsdim = obsdim))
```
willtebbutt, line 120:
```suggestion
    return kernelmatrix(κ, vec_of_vecs(X, obsdim = obsdim), vec_of_vecs(Y, obsdim = obsdim))
```

src/matrix/kernelmatrix.jl
```diff
@@ -140,25 +127,18 @@ Calculate the diagonal matrix of `X` with respect to kernel `κ`
...
-    kernel::Kernel,
-    X::AbstractVector;
-    obsdim::Int = defaultobs,
-)
-    return kerneldiagmatrix(kernel, reshape(X, 1, :); obsdim = 2)
+    κ::Kernel,
+    X::AbstractMatrix;
+    obsdim::Int = defaultobs
+    )
+    kerneldiagmatrix(κ, vec_of_vecs(X, obsdim = obsdim))
```
willtebbutt, line 137:
```suggestion
    return kerneldiagmatrix(κ, vec_of_vecs(X, obsdim = obsdim))
```

src/matrix/kernelmatrix.jl
```diff
@@ -168,31 +148,24 @@ In place version of [`kerneldiagmatrix`](@ref)
...
-        end
-        return K
+    K::AbstractVector,
+    κ::Kernel,
+    X::AbstractVector
+    )
+    if length(K) != length(X)
+        throw(DimensionMismatch("Dimensions of the target array K $(size(K)) are not consistent with X $(length(X))"))
+    end
+    map!(κ, K, X, X)
```
willtebbutt, line 170:
```suggestion
    return map!(κ, K, X, X)
```

src/matrix/kernelmatrix.jl
```diff
@@ -168,31 +148,24 @@ In place version of [`kerneldiagmatrix`](@ref)
...
-        else
-            for i in eachindex(K)
-                @inbounds @views K[i] = _kernel(κ,X[:,i],X[:,i])
-            end
-        end
-        return K
+    K::AbstractVector,
+    κ::Kernel,
+    X::AbstractVector
+    )
```
willtebbutt, line 166:
```suggestion
)
```

src/matrix/kernelmatrix.jl
```diff
@@ -168,31 +148,24 @@ In place version of [`kerneldiagmatrix`](@ref)
...
-            end
-        else
-            for i in eachindex(K)
-                @inbounds @views K[i] = _kernel(κ,X[:,i],X[:,i])
-            end
-        end
-        return K
+    K::AbstractVector,
+    κ::Kernel,
+    X::AbstractVector
```
willtebbutt, line 165:
```suggestion
    X::AbstractVector,
```

src/utils.jl
```diff
@@ -63,9 +65,18 @@ Base.getindex(D::RowVecs, i) = RowVecs(view(D.X, i, :))
 #     return T <: Real ? T : Float64
 # end
 
-check_dims(K,X,Y,featdim,obsdim) = check_dims(X,Y,featdim,obsdim) && (size(K) == (size(X,obsdim),size(Y,obsdim)))
+function check_dims(K, X::AbstractVector, Y::AbstractVector)
+    size(K) == (length(X), length(Y))
```
willtebbutt, line 69:
```suggestion
    return size(K) == (length(X), length(Y))
```

src/utils.jl
```diff
@@ -63,9 +65,18 @@ Base.getindex(D::RowVecs, i) = RowVecs(view(D.X, i, :))
...
+end
+
+
+## Won't be needed with full ColVecs implementation
+function check_dims(K, X::AbstractMatrix, Y::AbstractMatrix, featdim, obsdim)
+    check_dims(X, Y, featdim) &&
+    (size(K) == (size(X, obsdim), size(Y, obsdim)))
```
willtebbutt, line 76:
```suggestion
    return (size(K) == (size(X, obsdim), size(Y, obsdim)))
```

src/kernels/scaledkernel.jl
```diff
@@ -15,6 +15,8 @@ end
 
 kappa(k::ScaledKernel, x) = first(k.σ²) * kappa(k.kernel, x)
 
+kappa(k::ScaledKernel, x, y) = first(k.σ²) * kappa(k.kernel, x, y)
```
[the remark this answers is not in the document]
theogf: Because I realised that `ScaleKernel` could only be applied to `SimpleKernel`

src/kernels/kernelproduct.jl
```diff
@@ -29,23 +29,26 @@ hadamard(x,y) = x.*y
 function kernelmatrix(
     κ::KernelProduct,
     X::AbstractMatrix;
-    obsdim::Int=defaultobs)
-    reduce(hadamard,kernelmatrix(κ.kernels[i],X,obsdim=obsdim) for i in 1:length(κ))
+    obsdim::Int=defaultobs,
+)
+    reduce(hadamard, kernelmatrix(κ.kernels[i], X, obsdim = obsdim) for i in 1:length(κ))
```
devmotion, line 34: Typically
```suggestion
    return mapreduce(hadamard, kernels(κ)) do kernel
        kernelmatrix(kernel, X; obsdim = obsdim)
    end
```
should be more efficient since as the documentation says
```julia
mapreduce is functionally equivalent to calling reduce(op, map(f, itr); init=init), but will in general execute faster since no intermediate collection needs to be created.
```

devmotion: But again one maybe should do that in a separate PR :smile:

src/kernels/scaledkernel.jl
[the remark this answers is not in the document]
willtebbutt: Oh, okay. And this is `kappa` in the `kappa == kernelmatrix` sense, right?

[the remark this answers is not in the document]
theogf: Uh no, it's in the sense of `k(x,y)`, it multiplies by sigma for every function call. But this can get replaced when implementing #87

PR #84, opened 2020-04-17
Added ColVecs

theogf:
As a prerequisite of #83, this is the implementation of `ColVecs` taken from `Stheno.jl`


src/utils.jl
```diff
@@ -10,6 +10,30 @@ macro check_args(K, param, cond, desc=string(cond))
...
+
+Base.:(==)(D1::ColVecs, D2::ColVecs) = D1.X == D2.X
+Base.size(D::ColVecs) = (size(D.X, 2),)
+Base.length(D::ColVecs) = size(D.X, 2)
+Base.getindex(D::ColVecs, n::Int) = D.X[:, n]
+Base.getindex(D::ColVecs, n::CartesianIndex{1}) = getindex(D, n[1])
+Base.getindex(D::ColVecs, n) = ColVecs(D.X[:, n])
+Base.view(D::ColVecs, n::Int) = view(D.X, :, n)
+Base.view(D::ColVecs, n) = ColVecs(view(D.X, :, n))
+Base.eltype(D::ColVecs{T}) where T = Vector{T}
```
devmotion, line 31: `eltype` should be defined for types not instances (see https://docs.julialang.org/en/v1/base/collections/#Base.eltype):
```suggestion
Base.eltype(::Type{<:ColVecs{T}}) where T = Vector{T}
```

src/utils.jl
```diff
@@ -10,6 +10,30 @@ macro check_args(K, param, cond, desc=string(cond))
...
+    X::TX
+    ColVecs(X::TX) where {T, TX<:AbstractMatrix{T}} = new{T, TX}(X)
+end
+
+Base.:(==)(D1::ColVecs, D2::ColVecs) = D1.X == D2.X
+Base.size(D::ColVecs) = (size(D.X, 2),)
+Base.length(D::ColVecs) = size(D.X, 2)
+Base.getindex(D::ColVecs, n::Int) = D.X[:, n]
+Base.getindex(D::ColVecs, n::CartesianIndex{1}) = getindex(D, n[1])
+Base.getindex(D::ColVecs, n) = ColVecs(D.X[:, n])
```
devmotion, line 28: Are these implementations needed? I assumed, it might be sufficient to implement `Base.IndexStyle(::Type{<:ColVecs}) = Base.IndexLinear()` but maybe that's wrong.

src/utils.jl
```diff
@@ -10,6 +10,30 @@ macro check_args(K, param, cond, desc=string(cond))
...
+Base.view(D::ColVecs, n::Int) = view(D.X, :, n)
+Base.view(D::ColVecs, n) = ColVecs(view(D.X, :, n))
```
devmotion, line 30: Why is this line needed? For ranges? By default this falls back to scalar indexing, see https://docs.julialang.org/en/v1/manual/interfaces/#man-interface-array-1.

src/utils.jl
```diff
@@ -10,6 +10,30 @@ macro check_args(K, param, cond, desc=string(cond))
...
+Base.eltype(::Type{<:ColVecs{T}}) where T = Vector{T}
+Base.zero(D::ColVecs) = ColVecs(zero(D.X))
```
devmotion, line 32: Is this needed? I thought it's more important to implement `similar(::ColVecs, T, dims)`, but I guess `ColVecs` shouldn't be mutable?

willtebbutt:
Thanks for sorting this @theogf .

I implemented `ColVecs` quite a while ago now. If I'm honest I would consider removing most of the methods implemented, including `zero`,`view`, `iterate`. Basically everything that isn't necessary for `show` / `display` / generally printing the object. As @devmotion points out, it might be possible to go down to a single method of `getindex` as well, which would be great.

src/utils.jl
devmotion: Oh, I see the idea seems to be that for multiple indices you get back a `ColVecs` whereas for a single index only a column. The point about cartesian indices is still valid though.

src/utils.jl
devmotion: Again, seems we want to get a `ColVecs`.

devmotion:
It's also quite inconsistent that indexing yields arrays whereas iteration yields views. I guess that should be changed - maybe always return views? Then `iterate` could be removed.

theogf:
Yeah I am removing everything necessary and rely on `view` only

src/utils.jl
```diff
@@ -10,6 +10,30 @@ macro check_args(K, param, cond, desc=string(cond))
...
+A lightweight box for an `AbstractMatrix` to make it behave like a vector of vectors.
+"""
+struct ColVecs{T, TX<:AbstractMatrix{T}} <: AbstractVector{Vector{T}}
+    X::TX
+    ColVecs(X::TX) where {T, TX<:AbstractMatrix{T}} = new{T, TX}(X)
+end
+
+Base.:(==)(D1::ColVecs, D2::ColVecs) = D1.X == D2.X
+Base.size(D::ColVecs) = (size(D.X, 2),)
+Base.length(D::ColVecs) = size(D.X, 2)
```
devmotion, line 25: It would be sufficient to just define `size`.

src/utils.jl
```diff
@@ -10,6 +10,20 @@ macro check_args(K, param, cond, desc=string(cond))
...
 
 
+"""
+    ColVecs{T, TX<:AbstractMatrix}
+
+A lightweight box for an `AbstractMatrix` to make it behave like a vector of vectors.
+"""
+struct ColVecs{T, TX<:AbstractMatrix{T}} <: AbstractVector{SubArray}
+    X::TX
+    ColVecs(X::TX) where {T, TX<:AbstractMatrix{T}} = new{T, TX}(X)
```
devmotion, line 20: Can we make the element type concrete, e.g., by using
```suggestion
struct ColVecs{T, TX<:AbstractMatrix{T}, S<:AbstractVector{T}} <: AbstractVector{S}
    X::TX
    function ColVecs(X::TX) where {T, TX<:AbstractMatrix{T}}
        S = typeof(view(X, :, 1))
        return new{T, TX, S}(X)
    end
```
?

theogf: I am going to try to see if tests are still passing

theogf: Yay!

theogf:
I think this should be ready to merge.
I did not implement `view` as I don't think we will need it anyway...

src/utils.jl
```diff
@@ -10,6 +10,23 @@ macro check_args(K, param, cond, desc=string(cond))
 end
 
 
+"""
+    ColVecs{T, TX<:AbstractMatrix}
```
devmotion, line 14: Maybe just
```suggestion
    ColVecs
```
or add the third type parameter?

src/utils.jl
```diff
@@ -10,6 +10,23 @@ macro check_args(K, param, cond, desc=string(cond))
...
+    X::TX
+    function ColVecs(X::TX) where {T, TX<:AbstractMatrix{T}}
+        S = typeof(view(X, :, 1))
+        new{T, TX, S}(X)
+    end
+end
+
+Base.size(D::ColVecs) = (size(D.X, 2),)
+Base.getindex(D::ColVecs, i::Int) = view(D.X, :, i)
+Base.getindex(D::ColVecs, i::AbstractVector{Int}) = ColVecs(view(D.X, :, i))
```
devmotion, line 28: Maybe remove `AbstractVector{Int}` to get the same behaviour for `Colon`?

Issue #87, opened 2020-04-21
kappa vs _kernel vs k(x, y)


willtebbutt:
Related to #72 but more general.

### kappa

`kappa` can do one of several things, depending on context

1. if you can write the kernel as a function of the distance between two points (as in the so-called `SimpleKernel` case, `kappa` should do the rest of the computation. eg. for an EQ kernel, `kappa(k, d^2) = exp(-d^2 / 2)`.
2. if two inputs are provided and they're `Vector` or `Real`, treat them as single inputs and return a scalar that is the kernel evaluated at those inputs
3. if the inputs are matrices, compute `kernelmatrix`.

### `_kernel`

I'm really not sure what this does, and seems to be implemented only sporadically. I think it maybe computes the kernel for particular inputs, but I'm not really sure.

### `k(x, y)`

same as `kappa`, but without the `SimpleKernel` bit.

## Issue with the current situation

There are multiple ways to do the same thing, and `kappa` means different things depending upon the input types. This is bad, as it's confusing for new users. The best situation is that there is one, and only one, way to do things. This minimises the amount that users and developers have to learn to work with / develop the package. If they wish to have aliases, they should implement them themselves in their own code.

## Proposal

1. Completely remove `_kernel`. I don't think there will be much objection to this given #72
2. Remove all binary methods of `kappa`, so that only the unary method exists, and only for `SimpleKernel`s.
3. Change `k(x, y)` such that it only ever tries to evaluate the kernel at a pair of inputs (returns a scalar). It will never return a matrix.

With these changes, `kernelmatrix` / `kernelmatrix!` will become the sole way to compute the kernel matrix associated with a data set.

@theogf @devmotion what are your thoughts?

devmotion:
I completely agree and very much in favour of the proposed changes.

theogf:
I also fully agree with the proposal.
I am solving 1. with #83
2. and 3. Could be done in another PR

willtebbutt:
Great, will make a PR

devmotion:
Was fixed by https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/95.

Issue #94, opened 2020-04-23
ScaledKernel


willtebbutt:
Currently `ScaledKernel` implements the `kappa` / `metric` interface. This is wrong, as it prevents it from scaling non-`SimpleKernel`s. The fix is straightforward.

devmotion:
Was fixed by https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/99.

PR #95, opened 2020-04-23
Remove _kernel, sort out kappa

willtebbutt:
Addresses #87 .

Also no longer exports `kappa`, as it's now not needed as part of the public API.

I found a redundant function called `_scale`, so removed that.

Iteration over concrete subtypes of `Kernel` removed from `generic` as no longer required.

edit: the reason for quite a bit of code movement is that I originally hadn't realised I could implement `(k::SimpleKernel)(x, y)`, and had to modify most kernel files, so sorted out spacing while I was there. I stand my the changes in formatting, even if they're not _strictly_ the focus of this PR.


src/abstract_types.jl
```diff
@@ -0,0 +1,11 @@
...
+"""
+Abstract type defining a slice-wise transformation on an input matrix
+"""
+abstract type Transform end
+
+abstract type Kernel end
+abstract type BaseKernel <: Kernel end
+abstract type SimpleKernel <: BaseKernel end
+
+(k::SimpleKernel)(x, y) = kappa(k, evaluate(metric(k), x, y))
```
devmotion, line 11: This requires Julia >= 1.3 IIRC. I'm fine with dropping support for older Julia versions but so far the general opinion has been to support Julia >= 1.0. E.g., the `@eval` blocks and `concrete_type` "hacks" were added specifically to support the older versions and still get a similar (but reduced) functionality. IMO it would be much nicer to define methods for abstract types, but I think this change should be discussed at least.

devmotion:
>  I could implement (k::SimpleKernel)(x, y)

As I wrote in my comment above, I would be fine with this change but it requires us to drop support for Julia < 1.3 which we have been trying to avoid so far.

willtebbutt:
Oh, fair point. For some reason I thought it didn't affect this case... my mistake. Personally I'm completely happy to drop < 1.3 support -- it's not like we have packages that depend on `KernelFunctions` that are restricted to 1.0 as we've not been around for long enough. Moreover, 1.5 is likely to be the next LTS, at which point it really makes sense to insist that 1.5 is the minimum version of Julia that we support.

@theogf thoughts on this?

src/generic.jl
```diff
@@ -3,14 +3,6 @@ Base.length(::Kernel) = 1
 Base.iterate(k::Kernel) = (k,nothing)
 Base.iterate(k::Kernel, ::Any) = nothing
 
-# default fallback for evaluating a kernel with two arguments (such as vectors etc)
-kappa(κ::Kernel, x, y) = kappa(κ, evaluate(metric(κ), x, y))
-kappa(κ::TransformedKernel, x, y) = kappa(kernel(κ), apply(κ.transform,x), apply(κ.transform,y))
-kappa(κ::TransformedKernel{<:BaseKernel,<:ScaleTransform}, x, y) = kappa(κ, _scale(κ.transform, metric(κ), x, y))
-_scale(t::ScaleTransform, metric::Euclidean, x, y) =  first(t.s) * evaluate(metric, x, y)
-_scale(t::ScaleTransform, metric::Union{SqEuclidean,DotProduct}, x, y) =  first(t.s)^2 * evaluate(metric, x, y)
-_scale(t::ScaleTransform, metric, x, y) = evaluate(metric, apply(t, x), apply(t, y))
```
theogf, line 12: This should not be deleted, this should be adapted to `SimpleKernel`. This is an optimization for the simple case of a scalar lengthscale

src/kernels/scaledkernel.jl
```diff
@@ -15,7 +15,7 @@ end
 
 kappa(k::ScaledKernel, x) = first(k.σ²) * kappa(k.kernel, x)
```
theogf, line 16: Is this really needed anymore?

src/kernels/transformedkernel.jl
```diff
@@ -9,6 +9,12 @@ struct TransformedKernel{Tk<:Kernel,Tr<:Transform} <: Kernel
     transform::Tr
 end
 
+function (k::TransformedKernel)(x, y)
+    x′ = vec(apply(k.transform, reshape(x, :, 1); obsdim=2))
+    y′ = vec(apply(k.transform, reshape(y, :, 1); obsdim=2))
```
theogf, line 14:
```suggestion
    x′ = apply(k.transform, x)
    y′ = apply(k.transform, y)
```
All transforms are made to be applied on vectors as well

theogf:
I am also happy to drop support julia < 1.3. Given the new structure, it does not make sense anymore to keep it this way.

src/generic.jl
willtebbutt: Okay. I'll revert the change and open an issue.

src/kernels/scaledkernel.jl
willtebbutt: No, see #94 . Should be sorted in another PR

src/basekernels/rationalquad.jl
```diff
@@ -15,7 +15,8 @@ struct RationalQuadraticKernel{Tα<:Real} <: SimpleKernel
     end
 end
 
-kappa(κ::RationalQuadraticKernel, d²::T) where {T<:Real} = (one(T)+d²/first(κ.α))^(-first(κ.α))
+kappa(κ::RationalQuadraticKernel, d²::Real) = (1 + d² / first(κ.α))^(-first(κ.α))
```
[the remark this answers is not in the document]
willtebbutt: Because it's cleaner + sorts out style. Not technically part of this PR I guess. Will revert for now. I think we'll need another PR once things have settled down that just addresses style lol

willtebbutt:
All requested changes addressed. If either @theogf or @devmotion could re-review to approve / request more changes, that would be grand.

src/abstract_types.jl
devmotion, line 11: Not sure, but it seems it might be more fitting to put the definition of `(k::SimpleKernel)(x, y)` in `generic.jl`?

src/generic.jl
```diff
@@ -3,31 +3,20 @@ Base.length(::Kernel) = 1
...
 Base.iterate(k::Kernel, ::Any) = nothing
 
-# default fallback for evaluating a kernel with two arguments (such as vectors etc)
-kappa(κ::Kernel, x, y) = kappa(κ, evaluate(metric(κ), x, y))
-kappa(κ::TransformedKernel, x, y) = kappa(kernel(κ), apply(κ.transform,x), apply(κ.transform,y))
-kappa(κ::TransformedKernel{<:BaseKernel,<:ScaleTransform}, x, y) = kappa(κ, _scale(κ.transform, metric(κ), x, y))
+printshifted(io::IO, o, shift::Int) = print(io, o)
+Base.show(io::IO, κ::Kernel) = print(io, nameof(typeof(κ)))
+
+# See
```
devmotion, line 9: It seems you forgot to add a reference?

theogf:
Ok for me too! Just need to make the tests pass

src/generic.jl
```diff
@@ -3,31 +3,20 @@ Base.length(::Kernel) = 1
 Base.iterate(k::Kernel) = (k,nothing)
 Base.iterate(k::Kernel, ::Any) = nothing
 
-# default fallback for evaluating a kernel with two arguments (such as vectors etc)
-kappa(κ::Kernel, x, y) = kappa(κ, evaluate(metric(κ), x, y))
-kappa(κ::TransformedKernel, x, y) = kappa(kernel(κ), apply(κ.transform,x), apply(κ.transform,y))
-kappa(κ::TransformedKernel{<:BaseKernel,<:ScaleTransform}, x, y) = kappa(κ, _scale(κ.transform, metric(κ), x, y))
+printshifted(io::IO, o, shift::Int) = print(io, o)
+Base.show(io::IO, κ::Kernel) = print(io, nameof(typeof(κ)))
```
devmotion, line 7: I know you just copied this line, but is this in any way better than the default fallback for `show`? The default implementation of `show` would also show the field values which might actually be useful? I just think one needs a good reason for changing the default implementation for all `Kernel`s, and I'm not sure if there's one.

theogf: You're right! Probably the default show is better.

src/generic.jl
willtebbutt: Good catch, thanks :)

src/abstract_types.jl
willtebbutt: Fair point. I'm not sure that we want to the abstract types to live at the top level, but I'll open an issue about that, move them back, and move the method in question to generic.jl for now.

willtebbutt:
All comments addressed. Will merge once tests pass.

willtebbutt:
@theogf any idea what's going on with the test failures?

theogf:
It's weird I just tried the branch locally and it's passing the tests

willtebbutt:
Hmmm yeah, same. Presumably a dependency has broken somewhere in the pipeline. Maybe it's a Flux thing?

devmotion:
> It's weird I just tried the branch locally and it's passing the tests

It fails for me :thinking: Did you run `] up` before running the tests?

devmotion:
> Presumably a dependency has broken somewhere in the pipeline. Maybe it's a Flux thing?

I compared https://travis-ci.com/github/JuliaGaussianProcesses/KernelFunctions.jl/jobs/323153454 with https://travis-ci.com/github/JuliaGaussianProcesses/KernelFunctions.jl/jobs/322480149, and it seems the main difference in the dependencies are Zygote 0.4.17 instead of 0.4.15 and CUDAdrv 6.2.3 instead of 6.2.2. The Flux version was 0.10.4 in both cases.

willtebbutt:
[This change](https://github.com/FluxML/Zygote.jl/pull/611) in the [v0.4.17](https://github.com/FluxML/Zygote.jl/releases) release looks relevant...

devmotion:
Yep, seems someone else got similar issues with 0.4.17 (linked in the relevant PR). I just confirmed it locally, tests pass with 0.4.16.

willtebbutt:
I've temporarily pinned Zygote to 0.4.16. @devmotion do you have the bandwidth to chase this up on the Zygote repo?

PR #99, opened 2020-04-25
Move to AbstractVector

willtebbutt:
Addresses #43 .

In the process of refactoring, it became necessary to also address #94 as this turned out to be blocking.

Other notable changes:

- The `Transform` interface needed to change because the type of the input no longer determines whether you're operating on a single data point, or mapping across an entire data set. Consequently `apply` this PR removes `apply` in favour of evaluating the transform (transform a single data point) and `map(t, x)` to apply a transform over each element of a data set efficiently.
- Quite a lot of test have been added, but coverage is down a bit for some reason. I'll wait for tests to pass to figure out where we've lost stuff.

One thing I did that I should probably have opened an issue about is stop constraining the parameters of `ARDTransform` and `ScaleTransform` to be greater than zero as it's not a requirement of these transformations that you do so. I'm happy to revert the change in this PR and open an issue to discuss further if anyone objects strongly to this particular change.


src/generic.jl
```diff
@@ -25,3 +25,16 @@ end
 
 # Fallback implementation of evaluate for `SimpleKernel`s.
 (k::SimpleKernel)(x, y) = kappa(k, evaluate(metric(k), x, y))
+
+# This is type piracy. We should not doing this.
```
willtebbutt, line 29: I'm committing some type piracy here, but I've opened an [issue](https://github.com/JuliaStats/Distances.jl/issues/165) about this on Distances, so hopefully we'll be able to remove these particular bits of code.

src/kernels/transformedkernel.jl
```diff
@@ -44,14 +40,31 @@ end
...
+    return kernelmatrix!(K, kernel(κ), map(κ.transform, x))
+end
+
+function kernelmatrix!(
+    K::AbstractMatrix,
+    κ::TransformedKernel,
+    x::AbstractVector,
+    y::AbstractVector,
+)
+    return kernelmatrix!(K, kernel(κ), map(κ.transform, x), map(κ.transform, y))
```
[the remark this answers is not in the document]
theogf: Oh you did that later, nevermind!

src/transform/lowranktransform.jl
```diff
@@ -21,17 +21,13 @@ end
 Base.size(tr::LowRankTransform,i::Int) = size(tr.proj,i)
 Base.size(tr::LowRankTransform) = size(tr.proj) #  TODO Add test
 
-function apply(t::LowRankTransform, X::AbstractMatrix{<:Real}; obsdim::Int = defaultobs)
-    @boundscheck size(t,2) != size(X,feature_dim(obsdim)) ?
-        throw(DimensionMismatch("The projection matrix has size $(size(t)) and cannot be used on X with dimensions $(size(X))")) : nothing
-    @inbounds _transform(t,X,obsdim)
-end
+(t::LowRankTransform)(x::Real) = t([x])
```
theogf, line 24: Now that we use > 1.3 the wrapper for reals could be generalized to all `Transform` no?

src/transform/lowranktransform.jl
```diff
@@ -21,17 +21,13 @@ end
...
+(t::LowRankTransform)(x::AbstractVector{<:Real}) = t.proj * x
 
-function apply(t::LowRankTransform, x::AbstractVector{<:Real}; obsdim::Int = defaultobs) #TODO Add test
-    @assert size(t,2) == length(x) "Vector has wrong dimensions $(length(x)) compared to projection matrix"
-    t.proj*x
+function Base.map(t::LowRankTransform, x::AbstractVector{<:Real})
+    return ColVecs(t.proj * reshape(x, 1, :))
```
theogf, line 28:
```suggestion
    return ColVecs(t.proj * x')
```

src/transform/transform.jl
```diff
@@ -21,9 +21,7 @@ Return exactly the input
 """
 struct IdentityTransform <: Transform end
 
-apply(t::IdentityTransform, x; obsdim::Int = defaultobs) = x
-
-apply(t::Transform, x::ColVecs; obsdim::Int = defaultobs) = ColVecs(apply(t, x.X, obsdim = 2))
+apply(t::IdentityTransform, x::AbstractVector) = x
```
theogf, line 24: The change has been forgotten here

src/kernels/tensorproduct.jl
```diff
@@ -29,153 +29,85 @@ end
...
+        error("number of kernels and groups of features are not consistent")
+end
 
-    featuredim = feature_dim(obsdim)
-    if !check_dims(K, X, X, featuredim, obsdim)
-        throw(DimensionMismatch("Dimensions of the target array K $(size(K)) are not " *
-                                "consistent with X $(size(X))"))
-    end
+# Utility for slicing up inputs.
+slices(x::AbstractVector{<:Real}) = [x]
```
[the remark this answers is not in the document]
willtebbutt: This is in the context of pulling out length `N` slices for the tensor - product. In this context, the result should be a similar thing to if you call slices on a `ColVecs` of dimensionality `1`. Does that clear things up?

src/transform/lowranktransform.jl
willtebbutt: In this case I specifically want to ensure that the result is a vector, but in other cases when a `Real` is passed in you want to ensure that it remains a `Real`, so I don't think that such a fallback would be appropriate.

src/generic.jl
devmotion: An alternative to avoid the type piracy as long as the issue is not resolved would be to add an internal `_pairwise` function (or `pairwise` if `Distances.pairwise` is not imported) that forwards matrices just to `Distances.pairwise` and reshape vectors before forwarding them to `Distances.pairwise`.

willtebbutt: Yeah, I had wondered about doing that. I agree that it would make sense, but this just feels like the right definition lol. Would you object to including this piracy for the time being, and refactoring to your suggestion if the response on the `Distances.jl` issue is in the negative?

devmotion: I would be fine with adding it for now, but maybe let's open an issue so we don't forget it and fix it in case the upstream response is negative or the upstream changes take too long.

src/kernels/tensorproduct.jl
```diff
@@ -29,153 +29,85 @@ end
...
+slices(x::ColVecs) = eachslice(x.X; dims=1)
+slices(x::RowVecs) = eachslice(x.X; dims=2)
```
devmotion, line 40: Maybe just
```suggestion
slices(x::ColVecs) = eachrow(x.X)
slices(x::RowVecs) = eachcol(x.X)
```
?

src/kernels/tensorproduct.jl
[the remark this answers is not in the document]
devmotion: Maybe use a tuple instead of a vector here?
```suggestion
slices(x::AbstractVector{<:Real}) = (x,)
```

src/kernels/tensorproduct.jl
willtebbutt: Good shout.

src/kernels/tensorproduct.jl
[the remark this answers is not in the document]
willtebbutt: Also a good shout.

willtebbutt:
All comments addressed, ready for a second round of reviewing @devmotion @theogf

src/kernels/scaledkernel.jl
```diff
@@ -13,11 +13,42 @@ function ScaledKernel(kernel::Tk, σ²::Tσ²=1.0) where {Tk<:Kernel,Tσ²<:Real
...
+    return κ.σ² .* kerneldiagmatrix(κ.kernel, x)
+end
+
+function kernelmatrix!(
+    K::AbstractMatrix,
+    κ::ScaledKernel,
+    x::AbstractVector,
+    y::AbstractVector,
+)
+    K = kernelmatrix!(K, κ, x, y)
```
devmotion, line 36: It feels a bit weird to me to reassign `K` here (and in the implementations below):
```suggestion
    kernelmatrix!(K, κ, x, y)
```

src/kernels/scaledkernel.jl
```diff
@@ -13,11 +13,42 @@ function ScaledKernel(kernel::Tk, σ²::Tσ²=1.0) where {Tk<:Kernel,Tσ²<:Real
...
+    K .*= κ.σ² 
+    return K
+end
+
+function kernelmatrix!(K::AbstractMatrix, κ::ScaledKernel, x::AbstractVector)
+    K = kernelmatrix!(K, κ, x)
```
devmotion, line 42:
```suggestion
    kernelmatrix!(K, κ, x)
```

src/kernels/scaledkernel.jl
```diff
@@ -13,11 +13,42 @@ function ScaledKernel(kernel::Tk, σ²::Tσ²=1.0) where {Tk<:Kernel,Tσ²<:Real
...
+    K .*= κ.σ²
+    return K
+end
+
+function kerneldiagmatrix!(K::AbstractVector, κ::ScaledKernel, x::AbstractVector)
+    K = kerneldiagmatrix!(K, κ, x)
```
devmotion, line 48:
```suggestion
    kerneldiagmatrix!(K, κ, x)
```

src/kernels/scaledkernel.jl
devmotion: It might even break if a user implements `kernelmatrix!` but does not return `K`.

src/kernels/scaledkernel.jl
devmotion: It might even break if a user implements `kernelmatrix!` but does not return `K`.

willtebbutt:
Thanks for the review @devmotion

test/matrix/kernelmatrix.jl
```diff
@@ -1,103 +1,77 @@
-### Needs to be kept out for julia 1.0
-
-struct baseSE <: KernelFunctions.BaseKernel end
-(k::baseSE)(x, y) = exp(-evaluate(SqEuclidean(), x, y))
+# Custom kernel to test `SimpleKernel` interface on, independently the `SimpleKernel`s that
+# are implemented in the package. That this happens to be an exponentiated quadratic kernel
+# is a complete coincidence.
+struct ToySimpleKernel <: SimpleKernel end
+KernelFunctions.metric(::ToySimpleKernel) = SqEuclidean()
+KernelFunctions.kappa(::ToySimpleKernel, d) = exp(-d / 2)
```
theogf, line 6: I made a BaseKernel implementation to test directly if methods work with only the k(x,y) implemented.

willtebbutt: 🤦 good catch. Will restore that + associated functionality.

willtebbutt:
Thanks @theogf -- let me know if there's anything else that changing.

theogf:
I think that's all for me! We can improve the coverage with more tests later

src/matrix/kernelmatrix.jl
```diff
@@ -2,165 +2,143 @@
...
+Calculate the diagonal matrix of `X` with respect to kernel `κ`
+`obsdim = 1` means the matrix `X` has size #samples x #dimension
+`obsdim = 2` means the matrix `X` has size #dimension x #samples
+"""
+kerneldiagmatrix
+
+
+
+#
+# Kernel implementations. Generic fallbacks that shouldn't ever be hit in practice.
```
devmotion, line 39: Is that actually true? E.g., for the Wiener kernel we might want to use this instead of the initial custom implementation in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/77/commits/5722672eb0eda164ac0d45d2d64b95efc6a4f46b.

willtebbutt: Good point. I'll soften the language.

theogf:
LGTM !

Issue #90, opened 2020-04-22
pairwise! for kernelmatrix!


theogf:
Again something for a separate PR (maybe) but why don't we use `pairwise!`?

_Originally posted by @devmotion in https://github.com/_render_node/MDExOlB1bGxSZXF1ZXN0NDA0NDgzNzQ4/pull_requests/unread_timeline_

Issue #100, opened 2020-04-26
Type Pirating Distances.jl


willtebbutt:
I would be fine with adding it for now, but maybe let's open an issue so we don't forget it and fix it in case the upstream response is negative or the upstream changes take too long.

_Originally posted by @devmotion in https://github.com/_render_node/MDI0OlB1bGxSZXF1ZXN0UmV2aWV3Q29tbWVudDQxNTEzODc4OA==/comments/review_comment_

st--:
It's not clear (to someone not familiar with the original context) what this is about, and the link does not seem to work for me, would it be sensible to clarify this issue a bit ?

willtebbutt:
This was originally about whether or not we added some functionality to `Distances` that we needed inside this package or do things properly (I think it was probably something like implementing `pairwise(::SqEuclidean, ::AbstractVector)` or something that hadn't been implemented in `Distances` but seemed very reasonable.

In the end, we opted to just roll our own functions that default to calling `Distances.jl`'s functions, but gives us the flexibility to add the functionality we need.

@theogf @devmotion is this a fair summary? I'm a bit on the details because it all happened a while ago.

devmotion:
I think that's it pretty much. It seems this issue could be closed? IIRC there are some other open issues related to `metric` such as e.g. fixing the name to something more appropriate such as `binary_op` but this specific issue seems to be resolved by using Distances only as fallback.

willtebbutt:
Agreed. I'll close. Anyone should re-open if they feel there's more to be discussed.

PR #102, opened 2020-04-27
Use pairwise! for inplace operations

devmotion:
Fixes https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/90 and a bug in the implementation of `kernelmatrix!` for `FBMKernel`.

### DIFF
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -38,3 +38,16 @@ function Distances.pairwise(
 )
     return pairwise(d, reshape(x, :, 1), reshape(y, :, 1); dims=1)
 end
+
+function Distances.pairwise!(out::AbstractMatrix, d::PreMetric, x::AbstractVector{<:Real})
+    return pairwise!(out, d, reshape(x, :, 1); dims=1)
+end
+
+function Distances.pairwise!(
+    out::AbstractMatrix,
+    d::PreMetric,
+    x::AbstractVector{<:Real},
+    y::AbstractVector{<:Real},
+)
+    return pairwise!(out, d, reshape(x, :, 1), reshape(y, :, 1); dims=1)
+end
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -43,7 +43,12 @@ dim(x::ColVecs) = size(x.X, 1)
 
 Distances.pairwise(d::PreMetric, x::ColVecs) = pairwise(d, x.X; dims=2)
 Distances.pairwise(d::PreMetric, x::ColVecs, y::ColVecs) = pairwise(d, x.X, y.X; dims=2)
-
+function Distances.pairwise!(out::AbstractMatrix, d::PreMetric, x::ColVecs)
+    return pairwise!(out, d, x.X; dims=2)
+end
+function Distances.pairwise!(out::AbstractMatrix, d::PreMetric, x::ColVecs, y::ColVecs)
+    return pairwise!(out, d, x.X, y.X; dims=2)
+end
 
 """
     RowVecs(X::AbstractMatrix)
@@ -68,8 +73,12 @@ dim(x::RowVecs) = size(x.X, 2)
 
 Distances.pairwise(d::PreMetric, x::RowVecs) = pairwise(d, x.X; dims=1)
 Distances.pairwise(d::PreMetric, x::RowVecs, y::RowVecs) = pairwise(d, x.X, y.X; dims=1)
-
-
+function Distances.pairwise!(out::AbstractMatrix, d::PreMetric, x::RowVecs)
+    return pairwise!(out, d, x.X; dims=1)
+end
+function Distances.pairwise!(out::AbstractMatrix, d::PreMetric, x::RowVecs, y::RowVecs)
+    return pairwise!(out, d, x.X, y.X; dims=1)
+end
 
 """
 Will be implemented at some point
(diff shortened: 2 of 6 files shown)


src/generic.jl
```diff
@@ -38,3 +38,16 @@ function Distances.pairwise(
 )
     return pairwise(d, reshape(x, :, 1), reshape(y, :, 1); dims=1)
 end
+
+function Distances.pairwise!(out::AbstractMatrix, d::PreMetric, x::AbstractVector{<:Real})
+    return pairwise!(out, d, reshape(x, :, 1); dims=1)
```
theogf, line 43:
```suggestion
    return pairwise!(out, d, x'; dims=1)
```
?

src/generic.jl
```diff
@@ -38,3 +38,16 @@ function Distances.pairwise(
...
+end
+
+function Distances.pairwise!(
+    out::AbstractMatrix,
+    d::PreMetric,
+    x::AbstractVector{<:Real},
+    y::AbstractVector{<:Real},
+)
+    return pairwise!(out, d, reshape(x, :, 1), reshape(y, :, 1); dims=1)
```
theogf, line 52:
```suggestion
    return pairwise!(out, d, x', y'; dims=1)
```

src/generic.jl
```diff
@@ -38,3 +38,16 @@ function Distances.pairwise(
 )
     return pairwise(d, reshape(x, :, 1), reshape(y, :, 1); dims=1)
 end
+
+function Distances.pairwise!(out::AbstractMatrix, d::PreMetric, x::AbstractVector{<:Real})
```
theogf, line 42: Shouldn't it be put in `utils.jl` ?

theogf: Or maybe collect all the pairwise functions and store them in an extra file?

devmotion: This is not supposed to be part of KernelFunctions, as the definitions for `pairwise` it's type piracy (see https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/100).

src/generic.jl
devmotion: I would keep `reshape` (as in the definitions of `pairwise` above), since `x'` returns a `LinearAlgebra.Adjoint` whereas `reshape` returns an `Array{Float64,2}` (that shares the same underlying memory as `x`).

src/generic.jl
devmotion: Same as above.

devmotion:
> Maybe some tests on the added pairwise! would be nice

I assumed it would be sufficient that every test of `kernelmatrix!` for a `SimpleKernel` tests `pairwise` implicitly. Tests of the `pairwise!` type piracy definition should be part of Distances (if it is added there).

src/generic.jl
willtebbutt: I agree with @devmotion here. @theogf are you happy to mark this resolved?

src/generic.jl
willtebbutt: This feels like the appropriate place for these right now as the corresponding `pairwise` definitions are also here. Going to mark this resolved.

src/utils.jl
```diff
@@ -43,7 +43,12 @@ dim(x::ColVecs) = size(x.X, 1)
 
 Distances.pairwise(d::PreMetric, x::ColVecs) = pairwise(d, x.X; dims=2)
 Distances.pairwise(d::PreMetric, x::ColVecs, y::ColVecs) = pairwise(d, x.X, y.X; dims=2)
-
+function Distances.pairwise!(out::AbstractMatrix, d::PreMetric, x::ColVecs)
+    return pairwise!(out, d, x.X; dims=2)
+end
+function Distances.pairwise!(out::AbstractMatrix, d::PreMetric, x::ColVecs, y::ColVecs)
+    return pairwise!(out, d, x.X, y.X; dims=2)
+end
```
willtebbutt, line 51: Could we have some explicit tests that these definitions agree with the in-place versions? Presumably placed right next to [this](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/adf55bf374258f16f1d0a9c175a46ac0327295bf/test/utils.jl#L25)

src/utils.jl
```diff
@@ -68,8 +73,12 @@ dim(x::RowVecs) = size(x.X, 2)
...
 Distances.pairwise(d::PreMetric, x::RowVecs) = pairwise(d, x.X; dims=1)
 Distances.pairwise(d::PreMetric, x::RowVecs, y::RowVecs) = pairwise(d, x.X, y.X; dims=1)
-
-
+function Distances.pairwise!(out::AbstractMatrix, d::PreMetric, x::RowVecs)
+    return pairwise!(out, d, x.X; dims=1)
+end
+function Distances.pairwise!(out::AbstractMatrix, d::PreMetric, x::RowVecs, y::RowVecs)
+    return pairwise!(out, d, x.X, y.X; dims=1)
+end
```
willtebbutt, line 81: Same as above - unit tests please :)

src/generic.jl
theogf: Sure!

willtebbutt:
> I assumed it would be sufficient that every test of kernelmatrix! for a SimpleKernel tests pairwise implicitly. Tests of the pairwise! type piracy definition should be part of Distances (if it is added there).

I agree that the right place for this is there, but it's still a good idea to have direct tests for this implementation. Makes it unambiguously clear what's causing problems if something breaks at some point, rather than having to deduce that it _must_ be the `pairwise!` implementation because everything is breaking.

Issue #96, opened 2020-04-24
_scale


willtebbutt:
This should not be deleted, this should be adapted to `SimpleKernel`. This is an optimization for the simple case of a scalar lengthscale

_Originally posted by @theogf in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/95_

PR #104, opened 2020-04-27
Optimization of ScaleTransform for SimpleKernel

devmotion:
Fixes https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/96 by adapting the old code to the new interface. I'm not sure, maybe there's a better way to achieve this optimization?

### DIFF
--- a/src/kernels/transformedkernel.jl
+++ b/src/kernels/transformedkernel.jl
@@ -10,6 +10,19 @@ struct TransformedKernel{Tk<:Kernel,Tr<:Transform} <: Kernel
 end
 
 (k::TransformedKernel)(x, y) = k.kernel(k.transform(x), k.transform(y))
+function (k::TransformedKernel{<:SimpleKernel,<:ScaleTransform})(x, y)
+    return kappa(k.kernel, _scale(k.transform, metric(k.kernel), x, y))
+end
+
+function _scale(t::ScaleTransform, metric::Euclidean, x, y)
+    return first(t.s) * evaluate(metric, x, y)
+end
+function _scale(t::ScaleTransform, metric::Union{SqEuclidean,DotProduct}, x, y)
+    return first(t.s)^2 * evaluate(metric, x, y)
+end
+function _scale(t::ScaleTransform, metric, x, y)
+    evaluate(metric, apply(t, x), apply(t, y))
+end
 
 """
 ```julia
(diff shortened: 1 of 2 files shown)


src/kernels/transformedkernel.jl
```diff
@@ -10,6 +10,19 @@ struct TransformedKernel{Tk<:Kernel,Tr<:Transform} <: Kernel
...
+end
+
+function _scale(t::ScaleTransform, metric::Euclidean, x, y)
+    return first(t.s) * evaluate(metric, x, y)
+end
+function _scale(t::ScaleTransform, metric::Union{SqEuclidean,DotProduct}, x, y)
+    return first(t.s)^2 * evaluate(metric, x, y)
+end
+function _scale(t::ScaleTransform, metric, x, y)
+    evaluate(metric, apply(t, x), apply(t, y))
```
theogf, line 24: apply is not valid anymore. It should be `t(x)`, `t(y)`.

devmotion: Ah true, thanks for spotting it!

willtebbutt:
Thanks for taking a look at this. Before merging, could someone please explain to me why these methods are more efficient than what would otherwise happen? I'm struggling to figure it out.

theogf:
Well if you would multiply `x` by `s` before evaluating the metric you would need to multiply every element of `x` by `s` first. Whereas if you do it afterwards you just do a scalar-scalar multiplication

willtebbutt:
Ah, okay. @devmotion could you add a comment to this effect?

devmotion:
I added a comment and constrained the optimization to `AbstractVector{<:Real}` to be a bit more conservative and avoid surprises for users.

src/kernels/transformedkernel.jl
```diff
@@ -10,7 +10,14 @@ struct TransformedKernel{Tk<:Kernel,Tr<:Transform} <: Kernel
...
 
 (k::TransformedKernel)(x, y) = k.kernel(k.transform(x), k.transform(y))
-function (k::TransformedKernel{<:SimpleKernel,<:ScaleTransform})(x, y)
+
+# Optimizations for scale transforms of simple kernels to save allocations:
+# Instead of a multiplying every element of the inputs before evaluating the metric,
+# we perform a scalar multiplcation of the distance of the original inputs, if possible.
+function (k::TransformedKernel{<:SimpleKernel,<:ScaleTransform})(
+    x::AbstractVector{<:Real},
+    y::AbstractVector{<:Real}
```
willtebbutt, line 19:
```suggestion
    y::AbstractVector{<:Real},
```

Issue #97, opened 2020-04-24
LowRankTransform


willtebbutt:
This isn't the best name, as it doesn't have to map from high to low dimensions. A better name might just be `LinearTransform`.

theogf:
Wouldn't linear transform imply adding a bias as well (AX + b). I know that the exact name for this would be affine but people use linear implicitly for it

willtebbutt:
Do they? I wasn't aware that that's a thing that people do 😬 . In any case, I'm not sure it's a good reason not to use the name linear transform for a thing that is quite clearly a linear transform. I would be fine also having an affine transform 🤷‍♂️ . Or just having an affine transform with a linear transform being a special case.

theogf:
Well we mostly hear about linear regression not affine regression :D 
But anyway you're right. Linear transform is more adapted.

PR #106, opened 2020-04-27
Use LinearTransform instead of LowRankTransform

devmotion:
Fixes https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/97.

In principle, the ARDTransform could be a special case of a LinearTransform with a diagonal matrix, but I guess we want to keep them separate? IMO we should keep at least the ScaleTransform since it could be applied in infinite-dimensional vector spaces as well whereas that's not the case for the LinearTransform and ARDTransform.


src/transform/lineartransform.jl
```diff
@@ -0,0 +1,36 @@
...
+
+julia> tr = LinearTransform(A)
+```
+"""
+struct LinearTransform{T<:AbstractMatrix{<:Real}} <: Transform
+    A::T
+end
+
+function set!(t::LinearTransform{<:AbstractMatrix{T}}, A::AbstractMatrix{T}) where {T<:Real}
+    @assert size(t.A) == size(A) "Size of the given matrix $(size(A)) and the transformation matrix $(size(t.A)) are not the same"
```
willtebbutt, line 21: Could you sort out the line length while we're here?

src/transform/lineartransform.jl
```diff
@@ -0,0 +1,36 @@
...
+    t.A .= A
+end
+
+(t::LinearTransform)(x::Real) = vec(t.A * x)
+(t::LinearTransform)(x::AbstractVector{<:Real}) = t.A * x
+
+function Base.map(t::LinearTransform, x::AbstractVector{<:Real})
+    return ColVecs(t.A * x')
+end
```
willtebbutt, line 30:
```suggestion
Base.map(t::LinearTransform, x::AbstractVector{<:Real}) = ColVecs(t.A * x')
```
fine on a single line 🙂

src/transform/lineartransform.jl
theogf:
```suggestion
    size(t.A) == size(A) || 
        error("Size of the given matrix $(size(A)) and the transformation matrix $(size(t.A)) are not the same")
```

PR #108, opened 2020-04-30
Remove Zygote bound

devmotion:
Should not be needed anymore since https://github.com/FluxML/Zygote.jl/pull/622 is released.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -20,7 +20,6 @@ Requires = "1.0.1"
 SpecialFunctions = "0.8, 0.9, 0.10"
 StatsBase = "0.32, 0.33"
 StatsFuns = "0.8, 0.9"
-Zygote = "= 0.4.16"
 ZygoteRules = "0.2"
 julia = "1.3"
 


willtebbutt:
I agree we shouldn't be pinning Zygote to a particular patch version, but we _should_ probably lower bound it to the whichever is the most recent patch version that works.

devmotion:
It seems it's not fixed yet :stuck_out_tongue: 

However, I think there's no need to lower bound Zygote if the tests pass - since it only affects broadcasting and hence the tests but no functionality that we define or use in KernelFunctions.

willtebbutt:
> However, I think there's no need to lower bound Zygote if the tests pass - since it only affects broadcasting and hence the tests but no functionality that we define or use in KernelFunctions.

But if we have tests that don't work unless a minimum version of Zygote is used, doesn't that suggest that the package doesn't work with Zygote unless a certain minimum version is used, and that we should therefore constrain it to prevent headaches for users?

devmotion:
No, I don't think that's the case here. We provide the ability to call `params(k)` for kernels (and that still works) but the tests fail because broadcasting of `Params` objects is broken in specific Zygote versions. As long as users don't try to broadcast with `Params` (which is neither done nor encouraged in this package and not needed for the provided functionality) everything will work fine.

willtebbutt:
Oh I see. Well in that case we're probably fine for now, but we'll almost certainly need to lower bound when we actually start testing whether or not Zygote is able to successfully operate on all of KernelFunctions' kernels.

devmotion:
It seems it's properly fixed in Zygote now: https://github.com/FluxML/Zygote.jl/pull/631

devmotion:
Closed in favour of https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/110.

PR #110, opened 2020-05-01
Rewrite Zygote tests without broadcasting

devmotion:
This is an alternative to https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/108 that should work with all (more or less) recent versions of Zygote since it does not depend on broadcasting. Note that the broadcasting fix in Zygote is not released yet, so broadcasting-based tests still fail with Zygote >= 0.4.17.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -20,7 +20,6 @@ Requires = "1.0.1"
 SpecialFunctions = "0.8, 0.9, 0.10"
 StatsBase = "0.32, 0.33"
 StatsFuns = "0.8, 0.9"
-Zygote = "= 0.4.16"
 ZygoteRules = "0.2"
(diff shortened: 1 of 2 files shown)


theogf:
So your replacement still fails with Zygote 0.4.17 right?

devmotion:
Hmm sorry, I don't know what you mean :grimacing: This PR works with basically all recent Zygote versions, it works with Zygote 0.4.16 (and most earlier versions, but they are untested) but also with Zygote >= 0.4.17. Only the broadcasting version of the tests (in the other PR) fails with the recent Zygote versions. In this PR there are no bounds on Zygote anymore (IIRC tests used Zygote 0.4.19).

PR #111, opened 2020-05-04
Length check before applying TensorProduct

sharanry:
As discussed with @willtebbutt earlier, `TensorProduct` doesn't check the dimension of the input before applying. This can lead to bugs which are hard to trace.

Currently:
```julia
julia> k = TensorProduct(ConstantKernel(c=2.0), ConstantKernel(c=2.0))
Tensor product of 2 kernels:
	- Constant Kernel (c = 2.0)
	- Constant Kernel (c = 2.0)

julia> k(rand(3), rand(3))
4.0
```  
It picks up the first `length(k)` dimensions of the input, ignoring the rest.

### DIFF
--- a/src/kernels/tensorproduct.jl
+++ b/src/kernels/tensorproduct.jl
@@ -23,6 +23,8 @@ end
 Base.length(kernel::TensorProduct) = length(kernel.kernels)
 
 function (kernel::TensorProduct)(x, y)
+    @assert length(x) == length(kernel) && length(x) == length(kernel) "number
+of kernels and number of features are not consistent"
     return prod(k(xi, yi) for (k, xi, yi) in zip(kernel.kernels, x, y))
 end
 
--- a/src/kernels/tensorproduct.jl
+++ b/src/kernels/tensorproduct.jl
@@ -23,8 +23,10 @@ end
 Base.length(kernel::TensorProduct) = length(kernel.kernels)
 
 function (kernel::TensorProduct)(x, y)
-    @assert length(x) == length(kernel) && length(x) == length(kernel) "number
-of kernels and number of features are not consistent"
+    if !(length(x) == length(y) == length(kernel))
+        throw(DimensionMismatch("number of kernels and number of features
+are not consistent"))
+    end
     return prod(k(xi, yi) for (k, xi, yi) in zip(kernel.kernels, x, y))
 end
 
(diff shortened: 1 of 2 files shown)


src/kernels/tensorproduct.jl
```diff
@@ -23,6 +23,8 @@ end
 Base.length(kernel::TensorProduct) = length(kernel.kernels)
 
 function (kernel::TensorProduct)(x, y)
+    @assert length(x) == length(kernel) && length(x) == length(kernel) "number
```
willtebbutt, line 26: This should probably throw a `DimensionMismatch`, rather than an `AssertionError`.

devmotion:
> Not all kernels have a fixed / known input dimensionality, so a sensible fallback for dim(::Kernel) would be a singleton type called ArbitraryDimensionality with no fields that you can error check on. I'm thinking something like

An alternative would be to define
```julia
isdimcompatible(::Kernel, x) = true
isdimcompatible(kernel::TensorProductKernel, x) = length(kernel.kernels) == dim(x)

function validate_kernel_dims(k::Kernel, x::AbstractVector)
    if isdimcompatible(k, x)
        throw(DimensionMismatch(
            "Dimensionality of kernel not compatible ",
            "with that of input",
        ))
    end
end
```
and only define `dim(kernel, x)` if needed.

willtebbutt:
Think I would prefer the name `dims_are_compatible(k, x)`, but that's a good point @devmotion .

theogf:
Actually do you think this is something we could put in common to use with specific transforms. Like if one use `LinearTransform` or `ARDTransform` the dimension of the kernel is no more arbitrary.

devmotion:
I guess you could just define (not sure about the exact fields an types)
```julia
function dims_are_compatible(kernel::TransformedKernel{<:Kernel,<:LinearTransform}, x)
    return size(kernel.transform.P, 2) == dim(x)
end
```
Did you have that in mind?

theogf:
Yeah but it gets trickier for more complex structures. I think it's better to leave the error happen when the transformation is applied, and return a meaningful message.

sharanry:
>If you would rather open this out into a separate issue, that's also fine. But I think my comment regarding the type of error should definitely be addressed before merging as is.

@willtebbutt I have addressed this suggestion. I will create a separate PR for a more general solution?

willtebbutt:
LGTM. Will merge once tests pass.

willtebbutt:
> I will create a separate PR for a more general solution?

@sharanry yes please 🙂

PR #112, opened 2020-05-06
[WIP]Add a generic way to check kernel's input dimensions compatibility.

sharanry:
Related PR #111 

Please find a generic way to deal with dimensions of inputs to kernels as @willtebbutt and @devmotion had suggested in #111 .

I had to remove restrictions on input type for `validate_dims` to accommodate few test cases for `TensorProduct`. 

If this setup is okay, I will proceed to apply similar checks to all kernels.

### DIFF
--- a/src/kernels/tensorproduct.jl
+++ b/src/kernels/tensorproduct.jl
@@ -25,7 +25,7 @@ Base.length(kernel::TensorProduct) = length(kernel.kernels)
 dims_are_compatible(k::TensorProduct, x) = length(k) == length(x)
 
 function (kernel::TensorProduct)(x, y)
-    validate_kernel_dims(kernel, x)
+    validate_kernel_dims(kernel, x, y)
     return prod(k(xi, yi) for (k, xi, yi) in zip(kernel.kernels, x, y))
 end
 
--- a/src/kernels/tensorproduct.jl
+++ b/src/kernels/tensorproduct.jl
@@ -26,7 +26,6 @@ dims_are_compatible(k::TensorProduct, x) = length(k) == length(x)
 
 function (kernel::TensorProduct)(x, y)
     validate_kernel_dims(kernel, x)
-
     return prod(k(xi, yi) for (k, xi, yi) in zip(kernel.kernels, x, y))
 end
 
--- a/src/kernels/tensorproduct.jl
+++ b/src/kernels/tensorproduct.jl
@@ -22,11 +22,11 @@ end
 
 Base.length(kernel::TensorProduct) = length(kernel.kernels)
 
+dims_are_compatible(k::TensorProduct, x) = length(k) == length(x)
+
 function (kernel::TensorProduct)(x, y)
-    if !(length(x) == length(y) == length(kernel))
-        throw(DimensionMismatch("number of kernels and number of features
-are not consistent"))
-    end
+    validate_kernel_dims(kernel, x)
+
     return prod(k(xi, yi) for (k, xi, yi) in zip(kernel.kernels, x, y))
 end
 
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -5,6 +5,8 @@ Base.iterate(k::Kernel, ::Any) = nothing
 
 printshifted(io::IO, o, shift::Int) = print(io, o)
 
+dims_are_compatible(::Kernel, x) = true
+
 ### Syntactic sugar for creating matrices and using kernel functions
 function concretetypes(k, ktypes::Vector)
     isempty(subtypes(k)) ? push!(ktypes, k) : concretetypes.(subtypes(k), Ref(ktypes))
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -117,11 +117,27 @@ function validate_inplace_dims(K::AbstractVector, x::AbstractVector)
     end
 end
 
-function validate_dims(x::AbstractVector, y::AbstractVector)
+function validate_dims(x, y)
     if dim(x) != dim(y)
         throw(DimensionMismatch(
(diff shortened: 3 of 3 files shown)


src/kernels/tensorproduct.jl
```diff
@@ -22,11 +22,10 @@ end
...
 Base.length(kernel::TensorProduct) = length(kernel.kernels)
 
+dims_are_compatible(k::TensorProduct, x) = length(k) == length(x)
+
 function (kernel::TensorProduct)(x, y)
-    if !(length(x) == length(y) == length(kernel))
-        throw(DimensionMismatch("number of kernels and number of features
-are not consistent"))
-    end
+    validate_kernel_dims(kernel, x)
```
devmotion, line 28: IMO you'd want to call
```suggestion
    validate_kernel_dims(kernel, x, y)
```
here. However, the problem with the way it is implemented in this draft is that everyone has to implement `dim` for the input types they want to use with TensorProduct since `validate_kernel_dims(k, x, y)` calls `validate_dims(x, y)` which just compares `dim(x)` and `dim(y)`. IMO it might be better to use something more generic such as
```julia
dims_are_compatible(x, y) = true
dims_are_compatible(x::AbstractVector, y::AbstractVector) = length(x) == length(y)
```
for `x` and `y` as well and check `dims_are_compatible(x, y)` instead of `dim(x) == dim(y)`. The main reason would be that I think it is better if one can just apply a TensorProduct to arbitrary types and in the worst case it fails downstream than being forced to define `dim(x)`.

sharanry: Sorry, I don't seem to get what is wrong with being forced to define `dim` for new input types. Won't this happen rarely?

devmotion: I work a lot with kernels on spaces of distribution, so it would at least affect me :smile: In general, I just think the package should provide reasonable checks by default (e.g., for `AbstractVector`) but not be in the way of user that just want to use some other input types. `TensorProduct` does not restrict the types of the input arguments, since they are just forwarded to the different kernels, and hence it feels weird to force users to implement `dim` for other types (which they might not even own).

willtebbutt:
> I had to remove restrictions on input type for validate_dims to accommodate few test cases for TensorProduct.

@sharanry which test cases? Given that all collections of inputs should now be represented as `AbstractVector`s, I'm very surprised that this is an issue.

sharanry:
>@sharanry which test cases? Given that all collections of inputs should now be represented as AbstractVectors, I'm very surprised that this is an issue.

The 3rd iteration of [this](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/df3819a77830bfdf0033f0222c2f3de736ad4f22/test/kernels/tensorproduct.jl#L19) for loop.

willtebbutt:
Hmmmm... @devmotion I'm assuming you wrote these particular tests? (possibly I refactored them blindly when introducing the everything-is-a-vector changes though). Specifically, what does an input `[a, b]` represent?

devmotion:
> Specifically, what does an input [a, b] represent?

It represents the same as `(a, b)`, namely an input that consists of two components `a` and `b` (which can be of any (possibly different) type and do not have to be real numbers). We allow the individual kernels of `TensorProduct` to be saved as an arbitrary iterable container, and in the same way the implementation allows for arbitrary iterable container as inputs. This corresponds to the mathematical definition of a tensor product kernel: it is a kernel on a product space that is defined as a product of kernels evaluated on the individual spaces. Consistently, real-valued vectors as inputs are interpreted as an iterable container of scalar components.

> Given that all collections of inputs should now be represented as AbstractVectors, I'm very surprised that this is an issue.

As far as I understand, the aim of the refactoring was to 1) get rid of the `obsdim` keyword arguments by encoding it in the `ColVecs` and `RowVecs` types, and 2) to interpret `(::MyKernel)(x, y)` always as evaluation of a kernel for a single data point and use `kernelmatrix` etc for evaluation of kernels on data sets. However, I guess we don't want to restrict the inputs to vectors only (since that, e.g., would rule out distributions as inputs or tuples of distributions and real vectors).

willtebbutt:
🤦 I was misreading the tests and imagined a call to `kernelmatrix` in them.

> As far as I understand, the aim of the refactoring was to 1) get rid of the obsdim keyword arguments by encoding it in the ColVecs and RowVecs types, and 2) to interpret (::MyKernel)(x, y) always as evaluation of a kernel for a single data point and use kernelmatrix etc for evaluation of kernels on data sets. However, I guess we don't want to restrict the inputs to vectors only (since that, e.g., would rule out distributions as inputs or tuples of distributions and real vectors).

You're completely correct.

> It represents the same as (a, b), namely an input that consists of two components a and b (which can be of any (possibly different) type and do not have to be real numbers). We allow the individual kernels of TensorProduct to be saved as an arbitrary iterable container, and in the same way the implementation allows for arbitrary iterable container as inputs. This corresponds to the mathematical definition of a tensor product kernel: it is a kernel on a product space that is defined as a product of kernels evaluated on the individual spaces. Consistently, real-valued vectors as inputs are interpreted as an iterable container of scalar components.

👍 

I think I probably just need to reproduce Sharan's error to figure out what's going on with the error he's seeing.

We're definitely going to need to be able to distinguish between a call to `validate_kernel_dims` with a single input and a vector of inputs. We can't do that by looking at the types alone, so it might be that we have to add some more information somehow.

willtebbutt:
@sharanry what's the status of this?

willtebbutt:
This is definitely stale, so I'm closing. Please feel free to re-open if you wish to revive.

Issue #113, opened 2020-05-12
pullback on map(transform, colvec) breaks types


theogf:
Here is a MWE :
```julia
using KernelFunctions, Zygote
A = KernelFunctions.ColVecs(rand(10, 10))
t = ScaleTransform(rand())
typeof(map(t, A)) # ColVecs
v, g = pullback(map, t, A)
typeof(v) # Vector{Vector}
```
This make AD with Zygote not feasible at the moment.

willtebbutt:
You should pretty much be able to copy+paste code from [Stheno.jl](https://github.com/willtebbutt/Stheno.jl/blob/a9a41981b8d25ae5188bb3873369daa26ac6f2b2/src/util/abstract_data_set.jl#L16). I've adapted it a bit below to reflect what we've got going on:

```julia
@adjoint function ColVecs(X::AbstractMatrix)
    pullback_ColVecs(Δ::NamedTuple{(:X,)}) = (Δ.X,)
    return ColVecs(X), pullback_ColVecs
end
```

Stheno's version is a little more liberal with the types it allows for `Δ` -- this is a mistake that I've not gotten around to fixing yet.

Additionally, if it's going to be the case that we'll be using `getindex` on a `ColVecs` in any code that needs to be differentiated through, then we need to define a custom adjoint for `getindex` also, to make sure that it does the right thing (currently it will fall back to Zygote's default implementation for `AbstractVector`s, which is necessarily a bit overzealous in terms of the things that it thinks it's correct for)

I would strongly recommend trying to avoid backprop-ing through `getindex` at all costs -- it's incredibly slow. We might need to open a separate issue about this.

theogf:
I think you implemented these adjoints already however that's not the problem I think.
The problem comes from the fact that the value returned by `pullback` is currently a vector of vector which seems to indicate that the pullback for map on a vector is defined in a custom manner. I think we need to define our own adjoint for map(t::Transform, x::ColVecs)

willtebbutt:
Oh, weird, that might be the case. We probably need more careful testing of the transform code then. I'll address at the same time as #88 .

theogf:
Here is the bad guy : https://github.com/FluxML/Zygote.jl/blob/3e0a904cae5565bc7b458ad6d575c531ce70dd96/src/lib/array.jl#L180

theogf:
Ok so just using a new name (not `Base.map`) let Zygote runs smoothly.
[...]

Issue #88, opened 2020-04-22
Implement test suite for custom kernels


theogf:
How about we add a testing interface for users to check if their custom kernel satisfy all the conditions.
As described in this blog post : https://white.ucc.asn.au/2020/04/19/Julia-Antipatterns.html (see Aside: Test Suites, for interface testing.)
We can then define one for `SimpleKernel` and `BaseKernel`

willtebbutt:
I've done this extensively in Stheno. I've started work on a port for KernelFunctions, but it would be better if we had #43 and #87 resolved before sorting this to minimise work.

PR #114, opened 2020-05-12
Series of tests for AD

theogf:
I started to re-adapt the tests given the new structure and make them look a bit nicer.
Right now the idea is to compare the results (when possible) with FiniteDifferences.jl
The tests are quite generic and changeable.
~~The idea is to test differentiability given the inputs for all ADs and to test the differentiability given the kernel parameters only given Zygote. (other kernels would require more complex solutions)~~
The tests functions for AD are contained in `test/utils_AD.jl`.
A first function needs to be called with FiniteDifferences.jl to check that it runs without errors (eliminate one source of error when comparing)
A second function will compare a AD (so far implemented for Zygote, ForwardDiff and ReverseDiff) with a kernelfunction which may take arguments. The test setup for a kernel would look like this:
```
kfunction = (l -> transform(SEKernel(), first(l)))
args = [2.0]
v = test_FiniteDiff(kfunction, args)
if !v.anynonpass
    for AD in [:Zygote, :ForwardDiff, :ReverseDiff]
        test_AD(AD, kfunction, args)
    end
end
```
where `k = kernelfunction(args)` and args is a vector
Alternatively for no args :
```
kfunction = SqExponentialKernel
v = test_FiniteDiff(kfunction)
if !v.anynonpass
    for AD in [:Zygote, :ForwardDiff, :ReverseDiff]
        test_AD(AD, kfunction)
    end
end
```

Are tested :
- `kappa` for `SimpleKernel`
- `k(x, y)` given `k`, `x` AND `y`
- `kernelmatrix(A, B)` for both `RowVecs` and `ColVecs`, also given `A`, `B` and `k` (I use `sum` to get a scalar on it)
` `kernelmatrix(A)` same as above

Are added to make more tests pass:
- Indirection to `_map` from `Base.map` for avoid `Zygote` bug with `map` on `AbstractVector`
- More adjoints for `DotProduct` and `Sinus`
- Fixes for some function to be able to differentiate through (especially Matern Kernel)


theogf:
One major issue is currently #113. And I have no idea how to solve it

willtebbutt:
To be honest, I was planning to address this at the same time as #88.

In my opinion the correct way to do this is to incorporate differentiability tests in with the standard test suite, as opposed to having a collection of tests in a separate location.

No objection to this going in for now to get us in roughly the right place, but I don't think it's the correct solution long-term.

theogf:
So you propose to integrate each all the AD test separately for each type of kernel in their test file?
It sounds good to me, but what I show for now is what do we even want to test. This is a very rough draft, also to explore what is currently failing.

willtebbutt:
> So you propose to integrate each all the AD test separately for each type of kernel in their test file?

Yup.

> but what I show for now is what do we even want to test.

We want to test _everything_ in the public-facing API for every kernel that we write tests for.

> This is a very rough draft, also to explore what is currently failing.

👍

theogf:
@willtebbutt Would you be satisfied by these tests? I would lay them out all over the tests but I prefer to check with you first :P

willtebbutt:
It's definitely a reasonable starting point. We shouldn't really be testing gradients, rather we should be testing pushforwards / pullbacks. Unfortunately ForwardDiff and ReverseDiff don't expose these in their public APIs, so I'll need to do some digging after NeurIPS.

I would propose you merge this, because it's a definitely improvement over what we currently have, and we revisit after NeurIPS when refactoring the test suite.

theogf:
Coverage has reduced because of some additional adjoints and functions that were needed for AD. Once I put all the tests in place it should actually go up :)

theogf:
So I spread all the tests to the BaseKernels, Kernels and Transforms and when tests were failing I simply left out the AD in question and left a @test_broken with a comment. I listed all the failures in #116

theogf:
I was also unable to find the proper adjoint for the pairwise function with the `Sinus` metric

theogf:
Tests are failing but not because of new tests (some issue with the FBM kernel). Other than that it should be ready to merge

src/distances/delta.jl
```diff
@@ -1,12 +1,14 @@
...
 
-@inline function Distances._evaluate(::Delta,a::AbstractVector{T},b::AbstractVector{T}) where {T}
+@inline function Distances._evaluate(::Delta, a::AbstractVector, b::AbstractVector) where {T}
     @boundscheck if length(a) != length(b)
         throw(DimensionMismatch("first array has length $(length(a)) which does not match the length of the second, $(length(b))."))
     end
     return a == b
 end
 
+Distances.result_type(::Delta, Ta::Type, Tb::Type) = promote_type(Ta, Tb)
```
devmotion, line 11: This won't specialize on the types `Ta` and `Tb` (see https://docs.julialang.org/en/v1/manual/performance-tips/#Be-aware-of-when-Julia-avoids-specializing-1).

src/distances/dotproduct.jl
```diff
@@ -1,13 +1,15 @@
...
 
-@inline function Distances._evaluate(::DotProduct, a::AbstractVector{T}, b::AbstractVector{T}) where {T}
+@inline function Distances._evaluate(::DotProduct, a::AbstractVector, b::AbstractVector)
     @boundscheck if length(a) != length(b)
         throw(DimensionMismatch("first array has length $(length(a)) which does not match the length of the second, $(length(b))."))
     end
     return dot(a,b)
 end
 
+Distances.result_type(::DotProduct, Ta::Type, Tb::Type) = promote_type(Ta, Tb)
```
devmotion, line 11: Same here.

src/distances/sinus.jl
```diff
@@ -8,7 +8,9 @@ Distances.parameters(d::Sinus) = d.r
 @inline (dist::Sinus)(a::AbstractArray, b::AbstractArray) = Distances._evaluate(dist, a, b)
 @inline (dist::Sinus)(a::Number, b::Number) = abs2(sinpi(a - b) / first(dist.r))
 
-@inline function Distances._evaluate(d::Sinus, a::AbstractVector{T}, b::AbstractVector{T}) where {T}
+Distances.result_type(::Sinus{T}, Ta::Type, Tb::Type) where {T} = promote_type(T, Ta, Tb)
```
devmotion, line 11: And here.

src/utils.jl
```diff
@@ -1,5 +1,7 @@
 hadamard(x, y) = x .* y
 
+loggamma(x) = first(logabsgamma(x))
```
devmotion, line 3: There's no reason to define `loggamma` other than avoiding the type piracy when defining an adjoint for `logabsgamma`, I guess?

src/zygote_adjoints.jl
```diff
@@ -1,9 +1,88 @@
...
+
+@adjoint function pairwise(s::Sinus, X::AbstractMatrix, Y::AbstractMatrix; dims=2)
+    D = pairwise(d, X, Y; dims = dims)
+    throw(error("Sinus metric has no defined adjoint for now... PR welcome!"))
+end
+
+@adjoint function pairwise(s::Sinus, X::AbstractMatrix; dims=2)
+  D = pairwise(d, X; dims = dims)
+  throw(error("Sinus metric has no defined adjoint for now... PR welcome!"))
+end
```
devmotion, line 70: Maybe just not define an adjoint?

src/zygote_adjoints.jl
```diff
@@ -1,9 +1,88 @@
...
+
+@adjoint function kappa(κ::MaternKernel, d::Real)
+    ν = first(κ.ν)
+    val, grad = pullback(_matern, ν, d)
+    return ((iszero(d) ? one(d) : val),
+    Δ -> begin
+        ∇ = grad(Δ)
+        return ((ν = [∇[1]],), iszero(d) ? zero(d) : ∇[2])
+    end)
+end
```
devmotion, line 84: Why is this needed? The definition of `kappa(::MaternKernel, ::Real)` seems quite simple, so I would have assumed that AD just works?

test/runtests.jl
```diff
@@ -1,16 +1,15 @@
...
-using FiniteDifferences
-using Flux
 using Kronecker
 using LinearAlgebra
 using PDMats
 using Random
 using SpecialFunctions
 using Test
-using Zygote
+using Flux: params, Chain, Dense
```
devmotion, line 9: Is it not possible to keep the old `using Flux` import?

test/utils_AD.jl
```diff
@@ -1,39 +1,111 @@
-allapprox(x,y,tol=1e-8) = all(isapprox.(x,y,atol=tol))
-FDM = central_fdm(5,1)
 
+FDM = FiniteDifferences.central_fdm(5, 1)
```
devmotion, line 2: It seems `FDM` is only used once in the definition of `gradient(::Val{:FiniteDiff}`, ...)`, so maybe just move it there?

test/utils_AD.jl
```diff
@@ -1,39 +1,111 @@
...
 
-function kappa_fdm(k::Kernel,d::Real)
-    first(FiniteDifferences.grad(FDM,x->kappa(k,x),d))
+function gradient(::Val{:ReverseDiff}, f::Function, args)
+    ReverseDiff.gradient(f, args)
 end
 
+function gradient(::Val{:FiniteDiff}, f::Function, args)
+    first(FiniteDifferences.grad(FDM, f, args))
+end
```
devmotion, line 27: Maybe drop the `::Function` annotation and make it the first argument. That would allow to specify `f` with the `do` syntax and be consistent with the general recommendations for positions of input arguments.

devmotion:
> Tests are failing but not because of new tests (some issue with the FBM kernel).

I'm surprised about these test failures - it has to be related to the PR in some way, otherwise they wouldn't have passed all the time on master? Maybe try to set a seed (if we haven't fixed one yet)?

willtebbutt:
> I'm surprised about these test failures - it has to be related to the PR in some way, otherwise they wouldn't have passed all the time on master? Maybe try to set a seed (if we haven't fixed one yet)?

Yeah, would definitely be nice to resolve this before merging.

src/distances/delta.jl
theogf: Distances do this for determining the type of the allocated matrix in `pairwise!`, if this is not defined it defaults to `Float64` and therefore breaks AD for ForwardDiff and others :
https://github.com/JuliaStats/Distances.jl/blob/f69f7888c92458ae671c893d079ecf5fc8d8accd/src/generic.jl#L35
https://github.com/JuliaStats/Distances.jl/blob/f69f7888c92458ae671c893d079ecf5fc8d8accd/src/generic.jl#L203

src/utils.jl
theogf: Is that really type piracy? `logabsgamma` return a value and a sign, but we are only interested in the value so a wrapper solves this problems and allows to create a sensible adjoint.

src/zygote_adjoints.jl
theogf: Sure! I just could not figure out the algebraic formulation of the adjoint for the Sinus

src/zygote_adjoints.jl
theogf: It is needed for the edge case where `x = 0`, `matern` is defined as a `iszero(d) ? one(d) : _matern(d)` and `Zygote` will just return nothing on this branch kind of thing. It breaks the backpropagation.

test/runtests.jl
theogf: Sure I can export Flux generally

test/utils_AD.jl
theogf: Yes but it is reused everytime the gradient is called. I wanted to avoid useless allocations, I can make it a const!

test/utils_AD.jl
theogf: Oh nice I did not think about it this way. I will change it.

theogf:
I will try to figure out the issue with FBM, but I thought this was maybe irrelevant to this PR ?

theogf:
Can I merge this?

theogf:
Ah maybe before I can put all dependencies for the test directly in a `Project.toml`

src/distances/delta.jl
devmotion: Sure, we don't want to restrict it to `Float64` only. My comment was just that Julia actually won't compile specialized versions of this function for different input types since you used `Ta::Type` instead of `::Type{Ta}`.

src/utils.jl
devmotion: No, `loggamma` is no type piracy. I was just wondering why you define the function at all and not just define a custom adjoint for `logabsgamma` that outputs `nothing` for the second argument - and my assumption was that you didn't define one because you wanted to avoid this type piracy (of defining an adjoint for a function that we don't own).

src/zygote_adjoints.jl
devmotion: Does `ifelse(iszero(d), one(d), _matern(d))` fix the Zygote issue? I've seen this pattern quite a bit in the Zygote source code, so maybe it's helpful in such cases?

test/Project.toml
```diff
@@ -0,0 +1,21 @@
+[deps]
+Distances = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
+FiniteDifferences = "26cc04aa-876d-5657-8c51-4c34ba976000"
+Flux = "587475ba-b771-5e3f-ad9e-33799f191a9c"
+KernelFunctions = "ec8451be-7e33-11e9-00cf-bbf324bd1392"
```
devmotion, line 5:
```suggestion
```
According to the [docs](https://julialang.github.io/Pkg.jl/v1/creating-packages/#Test-specific-dependencies-in-Julia-1.2-and-above-1): "Note that Pkg will add the tested package itself implictly."

theogf:
All tests are passing, if there is no objections I am going to merge

willtebbutt:
@theogf do you know why coverage has dropped?

theogf:
@willtebbutt, yes, I needed to add some functionalities (mostly for Zygote) but sometimes it was not enough to make all tests pass (yet I left them there as they will be needed in the future). Right now tests are either all correct and removed. I want to tackle all of the different AD problems (all mentioned in #116 ) but would prefer to go one PR at a time, ss there are definitely a lot of issues.

Issue #11, opened 2019-12-13
Missing examples in docs


IsakFalk:
The [examples in the docs](https://theogf.github.io/KernelFunctions.jl/dev/example/) is currently lacking, so it would be good to provide some examples.

Currently we have to fill out the following
- [ ] [Kernel Ridge Regression](https://theogf.github.io/KernelFunctions.jl/dev/example/#Kernel-Ridge-Regression-1)
- [ ] [Gaussian Process Regression](https://theogf.github.io/KernelFunctions.jl/dev/example/#Gaussian-Process-Regression-1)
- [ ] [Deep Kernel Learning](https://theogf.github.io/KernelFunctions.jl/dev/example/#Deep-Kernel-Learning-1)
- [ ] [Kernel Selection](https://theogf.github.io/KernelFunctions.jl/dev/example/#Kernel-Selection-1)

I have a good knowledge of the first two, but shouldn't be very hard to fix the other ones as well, but will take some time.

theogf:
Thanks for opening the issue. I already have drafts for kernel ridge regression, and I also made a simple SVM example. I need to find the references for the two last. It should be straight forward to make!

IsakFalk:
No problem! Regarding the papers, the documentation references papers to use (I thought you added them, but maybe not).

For deep kernel learning we have [the original paper](Deep Kernel Learning). We could probably just redo one of the experiments combining Flux with KernelFunctions and optimise it directly (conditioned on that the kernels work with Zygote / AutoDiff which I think but I'm not sure).

For the kernel selection part I can't really find a reference (at least not from AISTATS 2018 as referenced on the docs).

theogf:
I finally found the AISTATS 2018 paper I was looking for! https://arxiv.org/abs/1706.02524

PR #119, opened 2020-06-04
Fixing some documentation bugs

theogf:
A friend notified me that a lot of the links were wrong in the documentations.
I fixed them and I also added (very succinctly) the definition of the kernels added since. That would be great if you look over  to suggest changes, as I butchered some of the explanations.
This PR does not intend to fix the docs globally but at least correct some issues and missing parts.
I am actually preparing a separate PR to add the examples mentioned in #11.


theogf:
Any objections to merge this? This should not affect anything as it is just changing the docs

PR #123, opened 2020-06-19
Fixes #115

### LINKED ISSUE #115: Using a SimpleKernel on a vector of vector will fail
MWE : 
```julia
A = fill(rand(3), 2)
k = SqExponentialKernel()
kernelmatrix(k, A) # Tries to call pairwise
```

theogf:
This PR is the answer for #115 , ~~and also solve another issue with `mapslices` (used in `FunctionTransform`) to replace it by `slicemap` from `SliceMap.jl`. Allowing AD for Zygote.~~


devmotion:
Could we define our own `KernelFunctions.pairwise` and `KernelFunctions.pairwise!` instead, as suggested in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/115#issuecomment-628198675, to avoid adding more type piracy? 

I didn't know about SliceMap - is `slicemap` only used because it provides an adjoint for Zygote or is it also more efficient than `mapslices`? SliceMap adds indirect dependencies on Tracker and ForwardDiff, I'm not sure if that's an issue.

willtebbutt:
> SliceMap adds indirect dependencies on Tracker and ForwardDiff, I'm not sure if that's an issue.

This is an issue. These are not dependencies that I would want to bake into this package. We should be aiming for `ChainRulesCore` only.

src/distances/pairwise.jl
```diff
@@ -0,0 +1,44 @@
...
+    out::AbstractMatrix,
+    d::PreMetric,
+    X::AbstractVector,
+    Y::AbstractVector,
+)
+    broadcast!(d, out, X, Y')
+end
+
+pairwise!(out::AbstractMatrix, d::PreMetric, X::AbstractVector) =
+    pairwise!(out, d, X, X)
```
willtebbutt, line 19:
```suggestion
pairwise!(out::AbstractMatrix, d::PreMetric, X::AbstractVector) = pairwise!(out, d, X, X)
```

src/distances/pairwise.jl
```diff
@@ -0,0 +1,44 @@
+# Add our own pairwise function to be able to apply it on vectors
+
+function pairwise(d::PreMetric, X::AbstractVector, Y::AbstractVector)
+    broadcast(d, X, Y')
+end
```
willtebbutt, line 5:
```suggestion
pairwise(d::PreMetric, X::AbstractVector, Y::AbstractVector) = broadcast(d, X, Y')
```

PR #125, opened 2020-06-22
Relaxed transform to be used on Kernel

theogf:
This is already part of #38 but since it is likely to be closed and that this bug is pretty bad I put it separately here.
This fix the problem of `transform(transform(SEKernel(), 2.0), 2.0)`.
Would not work because `transform` only accept `BaseKernel`.
Now when `transform` is applied on a `TransformedKernel`, the transform becomes a `ChainTransform`.

### DIFF
--- a/src/kernels/transformedkernel.jl
+++ b/src/kernels/transformedkernel.jl
@@ -51,9 +51,6 @@ transform(k::Kernel, ρ::Real) = transform(k, ScaleTransform(ρ))
 
 transform(k::Kernel, ρ::AbstractVector) = transform(k, ARDTransform(ρ))
 
-transform(k::Kernel, ::Nothing) = k
-
-
 kernel(κ) = κ.kernel
 
 Base.show(io::IO, κ::TransformedKernel) = printshifted(io, κ, 0)
--- a/src/kernels/transformedkernel.jl
+++ b/src/kernels/transformedkernel.jl
@@ -31,9 +31,9 @@ _scale(t::ScaleTransform, metric, x, y) = evaluate(metric, t(x), t(y))
 
 """
 ```julia
-    transform(k::BaseKernel, t::Transform) (1)
-    transform(k::BaseKernel, ρ::Real) (2)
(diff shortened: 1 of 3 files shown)


src/kernels/transformedkernel.jl
```diff
@@ -31,21 +31,27 @@ _scale(t::ScaleTransform, metric, x, y) = evaluate(metric, t(x), t(y))
...
 (3) Same as (1) with an `ARDTransform` with scales `ρ`
 """
 transform
 
-transform(k::BaseKernel, t::Transform) = TransformedKernel(k, t)
+transform(k::Kernel, t::Transform) = TransformedKernel(k, t)
 
-transform(k::BaseKernel, ρ::Real) = TransformedKernel(k, ScaleTransform(ρ))
+transform(k::TransformedKernel, t::Transform) =
+    TransformedKernel(k.kernel, t ∘ k.transform)
```
willtebbutt, line 47:
```suggestion
function transform(k::TransformedKernel, t::Transform)
    return TransformedKernel(k.kernel, t ∘ k.transform)
end
```

src/kernels/transformedkernel.jl
```diff
@@ -31,21 +31,27 @@ _scale(t::ScaleTransform, metric, x, y) = evaluate(metric, t(x), t(y))
...
+
+transform(k::Kernel, ρ::Real) = transform(k, ScaleTransform(ρ))
+
+transform(k::Kernel, ρ::AbstractVector) = transform(k, ARDTransform(ρ))
+
+transform(k::Kernel, ::Nothing) = k
```
devmotion, line 54: I'm wondering when this might be useful?

devmotion: If it's for the `@kernel` PR, one could just check if `transform === nothing` in the macro and doesn't need to support it generally, I guess.

devmotion:
Could we add a test for this bug?

src/kernels/transformedkernel.jl
theogf: Yep I will remove it.

PR #130, opened 2020-06-29
ENH: add ANOVA kernel

### LINKED ISSUE #67: Add ANOVA kernel
http://crsouza.com/2010/03/17/kernel-functions-for-machine-learning-applications/#anova

tirthasheshpatel:
Fixes #67 

I have tried to add a ANOVA kernel but am worried about optimizations. [This paper section 9.2, algorithm 9.14](https://people.eecs.berkeley.edu/~jordan/kernels/0521813972c09_p291-326.pdf) uses a DP approach to efficiently evaluate the kernel matrix. I have implemented a naive recurring algorithm for computing the `kernelmatrix`.

Changes:
 - [x] Add ANOVA Kernel.
 - [x] Add `ANOVAKernel`'s test suite.
 - [x] Add docs

### DIFF
--- a/src/basekernels/anova.jl
+++ b/src/basekernels/anova.jl
@@ -1,0 +1,56 @@
+struct ANOVAKernel{T<:Real} <: BaseKernel
+    d::Vector{T}
+    function ANOVAKernel(; d::T=1.0) where {T<:Real}
+        @assert d > 0 "ANOVAKernel: Given degree d is invalid."
+        return new{T}([d])
+    end
+end
+
+function (κ::ANOVAKernel)(x::AbstractVector{<:Real}, y::AbstractVector{<:Real})
+    return sum(exp.( - (x .- y).^2 ) .^ first(κ.d))
+end
+
+(κ::ANOVAKernel)(x::Real, y::Real) = exp( - (x - y)^2 )^first(κ.d)
+
+Base.show(io::IO, κ::ANOVAKernel) = print(io, "ANOVA Kernel d = ", first(κ.d), ")")
+
+function kernelmatrix(κ::ANOVAKernel, x::AbstractVector)
+    k = zeros(eltype(x), dim(x), dim(x))
+    for d ∈ size(x)
+        col = reshape(x[d], dim(x), 1)
+        k += exp( - (col .- col').^2 ) .^ first(κ.d)
+    end
+    return k
+end
+
+function kernelmatrix!(K::AbstractMatrix, κ::ANOVAKernel, x::AbstractVector)
+    for d ∈ size(x)
+        col = reshape(x[d], dim(x), 1)
+        K += exp( - (col .- col').^2 ) .^ first(κ.d)
+    end
+    return K
+end
+
+function kernelmatrix(κ::ANOVAKernel, x::AbstractVector, y::AbstractVector)
+    k = zeros(eltype(x), dim(x), dim(y))
+    for d ∈ size(x)
+        colx = reshape(x[d], dim(x), 1)
+        coly = reshape(y[d], dim(y), 1)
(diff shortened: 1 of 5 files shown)


theogf:
Hi @tirthasheshpatel and thanks for your contribution. 

It is not clear to me what kernel you are trying to implement, is it the one you mention from your reference or the one metntionned in #67? This would first help if you could write the analytic form of the kernel in the inline docs (see other kernel for examples) and general docs.

Regardless of it it would be better for the kernel functions to not mutate an array, i.e. preallocate and then fit, as it is incompatible with some AD algorithms.

tirthasheshpatel:
> It is not clear to me what kernel you are trying to implement

I am implementing the one in #67. But to be consistent with other kernels, I have not considered the parameter $\sigma$. I have added the docs now.

> Regardless of it it would be better for the kernel functions to not mutate an array, i.e. preallocate and then fit, as it is incompatible with some AD algorithms

Any hint on how to do that? Maybe the for loop can be vectorized and a summed result can be returned?

tirthasheshpatel:
By for loop can be vectorized, I mean using `map` method to somehow do it. But I don't see a straight forward way to do it.

theogf:
So I did a bit more investigation this time and I think I had the wrong idea of what an ANOVA kernel is.
[In this paper Equation(27)](https://projecteuclid.org/download/pdfview_1/euclid.aos/1211819561) the ANOVA is correctly formulated is quite different than what is described in [http://crsouza.com/2010/03/17/kernel-functions-for-machine-learning-applications/#anova](http://crsouza.com/2010/03/17/kernel-functions-for-machine-learning-applications/#anova). The reference you gave [this paper](https://people.eecs.berkeley.edu/~jordan/kernels/0521813972c09_p291-326.pdf) is also correct. An ANOVA kernel would actually be a generalization of `KernelSum` and `TensorProduct`.
The definition given in  [http://crsouza.com/2010/03/17/kernel-functions-for-machine-learning-applications/#anova](http://crsouza.com/2010/03/17/kernel-functions-for-machine-learning-applications/#anova) is probably just a very specific case, it might still be interesting to implement it but this would need a different name. Also your definition would be different, you do `exp((x-y)^2)^d` however it should be `exp((x^k - y^k)^2)^d`

tirthasheshpatel:
I did a bit more investigation and it seems that the `x^k` in http://crsouza.com/2010/03/17/kernel-functions-for-machine-learning-applications/#anova means `k'th` feature dimension and not `x` to the power of `k`. It is just a convention and the source I have followed is https://github.com/gmum/pykernels/blob/c8afdc79a15197ad3be2a0db0118f5e948577f49/pykernels/regular.py#L185. I hope this makes it all clear.

devmotion:
I think in general we would like to implement eq 27 in http://www.kernel-machines.org/publications/pdfs/0701907.pdf (which is given as reference in the implementation you linked) in its general form. As mentioned above and stated in this reference, in the extreme cases (with respect to the order) this is just the `KernelSum` and `TensorProduct` that are already implemented in this package.

theogf:
I guess this would mean more or less starting again from scratch. @tirthasheshpatel you can look at the implementation of `KernelSum` and `TensorProduct` to make it happen!

tirthasheshpatel:
Sure @theogf!

theogf:
BTW this would not be `KernelSum` in the case `P=1`, because the kernels would be applied dimension-wise.
Or one would need to apply an appropriate `SelectTransform` for each kernel.

theogf:
@tirthasheshpatel I think this worth starting a whole new PR. And `ANOVAKernel` should be a `Kernel` and not a `BaseKernel`.

tirthasheshpatel:
Yeah! Will open another pr for this then.

PR #133, opened 2020-07-01
Visual example for kernel via AbstractGPs

theogf:
In the style of [Duvenaud Cookbook](http://www.cs.toronto.edu/~duvenaud/cookbook/index.html), I automatically added GP samples for each of the kernels. It's automatically generated via Documenter.

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -1,5 +1,10 @@
 [deps]
+AbstractGPs = "99985d1d-32ba-4be9-9821-2ec096f28918"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+KernelFunctions = "ec8451be-7e33-11e9-00cf-bbf324bd1392"
+Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
 
 [compat]
 Documenter = "0.23, 0.24, 0.25"
+AbstractGPs = "0.1.2"
+Plots = "1"
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -1,0 +1,25 @@
+site_name:        KernelFunctions.jl
+repo_url:         https://github.com/JuliaGaussianProcesses/KernelFunctions.jl
+site_description: Documentation for KernelFunctions.jl
+site_url:         https://github.com/theogf/AugmentedGaussianProcesses.jl
+site_author:      The Turing Team
+
+theme: readthedocs
+
+extra_css:
+  - assets/Documenter.css
+
+extra_javascript:
+  - https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.1/MathJax.js?config=TeX-AMS_HTML
+  - assets/mathjaxhelper.js
+
+markdown_extensions:
+  - extra
+  - tables
+  - fenced_code
+  - mdx_math
+
+docs_dir: 'build'
+
+pages:
+  - Home: index.md
--- a/docs/src/kernels.md
+++ b/docs/src/kernels.md
@@ -58,6 +58,15 @@ The [`CosineKernel`](@ref) is defined as
 
 where $x\in\mathbb{R}$.
 
+```@example plots
+k = CosineKernel()
+gp = GP(k)(x, 1e-5)
+plot(x, rand(gp, 4))
+savefig("cosine.png"); nothing #hide
+```
+
+[]!(cosine.png)
+
 ## Exponential Kernels
 
 ### Exponential Kernel
--- a/docs/src/kernels.md
+++ b/docs/src/kernels.md
@@ -4,7 +4,20 @@
 
 # Base Kernels
 
-These are the basic kernels without any transformation of the data. They are the building blocks of KernelFunctions
+These are the basic kernels without any transformation of the data. They are the building blocks of `KernelFunctions.jl`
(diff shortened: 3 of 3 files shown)


willtebbutt:
1. I definitely like this feature, it's a great idea!
2. Is this repo really the right place to locate it? It feels wrong to introduce dependencies on `AbstractGP`s into this package, since it's not an inherently GP-specific thing. So I would prefer to see these either in `AbstractGPs` or on the `JuliaGaussianProcesses` webpage (if that's a thing we have?)

theogf:
The dependency is only in the docs! So this should not be a problem there!

devmotion:
I agree with @willtebbutt, it's a great idea but probably should be moved to a different repo or webpage. If we want to add visualizations to the documentation in KernelFunctions, maybe one could plot the actual kernel functions, either as a function of both `x` and `y` or just for fixed `y = 0`. To me it feels this repo is really about the kernel functions whereas AbstractGP would be more suitable for the "cool" GP stuff.

theogf:
> maybe one could plot the actual kernel functions, either as a function of both x and y or just for fixed y = 0

That's what I had in mind first, but for some kernels, it makes no sense. For example for all the linear kernels you would not see anything. Also in my opinion, seeing the kernel function usually does not tell you much :sweat_smile:

theogf:
The only solution to this would be to plot a contour everytime, but then this is even less clear to understand I think.

willtebbutt:
> The only solution to this would be to plot a contour everytime, but then this is even less clear to understand I think.

I'm a big fan of a heatmap in general i.e. plot the kernel matrix -- it doesn't tell you what sample paths look like, but it at least tells you roughly what covaries with what.

I would prefer that all of these visualisations went in a more general set of docs though -- we could really do with org-level docs, like what SciML does.

devmotion:
> For example for all the linear kernels you would not see anything.

Haha yeah maybe the choice `y = 0` wasn't a good one, but it should be fine for some other (random) choice of `y`?

In any case, to me it still feels more appropriate to visualize the actual functions that we implement in this repo if we want to add visualizations to KernelFunctions. Visualizations in AbstractGP or for JuliaGaussianProcesses might be the better option though.

willtebbutt:
The more I think about it the more I would _really_ like to see org-level docs so that we can easily explain to potential users how all of the package relate to one another in a single place.

theogf:
> The more I think about it the more I would _really_ like to see org-level docs so that we can easily explain to potential users how all of the package relate to one another in a single place.

That sounds really good indeed. Maybe we could get a subwebsite in [turing.ml](turing.ml) like turing.ml/juliagps and get a similar structure. I am not sure how this would go. Do we need to create an extra repo for it?

willtebbutt:
It would be good to be separate from the Turing website I think. Does github let orgs host their own website, e.g. I use willtebbutt.github.io as my personal webpage and github somehow knows how to make it based on one of my repos.

willtebbutt:
https://docs.github.com/en/github/working-with-github-pages/about-github-pages

Presumably there's something here that could help us.

theogf:
Well right now kernelfunctions is hosted on juliagaussianprocesses.github.io so yes it is possible.

willtebbutt:
> Well right now kernelfunctions is hosted on juliagaussianprocesses.github.io so yes it is possible.

Let's just put the general docs there then!

theogf:
But then how do we proceed? Actually how does Turing.jl proceeds to have general docs? If that's okay we could copy the turing.ml website structure

devmotion:
> If that's okay we could copy the turing.ml website structure

My impression over the last months has been that the Jekyll (?) based system of the Turing webpage and in particular the integration with Weave (regarding Plots) is a bit annoying to work with and broken partly (I think mostly on the Weave side). Maybe we could just use Documenter + Literate. Something more similar to the Turing setup would be Franklin (+ Literate?).

theogf:
I literally have no idea how to set that up. Could you create a repo with the bare-bones structure?

devmotion:
Depends on what we want to do exactly :smile: If we start with a simple Documenter page at `https://juliagaussianprocesses.github.io/` then we just have to create a repo "JuliaGaussianProcesses/juliagaussianprocesses.github.io" and setup Documenter therein (just like in KernelFunctions or AbstractGPs). The same applies if we want to use Franklin, we just have to setup Franklin in this repo according the Franklin docs. If we would like to use a different location for the webpage (if we would own one :stuck_out_tongue:), then we could name the repo in an arbitrary way and e.g. just add a CNAME file in the repo (as, e.g., done in https://github.com/SciML/DiffEqDocs.jl).

theogf:
Oh yeah you're right, I am completely stupid, it's just the equivalent of doing a personal webpage via github. I will investigate into it then! Maybe we can take this discussion to Zulip ? So we decide what do we use for the docs?

devmotion:
> If we want to add visualizations to the documentation in KernelFunctions, maybe one could plot the actual kernel functions

I just noticed that exactly this is actually done in Duvenaud's cookbook :smile:

willtebbutt:
> Oh yeah you're right, I am completely stupid, it's just the equivalent of doing a personal webpage via github. I will investigate into it then! Maybe we can take this discussion to Zulip ? So we decide what do we use for the docs?

Could you open an issue about it on `AbstractGPs` or `KernelFunctions`? Would prefer to keep as much of our discussions in public forums as possible :)

theogf:
Yeah that would be the best solution having a visualization of both!

theogf:
Closed in favor of #316

PR #135, opened 2020-07-08
[WIP] Adding examples

theogf:
This is definitely a WIP as not everything is working yet.
The idea is to implement the examples described in #11. 
This is using Literate.jl and a lot of other packages


theogf:
Also : I know the SVM solution is not correct

PR #138, opened 2020-07-17
Add datatype for multi-output GP input

sharanry:
Related Issue: [AbstractGPs#9](https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/issues/9)
Related old PR: [AbstractGPs#28](https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/pull/28)

This PR proposes a data type for inputs for a multi-output GP. 
A class of multi-output kernels which we hope to add to this repository will handle this datatype.


willtebbutt:
It would probably be a good idea to also add a single multi-output GP kernel to this PR. Would be good to have this functionality used for something before we merge it.

sharanry:
> It would probably be a good idea to also add a single multi-output GP kernel to this PR. Would be good to have this functionality used for something before we merge it.

Any preference on which kernel to start-off with?

willtebbutt:
It would probably make sense to do something really trivial, like a kernel that assumes each output is independent. So maybe you just have a kernel that wraps a single output kernel and a number that represents the number of outputs? I can't imagine why you would want this in practice, but it would probably give us what we need for this PR and would be really easy to test against single-output GPs.

sharanry:
@willtebbutt Could you take look at `IndependentKernel` which I just added and let me know what you think? I haven't added `kernelmatrix` support yet.

theogf:
That would be great to always have an indicator that one is using a multi-output kernel. For example having all multi-output kernels written as prefix+MOKernel, because IndependentKernel is a very confusing name.
Also independently from that, should we actually create a new package for multi output kernels? I feel multi-output kernels are a very specific task

willtebbutt:
> Also independently from that, should we actually create a new package for multi output kernels? I feel multi-output kernels are a very specific task

I wouldn't personally be in favour of that -- certainly they have their own peculiarities, but they're fundamentally kernels like any other. Indeed, part of my reason for wanting them to conform to the same API as regular kernels is to make them feel _less_ like something complicated and separate from other kernels.

I agree that all of the multi-output stuff should probably be kept in its own folder within this package though.

src/KernelFunctions.jl
```diff
@@ -49,6 +51,14 @@ abstract type Kernel end
 abstract type BaseKernel <: Kernel end
 abstract type SimpleKernel <: BaseKernel end
 
+"""
+    MOKernel
+
+An abstract kernel type for kernels which can model Multi-Output data. 
+"""
+abstract type MOKernel <: Kernel end
```
willtebbutt, line 59: Why do we need this type?

src/KernelFunctions.jl
```diff
@@ -69,6 +79,10 @@ include("kernels/tensorproduct.jl")
 include("approximations/nystrom.jl")
 include("generic.jl")
 
+include("mokernels/moinput.jl")
+include("mokernels/ind.jl")
```
willtebbutt, line 83: Could we rename this to independent please? We don't lose anything through being slightly more verbose here I reckon.

src/mokernels/ind.jl
```diff
@@ -0,0 +1,23 @@
...
+    IndependentKernel(k...) <: MOKernel
+
+A Multi-Output kernel which assumes each output is independent of the other.
+"""
+struct IndependentKernel{Tkernels<:AbstractVector} <: MOKernel
+    kernels::Tkernels
+    function IndependentKernel(k...)
+        k = collect(k)
+        return new{typeof(k)}(k)
+    end
```
willtebbutt, line 11: Should this be an outer "helper" constructor so that we can still construct one of these kernels using a vector of kernels directly?

src/mokernels/ind.jl
```diff
@@ -0,0 +1,23 @@
...
+end
+
+Base.length(κ::IndependentKernel) = length(κ.kernels)
+
+
+function (κ::IndependentKernel)(x::MOInput, y::MOInput)
+    @assert length(κ) == x.out_dim == y.out_dim
+    temp = vcat((κ.kernels[i](x.x, y.x) for i in 1:length(κ))...)
+    return hcat((temp for _ in 1:length(κ))...)
+end
```
willtebbutt, line 21: This function should map from a tuple to a scalar. See e.g. https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/fca569cb1304d1c6f51e4270cc17bbd87962b2b7/src/kernels/kernelsum.jl#L49

Additionally, we need the various `kernelmatrix` methods from the lines below the one linked.

src/mokernels/moinput.jl
```diff
@@ -0,0 +1,32 @@
+"""
```
willtebbutt, line 1: There are some indentation issues in this file -- I guess from when you copied stuff over from `AbstractGPs`?

sharanry:
> Indeed, part of my reason for wanting them to conform to the same API as regular kernels is to make them feel less like something complicated and separate from other kernels.

@willtebbutt Could you clarify what you mean by this? Wouldn't multi-output kernels accept a different type(vector of tuples) of inputs from regular kernels(vector of reals)?

test/mokernels/ind.jl
```diff
@@ -0,0 +1,14 @@
...
+
+    k = IndependentKernel(GaussianKernel(), Matern52Kernel(), GaborKernel())
+    @test k isa IndependentKernel
+    @test k isa KernelFunctions.MOKernel
+    @test k.kernels isa Vector{KernelFunctions.BaseKernel}
+    @test size(k(x, y)) == (3, 3)
+
+    v1 = [MOInput(rand(5), 3) for _ in 1:5]
+    v2 = [MOInput(rand(5), 3) for _ in 1:5]
+    # @info kernelmatrix(k, v1, v2)
```
willtebbutt, line 13: Does this need some more tests to check that e.g. `kernelmatrix` is implemented correctly?

src/KernelFunctions.jl
```diff
@@ -30,6 +30,8 @@ export NystromFact, nystrom
 
 export spectral_mixture_kernel, spectral_mixture_product_kernel
 
+export MOInput, length, size, lastindex, firstindex, getindex, iterate
```
willtebbutt, line 33:
```suggestion
export MOInput
```

test/mokernels/ind.jl
sharanry: Still working on `kernelmatrix`. None of the existing `kernelmatrix` functions work with vector of tuples.

src/KernelFunctions.jl
sharanry: I plan to use this to implement custom `kernelmatrix` functions for when the inputs are array of array of tuples.

willtebbutt:
> @willtebbutt Could you clarify what you mean by this? Wouldn't multi-output kernels accept a different type(vector of tuples) of inputs from regular kernels(vector of reals)?

Yes, that's correct. My point was more on the output side that it's not the case that our multi-output kernels will produce kernel matrix whose side-length is larger than the length of the input vector -- which is not what you get if you go down the matrix-valued kernel route. My point is that they _only_ differ in the domain of the input -- the rest of the concepts are the same: you provide an `AbstractVector` of inputs of length `N` (the `MOInput`) and get a kernel matrix of size `N x N`.

test/mokernels/ind.jl
willtebbutt: Ahh I see. Well if you start by only implementing them for the `MOInput` you should just be able to use them as usual by pulling out the input vector from `MOInput`, should you not?

src/KernelFunctions.jl
willtebbutt: I wouldn't have thought that you would need this -- can you not get it via usual dispatch?

e.g. for the fallback implementation in the case that someone _literally_ provides a vector of tuples could be something like:
```julia
(k::AnMOKernelOfSomeKind)(x, y) = <insert appropriate implementation>
kernelmatrix(k, x, y) = k.(x, permutedims(y))
```

(I don't think that you can escape implementing `(k::AnMOKernelOfSomeKind)` for each new multi-output kernel, and the `kernelmatrix` implementation doesn't need to know anything about multi-output kernels.

test/mokernels/ind.jl
sharanry: I imagine it is the other way. If we implement for Array of tuples we could use the same implementation of `MOInput` but collecting from `MOInput`. However this would not be efficient memory-wise.

willtebbutt: Yes. Having fallbacks that do this is fine, but we will certainly need code to compute `kernelmatrix` that exploits the structure provided in `MOInput` to do efficient computation.

src/mokernels/moinput.jl
sharanry: Sorry, I am not sure what you are referring to. Could you elaborate?

src/mokernels/ind.jl
sharanry: Wouldn't the output be a `out_dim X out_dim` matrix? This is what I understood from the presentation of _Scalable Exact Inference in Multi-Output Gaussian Processes_ which you referred me to.

willtebbutt: Right, so that particular paper (and a decent chunk of the literature) treats MOGPs as distributions over vector-valued functions, which produce a matrix-valued kernel. What we're doing (which isn't new -- lots of people subscribe to this view and it's definitely the more general one it's just less well-known) is viewing them as distributions over single-output functions, with an extra dimension that specifies which output you're referring to.

This gives you a unifying perspective on them, because it means that they're "just another GP", albeit often with structure in their inputs that you can exploit to accelerate computation etc. This is a good thing (imho)

src/KernelFunctions.jl
sharanry: Isn't this essential to use these functions?

willtebbutt: They're all defined and exported by `Base` :) Also, I wouldn't expect the casual user to need to use them -- they should just need to construct the input and pass it to the kernel.

src/matrix/kernelmatrix.jl
```diff
@@ -144,3 +144,13 @@ end
...
     return kerneldiagmatrix(κ, vec_of_vecs(X; obsdim=obsdim))
 end
+
+function kernelmatrix(
+    k::MOKernel, 
+    x::AbstractVector, 
+    y::AbstractVector
+    )
+    return k.(x, permutedims(collect(y)))
+end
```
willtebbutt, line 154: I don't know that this definition is necessary. Shouldn't it be handled by https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/fca569cb1304d1c6f51e4270cc17bbd87962b2b7/src/matrix/kernelmatrix.jl#L52
?

src/mokernels/independent.jl
```diff
@@ -0,0 +1,22 @@
+"""
+    IndependentMOKernel(k::Kernel) <: MOKernel
+
+A Multi-Output kernel which assumes each output is independent of the other.
+
```
willtebbutt, line 5:
```suggestion
```

src/mokernels/independent.jl
```diff
@@ -0,0 +1,22 @@
...
+"""
+struct IndependentMOKernel{Tkernel<:Kernel} <: MOKernel
+    kernel::Tkernel
+end
+
+function (κ::IndependentMOKernel)(x::Tuple{Vector, Int}, y::Tuple{Vector, Int})
+    return κ.kernel(x[1], y[1])
```
willtebbutt, line 12: Should this not return zero when `last(x) != last(y)`?

src/mokernels/independent.jl
```diff
@@ -0,0 +1,22 @@
...
+end
+
+function kernelmatrix(k::IndependentMOKernel, x::MOInput, y::MOInput)
+    @assert x.out_dim == y.out_dim
+    temp = k.kernel.(x.x, permutedims(y.x))
+    return repeat(temp, outer=[y.out_dim, y.out_dim])
+end
```
willtebbutt, line 19: I think this should return a block-diagonal matrix, should it not?

src/mokernels/independent.jl
```diff
@@ -0,0 +1,22 @@
...
+
+Base.show(io::IO, k::IndependentMOKernel) = 
+print(io, string("Independent Multi-Output Kernel\n\t", string(k.kernel)))
```
willtebbutt, line 22:
```suggestion
function Base.show(io::IO, k::IndependentMOKernel)
    print(io, string("Independent Multi-Output Kernel\n\t", string(k.kernel)))
end
```

src/mokernels/moinput.jl
willtebbutt: I think it's just lines 8 and 9 now aren't indented. Also line 2 should be

test/mokernels/independent.jl
```diff
@@ -0,0 +1,12 @@
...
+    y = MOInput([rand(5) for _ in 1:4], 3)
+
+    k = IndependentMOKernel(GaussianKernel())
+    @test k isa IndependentMOKernel
+    @test k isa KernelFunctions.MOKernel
+    @test k.kernel isa KernelFunctions.BaseKernel
+    @test k(x[2], y[2]) isa Real
+
+    @test kernelmatrix(k, x, y) == kernelmatrix(k, collect(x), collect(y))
+end
```
willtebbutt, line 12:
```suggestion
end

```

test/mokernels/moinput.jl
```diff
@@ -0,0 +1,17 @@
...
+    @test size(mgpi, 1) == 12
+    @test size(mgpi, 2) == 1
+    @test lastindex(mgpi) == 12
+    @test firstindex(mgpi) == 1
+    @test mgpi[2] == (x[2], 1)
+    @test mgpi[5] == (x[1], 2)
+    @test mgpi[7] == (x[3], 2)
+    @test all([(x_, i) for i in 1:3 for x_ in x ] .== mgpi)
+
+end
```
willtebbutt, line 17:
```suggestion
end

```

src/KernelFunctions.jl
```diff
@@ -69,6 +79,10 @@ include("kernels/tensorproduct.jl")
 include("approximations/nystrom.jl")
 include("generic.jl")
 
+include("mokernels/moinput.jl")
+include("mokernels/independent.jl")
+
+
```
willtebbutt, line 85:
```suggestion
```

src/matrix/kernelmatrix.jl
```diff
@@ -144,3 +144,13 @@ end
...
+
+kernelmatrix(k::MOKernel, x::AbstractVector) = kernelmatrix(k, x, x)
```
willtebbutt, line 156: Same here, I think we already have a fallback that should handle this.

src/matrix/kernelmatrix.jl
sharanry: The `validate_dims` step fails in that.

src/mokernels/independent.jl
sharanry: Yeah you are right. I forgot what the purpose of independent kernel is.

src/matrix/kernelmatrix.jl
willtebbutt: Is it because `dim` isn't defined for `MOInput`, or some other reason?

theogf: Actually that's another point : I think `validate_dims` implementation is uncorrect. As soon as you work with something else than `AbstractVector{Real}` it's going to wrongly return an error

theogf: I always end up commenting it out. I think we should remove it until we find a better way.

src/matrix/kernelmatrix.jl
sharanry: same as before, the fallback relies on `validate_dims`.

src/mokernels/independent.jl
sharanry: Yes, not sure what I was thinking.

src/KernelFunctions.jl
sharanry: It gave an error when I didn't export them earlier. Must have been for some other reason. Thanks for pointing it out!

src/matrix/kernelmatrix.jl
sharanry: @willtebbutt Yes it is because of `dims`.

willtebbutt: Well in that case why not just implement `dim` for the data type? For example,
```julia
dim(x::MOInput) = dim(x.x) + 1
```

willtebbutt: > I always end up commenting it out. I think we should remove it until we find a better way.

Why data types are you trying this with @theogf ? Is there a reason you can't implement `dim` for it?

theogf: Vectors of tuples. OK that's fair one can always implement `dim` for each case, but then we need a better error message. Any reason why don't use `length` instead?

willtebbutt: > but then we need a better error message.

I'm all for this.

> Any reason why don't use length instead?

Well you could ask about the `length` of the first element, but it's a bit different from promising that all of the elements are of the same dimensionality, which is what `dim` is meant to say. `dim` was an extension of the functionality that we had before to check that `AbstractMatrix`s passed as arguments are of the correct size.

To be honest I could probably get on board with not performing this check, but I think it's likely a valuable sanity check for the user.

sharanry: For now I have defined `dim(x::AbstractVector{Tuple{Any,Int}}) = 1` in the utils to accommodate `MOInput` and vector of tuples.

src/utils.jl
```diff
@@ -21,6 +21,7 @@ function vec_of_vecs(X::AbstractMatrix; obsdim::Int = 2)
 end
 
 dim(x::AbstractVector{<:Real}) = 1
+dim(x::AbstractVector{Tuple{Any,Int}}) = 1
```
willtebbutt, line 24:
```suggestion
dim(x::AbstractVector{Tuple{Any,Int}}) = dim(first(x)) + 1
dim(x::Tuple) = sum(dim, x)
```

src/matrix/kernelmatrix.jl
theogf: > Well you could ask about the length of the first element, but it's a bit different from promising that all of the elements are of the same dimensionality, which is what dim is meant to say.

Well you also going to ask only for the `dim` of the first object only no? 

> To be honest I could probably get on board with not performing this check, but I think it's likely a valuable sanity check for the user.

How about we create a `validate_dims` function which for unknown types (anything not `AbstractArray{Real}` just return true ?

sharanry: >How about we create a validate_dims function which for unknown types (anything not AbstractArray{Real} just return true ?

Doesn't this defeat the purpose of the sanity check?

willtebbutt: I'm actually fine with this. I think we need to re-work `validate_dims` anyway: https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/139

theogf: > Doesn't this defeat the purpose of the sanity check?

It does! But honestly it seems quite impossible to get an appropriate (and fast!) sanity check for arbitrary data. Hence the behavior : if we can do it fast and nice we do it, if we can't we just don't do it.

willtebbutt: I'm going to resolve this conversation. I'm happy with @sharanry 's current implementation for the sake of this PR . Possibly we'll want to re-work it in the future, but that's a different matter.

sharanry: Fri now we could selectively pass `validate_dims` without check for only type `AbstractVector{Tuple{Any,Int}}`?

sharanry:
Currently we are using `dim(x::AbstractVector{Tuple{Any,Int}}) = 1` which doesn't make much sense but helps pass the `validate_dims` for `MOInput` and vector of tuples without stringent check. Should I instead override `validate_dims` using type or is this fine?

sharanry:
Also, the build seems to fail for some inexact error outside the test.
```
Got exception outside of a @test

  InexactError: Int64(1.0e-20)

```
Seems to be a problem with AD tests which I haven't touched in this PR.

theogf:
Given the error messages I think it's due to https://github.com/JuliaDiff/FiniteDifferences.jl/pull/99
@willtebbutt probably has a better idea of what's going on

theogf:
Ok main issue is that `Delta` metric returns an `Int` and now `FiniteDifferences` is not happy about it.
The fix would be to replace
```
@inline function Distances._evaluate(::Delta, a::AbstractVector, b::AbstractVector) where {T}
        throw(DimensionMismatch("first array has length $(length(a)) which does not match the length of the second, $(length(b))."))
    end
    return a == b
end
```

by 

```
@inline function Distances._evaluate(::Delta, a::AbstractVector{Ta}, b::AbstractVector{Tb}) where {Ta, Tb}
        throw(DimensionMismatch("first array has length $(length(a)) which does not match the length of the second, $(length(b))."))
    end
    T = promote_type(Ta, Tb)
    return T(a == b)
end
```

src/distances/delta.jl
```diff
@@ -1,11 +1,12 @@
...
 end
 
-@inline function Distances._evaluate(::Delta, a::AbstractVector, b::AbstractVector) where {T}
+@inline function Distances._evaluate(::Delta, a::AbstractVector{Ta}, b::AbstractVector{Tb}) where {Ta, Tb}
     @boundscheck if length(a) != length(b)
         throw(DimensionMismatch("first array has length $(length(a)) which does not match the length of the second, $(length(b))."))
     end
-    return a == b
+    T = promote_type(Ta, Tb)
+    return T(a == b)
```
willtebbutt, line 9: This is amazingly picky, but I think it's probably more clear to write
```suggestion
    return convert(promote_type(Ta, Tb), a == b)
```

willtebbutt: Actually, I don't think this implementation will work properly with `ColVecs`. @sharanry could you test this please before we merge this PR?

willtebbutt:
Good catch btw @theogf . You're correct that this was introduced in the most recent PR. I think `FiniteDifferences` shouldn't be erroring with this, so I'll make + tag a fix today so that we can get this merged today.

src/distances/delta.jl
theogf: Oh yeah you're right! I just thought of the simplest case. Well it actually leads to the issue to what the output type of `Delta` should be...

willtebbutt: I mean, I think it's fine for it to be a `Bool`. I can't see what harm that causes -- when it interacts with a variance parameter that is e.g. a `Float32` it should be appropriately promoted.

willtebbutt:
@sharanry any idea where the drop in coverage is coming from?

sharanry:
> @sharanry any idea where the drop in coverage is coming from?

Few Base functions like `iterate` and few error throws were not tested. Should be fixed now.

PR #140, opened 2020-07-24
Tweak implementation

willtebbutt:
This reverts the change to the `Delta` distance made in #138 as it's not quite right, and FiniteDifferences 0.10.8 fixes the issue that we had before.

Also bumps the version so that we can release a new patch version.

### DIFF
--- a/src/distances/delta.jl
+++ b/src/distances/delta.jl
@@ -1,9 +1,12 @@
 struct Delta <: Distances.PreMetric
 end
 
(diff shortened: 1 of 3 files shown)


src/distances/delta.jl
```diff
@@ -5,7 +5,7 @@ end
     @boundscheck if length(a) != length(b)
```
theogf, line 5: You could also remove the parametrization on `Ta` and `Tb`

willtebbutt:
@sharanry let me know if you're happy with this, then I'll merge if tests pass.

PR #141, opened 2020-07-26
Fixing implicit gradients

### LINKED ISSUE #137: Implicit gradient failing with matrices
Here is a MWE:
```
using KernelFunctions, Flux, LinearAlgebra
k = transform(SqExponentialKernel(), 2.0)
ps = Flux.params(k)
X = rand(10, 1); x = vec(X)
A = rand(10, 10)
g = gradient(ps) do
  tr(kernelmatrix(k, X, obsdim = 1) * A)
end
g[ps[1]] == nothing

g2 = gradient(k) do k
  tr(kernelmatrix(k, X, obsdim = 1) * A)
end
g2[1].transform.s != nothing

g3 = gradient(ps) do
  tr(kernelmatrix(k, x) * A)
end
g3[ps[1]] != nothing
```
I think this is related to https://github.com/FluxML/Zygote.jl/issues/692
Any idea on how to solve this @willtebbutt ? It is probably connected to the `ColVecs` structure

theogf:
Until now there was a major issue where implicit gradients on `TransformedKernel`, i.e. :
```
k = transform(SqExponentialKernel(), 2.0)
gradient(params(k) do 
 something...
end
```
would return nothing.
This fixes #137 by not using map to apply the transform operation and adds tests to check the behaviour

### DIFF
--- a/src/transform/scaletransform.jl
+++ b/src/transform/scaletransform.jl
@@ -19,9 +19,9 @@ set!(t::ScaleTransform,ρ::Real) = t.s .= [ρ]
 
 (t::ScaleTransform)(x) = first(t.s) * x
 
-_map(t::ScaleTransform, x::AbstractVector{<:Real}) = t(x)
-_map(t::ScaleTransform, x::ColVecs) = ColVecs(t(x.X))
-_map(t::ScaleTransform, x::RowVecs) = RowVecs(t(x.X))
+_map(t::ScaleTransform, x::AbstractVector{<:Real}) = first(t.s) .* x
+_map(t::ScaleTransform, x::ColVecs) = ColVecs(first(t.s) .* x.X)
+_map(t::ScaleTransform, x::RowVecs) = RowVecs(first(t.s) .* x.X)
 
 Base.isequal(t::ScaleTransform,t2::ScaleTransform) = isequal(first(t.s),first(t2.s))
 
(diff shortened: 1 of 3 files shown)


theogf:
@willtebbutt @sharanry This correction is very important for me (AGP.jl relies completely on it). So I would merge it and release a new breaking release tonight.
Btw, I believe https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/issues/14 could be also solved by using another function then `Base.map`.

willtebbutt:
> @willtebbutt @sharanry This correction is very important for me (AGP.jl relies completely on it). So I would merge it and release a new breaking release tonight.

I'll try and take a look at this this evening.

src/transform/scaletransform.jl
```diff
@@ -17,11 +17,11 @@ end
...
 
-(t::ScaleTransform)(x) = first(t.s) .* x
+(t::ScaleTransform)(x) = first(t.s) * x
 
-_map(t::ScaleTransform, x::AbstractVector{<:Real}) = first(t.s) .* x
-_map(t::ScaleTransform, x::ColVecs) = ColVecs(first(t.s) .* x.X)
-_map(t::ScaleTransform, x::RowVecs) = RowVecs(first(t.s) .* x.X)
+_map(t::ScaleTransform, x::AbstractVector{<:Real}) = t(x)
+_map(t::ScaleTransform, x::ColVecs) = ColVecs(t(x.X))
+_map(t::ScaleTransform, x::RowVecs) = RowVecs(t(x.X))
```
willtebbutt, line 24: Do you know why this code needs to change? It's unclear to me as it's already using the `_map` function.

theogf: Yeah that is not necessary! Sorry for the noise. I thought it was nicer to have a more uniform approach

willtebbutt: I think I actually prefer the way it was written previously 😬

theogf: Ok I will revert it back

PR #143, opened 2020-07-30
Add Latent-factor multi-output kernel

sharanry:
This PR adds the Semi-Parametric Latent Factor kernel.

References: 
- https://willtebbutt.github.io/resources/mogp_duality.pdf
- https://infoscience.epfl.ch/record/161465/files/slfm-long.pdf


src/mokernels/slfm.jl
```diff
@@ -0,0 +1,67 @@
...
+        e::AbstractVector{<:Kernel}, 
+        A::AbstractMatrix
+        )
+        @assert (length(e), length(g)) == 
+            size(A) "Size of A not compatible to the given array of kernels"
+        return new{typeof(g), typeof(e), typeof(A)}(g, e, A)
+    end
+end
+
+function (κ::LatentFactorMOKernel)(x::Tuple{Vector, Int}, y::Tuple{Vector, Int})
```
devmotion, line 41: Maybe it would be more convenient and more readable to write
```suggestion
function (κ::LatentFactorMOKernel)((x, px)::Tuple{Vector, Int}, (y, py)::Tuple{Vector, Int})
```
and then use `x` and `px` instead of `first(x)` and `last(x)` in the function definition (and the same goes for `y` and `py`, of course).

src/mokernels/slfm.jl
```diff
@@ -0,0 +1,67 @@
...
+- `A::AbstractMatrix`: an matrix of weights for the kernels of shape (length(e), length(g))
+
+
+# Reference:
+- [Seeger, Teh, and Jordan (2005)](https://infoscience.epfl.ch/record/161465/files/slfm-long.pdf)
+
+"""
+struct LatentFactorMOKernel{
+    Tg <:AbstractVector{<:Kernel}, 
+    Te <:AbstractVector{<:Kernel},
```
devmotion, line 24: I guess it would be nice to not enforce these type constraints but just use
```suggestion
    Tg, 
    Te, 
```
Then one could use both tuples and vectors.

src/mokernels/slfm.jl
```diff
@@ -0,0 +1,67 @@
...
+    g::Tg
+    e::Te
+    A::TA
+    function LatentFactorMOKernel(
+        g::AbstractVector{<:Kernel}, 
+        e::AbstractVector{<:Kernel}, 
+        A::AbstractMatrix
+        )
+        @assert (length(e), length(g)) == 
+            size(A) "Size of A not compatible to the given array of kernels"
```
devmotion, line 36: IIRC `@assert` shouldn't be used for checks that are critical since it can potentially be removed by the compiler.
```suggestion
        (length(e), length(g)) == size(A)
            error("size of `A` is not compatible with the provided kernels")
```

src/mokernels/slfm.jl
```diff
@@ -0,0 +1,67 @@
...
+        return new{typeof(g), typeof(e), typeof(A)}(g, e, A)
+    end
+end
+
+function (κ::LatentFactorMOKernel)(x::Tuple{Vector, Int}, y::Tuple{Vector, Int})
+    if last(x) == last(y)
+        return sum([κ.g[i](first(x), first(y)) * κ.A[last(x), i] for i in 1:length(κ.g)]) +
```
devmotion, line 43: Maybe avoid creating the intermediate array by writing
```suggestion
        return sum(κ.g[i](first(x), first(y)) * κ.A[last(x), i] for i in 1:length(κ.g)) + 
```
or (if the function arguments are changed)
```suggestion
        return sum(κ.g[i](x, y) * κ.A[px, i] for i in 1:length(κ.g)) + 
```

src/mokernels/slfm.jl
```diff
@@ -0,0 +1,67 @@
...
+        return new{typeof(g), typeof(e), typeof(A)}(g, e, A)
+    end
+end
+
+function (κ::LatentFactorMOKernel)(x::Tuple{Vector, Int}, y::Tuple{Vector, Int})
+    if last(x) == last(y)
+        return sum([κ.g[i](first(x), first(y)) * κ.A[last(x), i] for i in 1:length(κ.g)]) + 
+        κ.e[last(x)](first(x), first(y))
+    else
+        return 0.0
```
devmotion, line 46: This introduces a type instability, maybe we can avoid it by computing some dummy value `z` of the same type as in the other branch and then return `zero(z)`?

src/mokernels/slfm.jl
```diff
@@ -0,0 +1,67 @@
...
+function Base.show(io::IO, k::LatentFactorMOKernel)
+    print(
+        io, 
+        string(
+            "Semi-parametric Latent Factor Multi-Output Kernel\n\t gs: ", 
+            string(k.g),
+            "\n\t es: ",
+            string(k.e),
+            )
+        )
```
devmotion, line 66: It's better to not convert to strings explicitly when printing stuff, i.e., something like
```suggestion
    print(
        io, 
        "Semi-parametric Latent Factor Multi-Output Kernel\n\t gs: ",
        k.g,
        "\n\t es: ",
        k.e,
    )
```
I'm wondering as well if this is already too verbose - the default `show` method should be very concise and ideally only print a single line (otherwise printing, e.g., an array of these kernels will be quite messy). One could still define a more verbose multi-line format by overloading `Base.show(io::IO, ::MIME"text/plain", k::LatentFactorMOKernel)` as well (see https://docs.julialang.org/en/v1/manual/types/#man-custom-pretty-printing-1).

src/mokernels/slfm.jl
```diff
@@ -0,0 +1,67 @@
...
+        κ.e[last(x)](first(x), first(y))
+    else
+        return 0.0
+    end
+end
+
+function kernelmatrix(k::LatentFactorMOKernel, x::MOInput, y::MOInput)
+    @assert x.out_dim == y.out_dim
+    @assert x.out_dim == 
+        size(k.A, 1) "Kernel not compatible with the given multi-output inputs"
```
devmotion, line 53: Maybe also here use `error` instead of `@assert`, if you want to ensure that it's checked?

src/mokernels/slfm.jl
sharanry: I am not sure what you mean. How do we ensure the computed dummy value is of the same type without actually executing it?

devmotion: I'm not completely sure, I was still wondering what's the best way to do it. Sometimes one can perform some possibly cheaper dummy operation by, e.g., using zeros instead of actually indexing etc. Alternatively one could always evaluate the first branch and just call `zero(res)` on its results `res` - however, that will perform superfluous computations if `last(x) != last(y)`.

sharanry: Are you worried about the first branch not giving a `Float64` output? When could this possibly happen?
Maybe we could explicity impose `Float64` on both the outputs. 
```julia
function (κ::LatentFactorMOKernel)((x, px)::Tuple{Vector, Int}, (y, py)::Tuple{Vector, Int})
    if px == py
        return Float64(sum([κ.g[i](x, y) * κ.A[px, i] for i in 1:length(κ.g)]) + 
            κ.e[px](x, y))
    else
        return Float64(0.0)
    end
end
```

I am not sure if this would cause any problems with AD.

devmotion: Yeah it could happen for dual numbers (or basically any other number types). Type instability occurs even if the kernels return values of type Float32 and the elements of A are of type Float32. IMO this example also shows that it's not a good idea to enforce Float64.

theogf:
I dont know if it's worth opening another PR for it specifically, but it would be nice to add a multi output section to the docs

src/mokernels/slfm.jl
```diff
@@ -19,49 +19,45 @@ A semiparametric kernel for problems involving multiple response variables.
...
-            string(k.e),
-            )
-        )
-end
\ No newline at end of file
+        "Semi-parametric Latent Factor Multi-Output Kernel\n\tgᵢ: ", 
+        [string(gi, "\n\t\t") for gi in k.g]...,
+        "\n\teᵢ: ",
+        [string(ei, "\n\t\t") for ei in k.e]...,
+    )
```
devmotion, line 62: You could use (something like)
```suggestion
        "Semi-parametric Latent Factor Multi-Output Kernel\n\tgᵢ: ")
    join(io, k.g, "\n\t\t")
    print(io, "\n\teᵢ: ")
    join(io, k.e, "\n\t\t")
```
to avoid the intermediate arrays and string conversions.

src/mokernels/slfm.jl
```diff
@@ -0,0 +1,63 @@
...
+- [Seeger, Teh, and Jordan (2005)](https://infoscience.epfl.ch/record/161465/files/slfm-long.pdf)
+
+"""
+struct LatentFactorMOKernel{Tg, Te, TA <: AbstractMatrix} <: Kernel
+    g::Tg
+    e::Te
+    A::TA
+    function LatentFactorMOKernel(g, e, A::AbstractMatrix)
+        all(isa.(g, Kernel)) || error("`g` should be an collection of kernels")
+        all(isa.(e, Kernel)) || error("`e` should be an collection of kernels")
```
devmotion, line 28: Maybe get rid of the allocations here by writing
```suggestion
        all(gi isa Kernel for gi in g) || error("`g` should be an collection of kernels")
        all(ei isa Kernel for ei in e) || error("`e` should be an collection of kernels")
```
here?

sharanry:
> I dont know if it's worth opening another PR for it specifically, but it would be nice to add a multi output section to the docs

I think docs as a whole require some fixing. https://juliagaussianprocesses.github.io/KernelFunctions.jl/stable/ they don't seem to load properly.

devmotion:
https://juliagaussianprocesses.github.io/KernelFunctions.jl/ works though.

sharanry:
> https://juliagaussianprocesses.github.io/KernelFunctions.jl/ works though.

Oh that's weird. Maybe we should change the link in the description then.

src/mokernels/slfm.jl
sharanry: @willtebbutt @theogf Any suggestions on this?

src/mokernels/slfm.jl
```diff
@@ -0,0 +1,60 @@
+@doc raw"""
+    LatentFactorMOKernel(
+        g::AbstractVector{<:Kernel}, 
+        e::AbstractVector{<:Kernel}, 
+        A::AbstractMatrix
+        )
+
+A semiparametric kernel for problems involving multiple response variables.
```
willtebbutt, line 8:
```suggestion
The kernel associated with the Semiparametric Latent Factor Model, introduced by Seeger, Teh and Jordan (2005).
```

src/mokernels/slfm.jl
```diff
@@ -0,0 +1,60 @@
...
+
+``k((x, p), (y, p)) = k_p(x, y) = \Sum^{Q}_{q=1} A_{pq}g_q(x, y) + e_p(x, y)``
+
+# Arguments
+- `g::AbstractVector{<:Kernel}`: an array of kernels
```
willtebbutt, line 13: Would help to make this a little more descriptive:
```suggestion
- `g::AbstractVector{<:Kernel}`: a vector of kernels, one for each latent process
```

src/mokernels/slfm.jl
```diff
@@ -0,0 +1,60 @@
...
+- `e::AbstractVector{<:Kernel}`: an array of kernels
```
willtebbutt, line 14: Same idea here
```suggestion
- `e::AbstractVector{<:Kernel}`: an vector of kernels, one for each output.
```

src/mokernels/slfm.jl
```diff
@@ -0,0 +1,60 @@
...
+- `A::AbstractMatrix`: an matrix of weights for the kernels of shape (length(e), length(g))
```
willtebbutt, line 15:
```suggestion
- `A::AbstractMatrix`: an matrix of weights for the kernels of size `(length(e), length(g))`
```

src/mokernels/slfm.jl
```diff
@@ -0,0 +1,60 @@
...
+- [Seeger, Teh, and Jordan (2005)](https://infoscience.epfl.ch/record/161465/files/slfm-long.pdf)
+
+"""
+struct LatentFactorMOKernel{Tg, Te, TA <: AbstractMatrix} <: Kernel
+    g::Tg
+    e::Te
+    A::TA
+    function LatentFactorMOKernel(g, e, A::AbstractMatrix)
+        all(gi isa Kernel for gi in g) || error("`g` should be an collection of kernels")
+        all(ei isa Kernel for ei in e) || error("`e` should be an collection of kernels")
```
willtebbutt, line 28: Could this not be achieved by type checking? i.e. constrain `g::AbstractVector{<:Kernel}` or something?

src/mokernels/slfm.jl
willtebbutt: I'm not completely clear on why there's this zero branch as the SLFM produces non-zero covariance between all outputs. Am I missing something?

src/mokernels/slfm.jl
sharanry: Yes we can but @devmotion suggested we allow tuples of Kernels as inputs too.

src/mokernels/slfm.jl
sharanry: I am confused. Which kernel(s) would we use if `px!=py?`

src/mokernels/slfm.jl
```diff
@@ -0,0 +1,61 @@
...
+
+The kernel associated with the Semiparametric Latent Factor Model, introduced by 
+Seeger, Teh and Jordan (2005).
+
+``k((x, p), (y, p)) = k_p(x, y) = \Sum^{Q}_{q=1} A_{pq}g_q(x, y) + e_p(x, y)``
+
+# Arguments
+- `g`: a collection of kernels, one for each latent process
+- `e`: a collection of kernels, one for each output
+- `A::AbstractMatrix`: an matrix of weights for the kernels of size `(length(e), length(g))`
```
theogf, line 16:
```suggestion
- `A::AbstractMatrix`: a matrix of weights for the kernels of size `(length(e), length(g))`
```

sharanry:
Any objections to merging this?

src/mokernels/slfm.jl
sharanry: @willtebbutt Is this resolved? I am still unsure how to handle cases when `px!=py`.

willtebbutt:
> Any objections to merging this?

Yes -- we still need to figure out what's going on with the off-diagonal kernel elements.

> @willtebbutt Is this resolved? I am still unsure how to handle cases when px!=py.

The best equation to look at is 1.7 from [my doc](https://willtebbutt.github.io/resources/mogp_duality.pdf). This lays out how to compute the covariance between two inputs that correspond to different outputs.

sharanry:
> > 
> 
> The best equation to look at is 1.7 from [my doc](https://willtebbutt.github.io/resources/mogp_duality.pdf). This lays out how to compute the covariance between two inputs that correspond to different outputs.

How  many `κ^e` kernels would there be? My initial assumption was that there would be `out_dim` of them. However the second half of the eq 1.7 would indicate we would need `out_dim*out_dim` of them. Is this true or am I missing something?

devmotion:
There's exactly one kernel k^e and one kernel k^g in this setup. The output dimension is taken care of by the modified input space (showing up by the additional p, p', and q in the equations).

sharanry:
> There's exactly one kernel k^e and one kernel k^g in this setup. The output dimension is taken care of by the modified input space (showing up by the additional p, p', and q in the equations).

So this implies both k^e and k^g would be multioutput kernels as they would need to handle modified input space?

willtebbutt:
> So this implies both k^e and k^g would be single multioutput kernels as they would need to handle modified input space?

Ish. You can definitely look at them that way:
```julia
k_g((x, p), (y, q)) = p == q ? kernels_g[p](x, y) : 0 
```
but notice that in practice you'll only ever query `k_g` when `p == q` because equation 1.7 has already cancelled the zero terms. So from an implementation perspective, you can just utilise vectors of kernels. But it might actually make sense to make `k_e` an `IndependentMOKernel` (or whatever it is we called the multi-output kernels).

src/mokernels/slfm.jl
```diff
@@ -0,0 +1,52 @@
...
+function (κ::LatentFactorMOKernel)((x, px)::Tuple{Any, Int}, (y, py)::Tuple{Any, Int})
+        return sum(κ.A[px, q] * κ.g[q](x, y) * κ.A[py, q] for q in 1:length(κ.g)) + 
+            κ.e((x, px), (y, py))
+end
+
+function kernelmatrix(k::LatentFactorMOKernel, x::MOInput, y::MOInput)
+    x.out_dim == y.out_dim || error("`x` and `y` should have the same output dimension")
+    x.out_dim == size(k.A, 1) ||
+        error("Kernel not compatible with the given multi-output inputs")
+    return k.(x, permutedims(collect(y)))
```
willtebbutt, line 40: I'm reluctant to let this go in without implementing this in terms of `kernelmatrix` on the kernels stored in `g` and `e`, as this is going to be the most efficient way to do this in general. e.g. if the kernels in `g` are `SqExponential` and the input-space is high-dimensional, you really need to use `kernelmatrix` to get reasonable performance.

@sharanry could you take a look at doing this? We can chat about it if it's unclear how to do this.

sharanry:
> The only other thing that I would ask for are that the AD tests are run for at least Zygote.

@willtebbutt I am not sure how to move forward with this. I am facing problems while defining adjoints for `MOKernels` which expect a pair of tuples whose second element is a integer. I there any way to indicate to `Zygote`/`FiniteDifferences` that we are not looking for a adjoint w.r.t to these integer parts of the inputs? I believe we would just need to return `nothing` for these parts to indicate to the calling functions that there is no adjoint information available.

willtebbutt:
This is actually a little tricky at the minute. I would suggest writing out the test yourself (using the same structure as the `test_AD` tests) and get the gradient w.r.t. the kernel parameters and the vector of inputs that you use to parametrise the `MOInput` object. This way you should be able to avoid handling the integer values entirely.

src/KernelFunctions.jl
```diff
@@ -50,6 +50,12 @@ abstract type Transform end
 
 abstract type Kernel end
 abstract type SimpleKernel <: Kernel end
+"""
+    MOKernel
+
+An abstract type for multi-output kernels.
+"""
+abstract type MOKernel <: Kernel end
```
willtebbutt, line 58: I think it would be better if this were moved to it's own file in the `mokernels` directory.

I'm just broadly not a fan of having actual code in the top-level file of a package.

test/mokernels/slfm.jl
```diff
@@ -0,0 +1,46 @@
...
+        "\t\tFractional Brownian Motion Kernel (h = 0.5)\n\teᵢ: " *
+        "Independent Multi-Output Kernel\n\tSquared Exponential Kernel"
+    )
+
+    # AD test
+    function test_slfm(A::AbstractMatrix, x1, x2)
+        k = LatentFactorMOKernel(
+            [MaternKernel(), SqExponentialKernel(), FBMKernel()],
+            IndependentMOKernel(GaussianKernel()),
+            A
```
willtebbutt, line 36:
```suggestion
            A,
```

test/mokernels/slfm.jl
```diff
@@ -0,0 +1,46 @@
...
+        )
+        return k((x1, 1), (x2, 1))
+    end
+
+    @test all(
+        FiniteDifferences.j′vp(FDM, test_slfm, 1, k.A, x1[1][1], x2[1][1]) .≈
```
willtebbutt, line 42: Better to use a random number rather than `1` -- being the multiplicative identity for scalars `1` can mask some bugs when testing AD.

test/mokernels/slfm.jl
```diff
@@ -0,0 +1,46 @@
...
+    N = 10
+    in_dim = 5
+    out_dim = 4
+    x1 = MOInput([rand(rng, in_dim) for _ in 1:N], out_dim)
+    x2 = MOInput([rand(rng, in_dim) for _ in 1:N], out_dim)
+
+    k = LatentFactorMOKernel(
+        [MaternKernel(), SqExponentialKernel(), FBMKernel()],
+        IndependentMOKernel(GaussianKernel()),
+        rand(rng, out_dim, 3)
```
willtebbutt, line 13:
```suggestion
        rand(rng, out_dim, 3),
```

src/mokernels/slfm.jl
```diff
@@ -0,0 +1,62 @@
...
+        all(gi isa Kernel for gi in g) || error("`g` should be an collection of kernels")
+        length(g) == size(A, 2) || 
+            error("Size of `A` not compatible to the given array of kernels `g`")
+        return new{typeof(g), typeof(e), typeof(A)}(g, e, A)
+    end
+end
+
+function (κ::LatentFactorMOKernel)((x, px)::Tuple{Any, Int}, (y, py)::Tuple{Any, Int})
+        return sum(κ.A[px, q] * κ.g[q](x, y) * κ.A[py, q] for q in 1:length(κ.g)) + 
+            κ.e((x, px), (y, py))
```
willtebbutt, line 33: Could this please be re-written as two statements? I think that would be easier to parse. Something like
```julia
cov_f = sum(...)
return cov_f + κ.e(...)
```
?

src/mokernels/slfm.jl
```diff
@@ -0,0 +1,62 @@
...
+
+"""
+struct LatentFactorMOKernel{Tg, Te <: MOKernel, TA <: AbstractMatrix} <: MOKernel
+    g::Tg
+    e::Te
+    A::TA
+    function LatentFactorMOKernel(g, e::MOKernel, A::AbstractMatrix)
+        all(gi isa Kernel for gi in g) || error("`g` should be an collection of kernels")
+        length(g) == size(A, 2) || 
+            error("Size of `A` not compatible to the given array of kernels `g`")
```
willtebbutt, line 26:
```suggestion
            error("Size of `A` not compatible with the given array of kernels `g`")
```

willtebbutt (approved):
This looks great. Once the minor suggestions have been addressed and the patch version bumped, please feel free to go ahead and merge.

Issue #139, opened 2020-07-23
Make validate_dims opt-in


willtebbutt:
`validate_dims` is slightly too aggressive at the minute. In particular, enforcing that the `dim` of each input is the same is expensive to check in general (e.g. to validate this properly you would need to check the length of each element of a `Vector{Vector{Float64}}`) and really just incorrect (e.g. string kernels can compare between strings of different lengths / dimensions) so it doesn't make sense to enforce this in general.

I propose instead to add a default implementation of `validate_dims` that doesn't error by default, i.e.
`validate_dims(x::AbstractVector, y::AbstractVector) = nothing`.

Then for any particular type we can offer it as an interface that you can implement if you want to be able to provide an informative error message.

It might also make sense to rename it to `validate_inputs` and generalise the idea slightly.

The `validate_inplace_dims` implementation, however, does make sense and does appear to be correct -- we really do want to make sure that the inputs are the correct length for the results to be stored in a matrix of the size of the one provided.

@theogf @devmotion @sharanry what are your thoughts?

theogf:
I fully approve! That's also what I kinda proposed in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/138#discussion_r459352030

willtebbutt:
Excellent. Once #138 is in I'll make the changes.

I think it's fine to view this particular piece of infrastructure as something that _hopefully_ catches errors, but isn't guaranteed to in all cases.

theogf:
Solved by #147

PR #147, opened 2020-08-02
Input check relaxation and change of `pairwise

theogf:
This PR aims at :
 - solving #139, `validate_dims` is replaced by `validate_inputs`. It will only check dimensions for `ColVecs` and `RowVecs`.
- Replacing `broadcast(k, x, y')` by `broadcast(k, x, permutedims(y))` in our `pairwise` for vectors. It appears to solve some Zygote issue (for me) and make it more coherent with the rest of the package

### DIFF
--- a/docs/src/create_kernel.md
+++ b/docs/src/create_kernel.md
@@ -8,7 +8,7 @@ Here are a few ways depending on how complicated your kernel is :
 
 If your kernel function is of the form `k(x, y) = f(d(x, y))` where `d(x, y)` is a `PreMetric`,
 you can construct your custom kernel by defining `kappa` and `metric` for your kernel.
-Here is for example how one can define the `SqExponentialKernel` again:
+Here is for example how one can define the `SqExponentialKernel` again :
 
 ```julia
 struct MyKernel <: KernelFunctions.SimpleKernel end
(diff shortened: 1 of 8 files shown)


devmotion:
I guess it would be good to add some tests?

src/utils.jl
```diff
@@ -94,9 +91,24 @@ For a transform return its parameters, for a `ChainTransform` return a vector of
 """
 #params
 
+dim(x) = 0
```
willtebbutt, line 94: Could you add another comment here pointing out that this is a passes-by-default choice.

theogf:
@willtebbutt I added some docs and changed slightly the `dim` approach

src/distances/pairwise.jl
```diff
@@ -1,6 +1,6 @@
 # Add our own pairwise function to be able to apply it on vectors
 
-pairwise(d::PreMetric, X::AbstractVector, Y::AbstractVector) = broadcast(d, X, Y')
+pairwise(d::PreMetric, X::AbstractVector, Y::AbstractVector) = broadcast(d, X, permutedims(Y))
```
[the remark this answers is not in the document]
theogf: For my experience `permutedims` is the most stable approach. I think `transpose` require the element of the vector to be "transposable" as well. For example `transpose(["string"])` will not work.

docs/src/create_kernel.md
```diff
@@ -2,19 +2,41 @@
...
+If your kernel does not satisfy such a representation, all you need to do is define `(k::MyKernel)(x, y)` and inherit from `Kernel`.
+For example we recreate here the `NeuralNetworkKernel`
+
+```julia
+struct MyKernel <: KernelFunctions.Kernel end
+
+(::MyKernel)(x, y) = asin(dot(x, y) / sqrt((1 + sum(abs2, x)) * (1 + sum(abs2, y))))
+```
+
+Note that `BaseKernel` do not use `Distances.jl` and can therefore be a bit slower.
```
theogf, line 31:
```suggestion
Note that `Kernel` do not use `Distances.jl` and can therefore be a bit slower.
```

docs/src/create_kernel.md
```diff
@@ -2,19 +2,41 @@
...
+Note that `Kernel` do not use `Distances.jl` and can therefore be a bit slower.
+
+### Additional Options
 
-Once these functions are defined, you can use all the wrapping functions of KernelFuntions.jl
+Finally there are additional functions you can define to bring in more features:
+ - Define the trainable parameters of your kernel with `KernelFunctions.trainable(k)` which should return a `Tuple` of your parameters.
+This parameters will be then passed to `Flux.params` function
+ - `KernelFunctions.iskroncompatible(k)`, if your kernel factorizes in the dimensions. You can declare your kernel as `iskroncompatible(k) = true`
+ - `KernelFunctions.dim`: by default the dimension of the inputs will only be checked for vectors of `AbstractVector{<:Real}`.
```
willtebbutt, line 39: It might be good to generalise this to say that you're allowed to add methods to `validate_inputs` if you've got special checks that you want to run for your input types, with `dim` being a nice special case.

docs/src/create_kernel.md
```diff
@@ -2,19 +2,41 @@
...
+
+### Additional Options
 
-Once these functions are defined, you can use all the wrapping functions of KernelFuntions.jl
+Finally there are additional functions you can define to bring in more features:
+ - `KernelFunctions.trainable(k::MyKernel)`: it defines the trainable parameters of your kernel, it should return a `Tuple` of your parameters.
+These parameters will be passed to the `Flux.params` function. For some examples see the `trainable.jl` file in `src/`
+ - `KernelFunctions.iskroncompatible(k::MyKernel)`: if your kernel factorizes in dimensions, you can declare your kernel as `iskroncompatible(k) = true` to use Kronecker methods.
+ - `KernelFunctions.dim(x::MyDataType)`: by default the dimension of the inputs will only be checked for vectors of type `AbstractVector{<:Real}`. If you want to check the dimensionality of your inputs, dispatch the `dim` function on your datatype. Note that `0` is the default.
+ - You can also directly overload `KernelFunctions.validate_inputs(x::MyDataType, y::MyDataType)` if you want to run special checks for your input types.
```
willtebbutt, line 40:
```suggestion
 - `dim` is called within `KernelFunctions.validate_inputs(x::MyDataType, y::MyDataType)`, which can instead be directly overloaded if you want to run special checks for your input types.
```

src/distances/pairwise.jl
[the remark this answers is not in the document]
devmotion: I see 👍

Issue #149, opened 2020-08-03
Removing `BaseKernel`


theogf:
I am not sure what changed, but now `BaseKernel` has no reason to exist apparently. Doing a search on where it is used shows it clearly. I think we can also drop these deprecated methods:

https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/86cde328efdf162371bef45334feb10f3c05a363/src/generic.jl#L14-L18

The only question is : Are we replacing `SimpleKernel` by `BaseKernel` for kernels using `Distances.jl` or are we keeping `SimpleKernel` and simply remove `BaseKernel` ?

willtebbutt:
I can't see a reason to keep `BaseKernel`.

theogf:
Solved by #150

PR #150, opened 2020-08-04
Remove BaseKernel and deprecated functions

theogf:
Following #149, this removes BaseKernel as well as the transition for deprecated constructors

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "KernelFunctions"
 uuid = "ec8451be-7e33-11e9-00cf-bbf324bd1392"
-version = "0.4.4"
+version = "0.4.5"
 
 [deps]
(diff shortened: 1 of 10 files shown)


theogf:
@willtebbutt Given that we remove the deprecated functions shouldn't this be a breaking release actually?

theogf:
At least after #147 is merged we should make a breaking release. A lot of major bugs have been fixed!

willtebbutt:
> At least after #147 is merged we should make a breaking release. A lot of major bugs have been fixed!

If they're just bug fixes then we shouldn't make a breaking release -- we should only do that if there are breaking changes.

willtebbutt:
I think you're right about the deprecated functions though -- a minor release is probably in order.

Issue #116, opened 2020-05-16
AD Failure


theogf:
Here is a list of the failures in the tests made in #114 I observed with the different ADs : ForwardDiff.jl, Zygote.jl and ReverseDiff.jl :

- [ ] GammaExponentialKernel : Zygote
- [ ] FBMKernel : ForwardDiff/Zygote
- [ ] GaborKernel : Zygote
- [ ] MahalanobisKernel : All
- [ ] MaternKernel : All
- [ ] NeuralNetworkKernel : Zygote
- [ ] PeriodicKernel : Zygote
- [ ] PiecewiseLinearKernel : All
- [ ] PolynomialKernel : All
- [ ] GammaRationalQuadratic : All
- [ ] SpectralMixture : All
- [ ] Wiener : All
- [ ] KernelProduct : Zygote
- [ ] KernelSum : Zygote
- [ ] FunctionTransform : Zygote 

This is a good starting point to try to find solutions

devmotion:
The Zygote problems with MaternKernel are caused by the fact that the partial derivative of `besselk` with respect to the first argument is defined as `NaN` in https://github.com/JuliaDiff/ChainRules.jl/blob/98c54587257b86cce6eb45f7870a75f897058d21/src/rulesets/packages/SpecialFunctions.jl#L46-L47 (and I assume the same problem exists for the other AD backends, since I get `NaN` for all of them when I try to run the commented out AD tests). I guess one would have to implement https://dlmf.nist.gov/10.38 to fix it.

theogf:
Haha making these derivatives sounds like one should write a whole package about bessel functions

yebai:
@sharanry Can you prioritise these AD issues? It would be great if these issues can be addressed during the summer.

devmotion:
BTW I found some publication from 2016 with closed-form expressions of the derivatives of the Bessel functions with respect to the order. I opened an issue at https://github.com/JuliaDiff/ChainRules.jl/issues/208 to discuss how one would deal with the additional dependencies needed for their implementations (they contain hypergeometric functions).

devmotion:
We might want to refactor KernelSum and KernelProd (making them concretely typed and allowing both tuples and vectors of kernels similar to TensorProduct, and probably removing the weights in KernelSum) before fixing any AD issues there.

theogf:
Agreed!
[...]

PR #152, opened 2020-08-11
Fix Zygote issues with FunctionTransform

sharanry:
https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/116#issuecomment-670881578

The problem was due to the usage of mutating function `mapslices`. We now avoid using it.

### DIFF
--- a/src/transform/functiontransform.jl
+++ b/src/transform/functiontransform.jl
@@ -16,8 +16,25 @@ end
 (t::FunctionTransform)(x) = t.f(x)
 
 _map(t::FunctionTransform, x::AbstractVector{<:Real}) = map(t.f, x)
-_map(t::FunctionTransform, x::ColVecs) = ColVecs(mapslices(t.f, x.X; dims=1))
+     
+
+function _map(t::FunctionTransform, x::ColVecs)
+    vals = map(axes(x.X, 2)) do i
+        t.f(view(x.X, :, i))
+    end
+    return ColVecs(hcat(vals))
+end
+
+# function _map(t::FunctionTransform, x::RowVecs)
+#     vals = map(axes(x.X, 1)) do i
+#         t.f(view(x.X, i, :))
+#     end
+#     return RowVecs(hcat(vals...))
+# end
+
+# _map(t::FunctionTransform, x::ColVecs) = ColVecs(mapslices(t.f, x.X; dims=1))
 _map(t::FunctionTransform, x::RowVecs) = RowVecs(mapslices(t.f, x.X; dims=2))
+
 
 duplicate(t::FunctionTransform,f) = FunctionTransform(f)
 
--- a/src/transform/functiontransform.jl
+++ b/src/transform/functiontransform.jl
@@ -22,14 +22,14 @@ function _map(t::FunctionTransform, x::ColVecs)
     vals = map(axes(x.X, 2)) do i
         t.f(view(x.X, :, i))
     end
-    return ColVecs(hcat(vals...))
+    return ColVecs(reduce(hcat, vals))
 end
 
 function _map(t::FunctionTransform, x::RowVecs)
     vals = map(axes(x.X, 1)) do i
         t.f(view(x.X, i, :))
     end
-    return RowVecs(hcat(vals...)')
+    return RowVecs(reduce(hcat, vals)')
 end
 
 duplicate(t::FunctionTransform,f) = FunctionTransform(f)
--- a/src/transform/functiontransform.jl
+++ b/src/transform/functiontransform.jl
@@ -32,10 +32,6 @@ function _map(t::FunctionTransform, x::RowVecs)
     return RowVecs(hcat(vals...)')
 end
 
-# _map(t::FunctionTransform, x::ColVecs) = ColVecs(mapslices(t.f, x.X; dims=1))
-# _map(t::FunctionTransform, x::RowVecs) = RowVecs(mapslices(t.f, x.X; dims=2))
-
-
 duplicate(t::FunctionTransform,f) = FunctionTransform(f)
 
 Base.show(io::IO, t::FunctionTransform) = print(io, "Function Transform: ", t.f)
--- a/src/transform/functiontransform.jl
+++ b/src/transform/functiontransform.jl
@@ -22,18 +22,18 @@ function _map(t::FunctionTransform, x::ColVecs)
     vals = map(axes(x.X, 2)) do i
         t.f(view(x.X, :, i))
     end
-    return ColVecs(hcat(vals))
+    return ColVecs(hcat(vals...))
 end
 
-# function _map(t::FunctionTransform, x::RowVecs)
-#     vals = map(axes(x.X, 1)) do i
-#         t.f(view(x.X, i, :))
-#     end
-#     return RowVecs(hcat(vals...))
-# end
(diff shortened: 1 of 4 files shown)


sharanry:
Slightly unrelated. I was wondering if we could add efficient ways to initialize `ColVecs` and `RowVecs` from a `Vector` of `Vector`s. I feel this is a common scenario and it might be useful to implement it the right way once.

src/transform/functiontransform.jl
```diff
@@ -16,8 +16,21 @@ end
 (t::FunctionTransform)(x) = t.f(x)
 
 _map(t::FunctionTransform, x::AbstractVector{<:Real}) = map(t.f, x)
-_map(t::FunctionTransform, x::ColVecs) = ColVecs(mapslices(t.f, x.X; dims=1))
-_map(t::FunctionTransform, x::RowVecs) = RowVecs(mapslices(t.f, x.X; dims=2))
+     
+
+function _map(t::FunctionTransform, x::ColVecs)
+    vals = map(axes(x.X, 2)) do i
```
devmotion, line 22: I still think `mapreduce(hcat, axes(x.X, 2))...` should be more performant than `map` and splatting - at least the former feels like the "correct" approach. Have you checked that your benchmarks are not affected by some type instability?

sharanry:
```julia
julia> @benchmark mapreduce(x->sin.(x), hcat, eachslice($(rand(1000,1000)); dims=2))
BenchmarkTools.Trial: 
  memory estimate:  3.74 GiB
  allocs estimate:  5002
  --------------
  minimum time:     866.146 ms (5.21% GC)
  median time:      1.031 s (5.28% GC)
  mean time:        1.003 s (5.43% GC)
  maximum time:     1.142 s (5.86% GC)
  --------------
  samples:          6
  evals/sample:     1

julia> @benchmark _map(t, $(RowVecs(rand(1000, 1000))))
BenchmarkTools.Trial: 
  memory estimate:  15.46 MiB
  allocs estimate:  2014
  --------------
  minimum time:     15.741 ms (0.00% GC)
  median time:      16.256 ms (0.00% GC)
  mean time:        16.432 ms (2.03% GC)
  maximum time:     18.097 ms (4.95% GC)
  --------------
  samples:          304
  evals/sample:     1


```

I am not sure what I might be doing wrong. 😅

devmotion: As I suspected, `f(x) = mapreduce(x->sin.(x), hcat, eachcol(x))` is not type stable:
```julia
julia> f(x) = mapreduce(x -> sin.(x), hcat, eachcol(x))

julia> @code_warntype f(rand(3, 3))
Variables
  #self#::Core.Compiler.Const(f, false)
  x::Array{Float64,2}
  #32::var"#32#33"

Body::Array
1 ─      (#32 = %new(Main.:(var"#32#33")))
│   %2 = #32::Core.Compiler.Const(var"#32#33"(), false)
│   %3 = Main.eachcol(x)::Base.Generator{Base.OneTo{Int64},Base.var"#194#195"{Array{Float64,2}}}
│   %4 = Main.mapreduce(%2, Main.hcat, %3)::Array
└──      return %4
```
However, the real issue here is https://github.com/JuliaLang/julia/issues/31137. Apparently, `mapreduce(f, ::typeof(hcat), ...)` is not as optimized as `reduce(::typeof(hcat), ...)` which is quite surprising and unexpected IMO. So probably we should just replace the splatting `hcat(x...)` with `reduce(hcat, x)` for now (BTW the latest version of Tracker contains a custom adjoint for `reduce(hcat, x)` and `reduce(vcat, x)`).

Maybe add a comment with a link to the issue in the Julia repo as an explanation.

sharanry: Benchmark doesn't suggest much difference between splatting and using `reduce`. Is there some other reason you are trying to avoid splatting?
```julia
julia> @benchmark reduce(hcat, $([randn(1000) for _ in 1:1000]))
BenchmarkTools.Trial: 
  memory estimate:  7.63 MiB
  allocs estimate:  2
  --------------
  minimum time:     1.278 ms (0.00% GC)
  median time:      1.444 ms (0.00% GC)
  mean time:        2.092 ms (20.13% GC)
  maximum time:     11.189 ms (54.80% GC)
  --------------
  samples:          2377
  evals/sample:     1

julia> @benchmark hcat($([randn(1000) for _ in 1:1000])...)
BenchmarkTools.Trial: 
  memory estimate:  7.65 MiB
  allocs estimate:  8
  --------------
  minimum time:     2.214 ms (0.00% GC)
  median time:      2.437 ms (0.00% GC)
  mean time:        2.854 ms (13.65% GC)
  maximum time:     8.461 ms (64.68% GC)
  --------------
  samples:          1742
  evals/sample:     1

```

I have already made the change you requested by adding a custom adjoint for `reduce`.

devmotion: > However, benchmark doesn't suggest much difference between splatting and using reduce. Is there some other reason you are trying to avoid splatting?

I see a performance difference and a clear difference in number of allocations. I guess https://discourse.julialang.org/t/splatting-arguments-causes-30x-slow-down/16964/2?u=devmotion might be an interesting read.

```julia
julia> f(x) = reduce(hcat, x)

julia> g(x) = hcat(x...)

julia> @btime f($([rand(100) for _ in 1:100]));
  6.052 μs (2 allocations: 78.20 KiB)

julia> @btime g($([rand(100) for _ in 1:100]));
  19.839 μs (6 allocations: 79.11 KiB)

julia> @btime f($([rand(100) for _ in 1:10_000]));
  1.754 ms (2 allocations: 7.63 MiB)

julia> @btime g($([rand(100) for _ in 1:10_000]));
  2.262 ms (8 allocations: 7.86 MiB)
```

src/zygote_adjoints.jl
```diff
@@ -77,6 +77,24 @@ end
...
+            Δ[:, start+1:start+size(xsi,2), i...]
+            end
+            start += size(xsi, 2)
+            d
+        end for xsi in xs]
+        return (nothing, Δs)
+    end
+    return reduce(hcat, xs), back
+end
+
```
devmotion, line 97: The adjoint is already defined in ChainRules (https://github.com/JuliaDiff/ChainRules.jl/blob/026a118a78e7961a31037c689fd3c6b347fa3efa/src/rulesets/Base/array.jl#L42):
```suggestion
```

sharanry: I get a `Mutating arrays is not supported` error if I remove this. :/

sharanry: I probably need to import chain rules.

devmotion: You shouldn't have to import ChainRules, Zygote does it internally. Maybe it's not released yet or you're using an old version?

sharanry: Am I doing something wrong or could this be a bud with `ChainRules`/`Zygote`?

sharanry: > You shouldn't have to import ChainRules, Zygote does it internally. Maybe it's not released yet or you're using an old version?

Just saw this. I will double check and get back,

sharanry: I think it works now. The tests seem to pass.

PR #153, opened 2020-08-13
[WIP] Replace Zygote adjoints with ChainRules' rrules

sharanry:
https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/116#issuecomment-671212997

Currently a lot of AD tests fail. Still need to work on revamping the AD tests to use [ChainRulesTestUtils.jl](https://github.com/JuliaDiff/ChainRulesTestUtils.jl).

TODO

- [ ] Use `rrule_test` for each of the `rrules`. Plain usage of them results in errors with `FiniteDifferences`. Figure of if this is due to the incorrect rrules.
- [ ] `test/utils.jl` seems pass with minimal imports but fail other wise. Figure out why.

PR #154, opened 2020-08-16
[WIP] Fix AD issues with various kernels

sharanry:
#116


sharanry:
I think the reason for failure of Zygote with `MahalanobisKernel` is mutating arrays at 
https://github.com/JuliaStats/Distances.jl/blob/44036b573ec85287022f4368c4e1e279698bd031/src/mahalanobis.jl#L75.

src/basekernels/nn.jl
```diff
@@ -20,7 +20,35 @@ Bayesian neural network with erf (Error Function) as activation function.
 struct NeuralNetworkKernel <: Kernel end
 
 function (κ::NeuralNetworkKernel)(x, y)
-    return asin(dot(x, y) / sqrt((1 + sum(abs2, x)) * (1 + sum(abs2, y))))
+    return asin(dot(x, y) / sqrt((1 + sum(abs2.(x))) * (1 + sum(abs2.(y)))))
```
devmotion, line 23: This should not be needed since `Zygote.gradient(x -> sum(abs2, x), rand(5))` works fine. Revert it to
```suggestion
    return asin(dot(x, y) / sqrt((1 + sum(abs2, x)) * (1 + sum(abs2, y))))
```
?

src/basekernels/nn.jl
```diff
@@ -20,7 +20,35 @@ Bayesian neural network with erf (Error Function) as activation function.
...
+end
+
+function kernelmatrix(::NeuralNetworkKernel, x::ColVecs, y::ColVecs)
```
devmotion, line 26: Was this not working before or did you add it to improve performance?

src/basekernels/nn.jl
```diff
@@ -20,7 +20,35 @@ Bayesian neural network with erf (Error Function) as activation function.
...
+    validate_inputs(x, y)
+    X_2 = sum(x.X .* x.X, dims=1)
+    Y_2 = sum(y.X .* y.X, dims=1)
+    XY = x.X' * y.X
+    return asin.(XY ./ sqrt.((X_2 .+ 1.0)' * (Y_2 .+ 1.0)))
```
devmotion, line 31: One should avoid literals of type Float64 since it will promote the computations (which is not intended, e.g., if you work with Float32).
```suggestion
    return asin.(XY ./ sqrt.((X_2 .+ 1)' * (Y_2 .+ 1)))
```

src/basekernels/nn.jl
```diff
@@ -20,7 +20,35 @@ Bayesian neural network with erf (Error Function) as activation function.
...
+end
+
+function kernelmatrix(::NeuralNetworkKernel, x::ColVecs)
```
devmotion, line 34: Same here, why was it added?

src/basekernels/nn.jl
```diff
@@ -20,7 +20,35 @@ Bayesian neural network with erf (Error Function) as activation function.
...
+    X_2_1 = sum(x.X .* x.X, dims=1) .+ 1.0
```
devmotion, line 35:
```suggestion
    X_2_1 = sum(x.X .* x.X, dims=1) .+ 1
```

src/basekernels/nn.jl
```diff
@@ -20,7 +20,35 @@ Bayesian neural network with erf (Error Function) as activation function.
...
+    XX = x.X' * x.X
+    return asin.(XX ./ sqrt.(X_2_1' * X_2_1))
+end
+
+function kernelmatrix(::NeuralNetworkKernel, x::RowVecs, y::RowVecs)
```
devmotion, line 40: Same here?

src/basekernels/nn.jl
```diff
@@ -20,7 +20,35 @@ Bayesian neural network with erf (Error Function) as activation function.
...
+    validate_inputs(x, y)
+    X_2 = sum(x.X .* x.X, dims=2)
+    Y_2 = sum(y.X .* y.X, dims=2)
+    XY = x.X * y.X'
+    return asin.(XY ./ sqrt.((X_2 .+ 1.0)' * (Y_2 .+ 1.0)))
```
devmotion, line 45:
```suggestion
    return asin.(XY ./ sqrt.((X_2 .+ 1)' * (Y_2 .+ 1)))
```

src/basekernels/nn.jl
```diff
@@ -20,7 +20,35 @@ Bayesian neural network with erf (Error Function) as activation function.
...
+end
+
+function kernelmatrix(::NeuralNetworkKernel, x::RowVecs)
```
devmotion, line 48: Same here?

src/basekernels/nn.jl
```diff
@@ -20,7 +20,35 @@ Bayesian neural network with erf (Error Function) as activation function.
...
+    X_2_1 = sum(x.X .* x.X, dims=2) .+ 1.0
```
devmotion, line 49:
```suggestion
    X_2_1 = sum(x.X .* x.X, dims=2) .+ 1
```

test/basekernels/fbm.jl
```diff
@@ -22,6 +22,6 @@
     @test kernelmatrix(k, x1*ones(1,1), x2*ones(1,1))[1] ≈ k(x1, x2) atol=1e-5
 
     @test repr(k) == "Fractional Brownian Motion Kernel (h = $(h))"
-    test_ADs(FBMKernel, ADs = [:ReverseDiff])
-    @test_broken "Tests failing for kernelmatrix(k, x) for ForwardDiff and Zygote"
+    test_ADs(FBMKernel, ADs = [:ReverseDiff, :Zygote])
+    @test_broken "Tests failing for kernelmatrix(k, x) for ForwardDiff"
```
devmotion, line 26: IMO this is a pretty unfortunate and unintended use of `@test_broken` (which was not introduced in this PR, I know). Instead we should just change `test_ADs` in such a way that instead of specifying the tested ADs we state the broken ADs, i.e.,
```suggestion
    test_ADs(FBMKernel, broken = (:ForwardDiff,))
```
Then `test_ADs` would test all AD backends but expect failing tests for the broken backends and tests should pass for all other backends. In this way it is also much easier to keep track of if some backend is actually working or not (might change, e.g., due to changes in the backends without us noticing otherwise).

test/basekernels/gabor.jl
```diff
@@ -17,7 +17,6 @@
     @test k.ell ≈ 1.0 atol=1e-5
     @test k.p ≈ 1.0 atol=1e-5
     @test repr(k) == "Gabor Kernel (ell = 1.0, p = 1.0)"
-    #test_ADs(x -> GaborKernel(ell = x[1], p = x[2]), [ell, p])#, ADs = [:ForwardDiff, :ReverseDiff])
-    @test_broken "Tests failing for Zygote on differentiating through ell and p"
+    test_ADs(x -> GaborKernel(ell = x[1], p = x[2]), [ell, p], ADs = [:Zygote])
```
devmotion, line 20: Same here, would be good to actually specify the broken backends.

src/basekernels/nn.jl
sharanry: It wasn't. It used to give the error "In slow mode".

src/basekernels/nn.jl
sharanry: Yes I was testing if this was the issue. Forgot to revert it back.

test/basekernels/fbm.jl
sharanry: Yes. It is seems more clean way of doing this. However, I think we should address this in the PR where we are revamping AD rules #153 ? Or in a separate PR? 
I hoped for this PR to just get as many AD tests to pass as possible.

devmotion: I guess it might be good to do this in a separate PR before actually switching to ChainRules in #153 (so we clearly know what's the current state). I don't think it should be part of this PR.

sharanry:
> I think the reason for failure of Zygote with `MahalanobisKernel` is mutating arrays at
> https://github.com/JuliaStats/Distances.jl/blob/44036b573ec85287022f4368c4e1e279698bd031/src/mahalanobis.jl#L75.

@devmotion Do you suggest I override their pairwise implementation for now or open an issue/PR to `Distances.jl`?

devmotion:
I guess that should/could be resolved by adding a custom ChainRules-based adjoint for `pairwise(::Mahalanobis, ...)`, similar to https://github.com/FluxML/Zygote.jl/blob/956575ee2c732dee25324b59ba43fbb471a52d9a/src/lib/distances.jl#L19-L25. It seems it's not yet decided if Distances would accept ChainRules PRs (see https://github.com/JuliaStats/Distances.jl/issues/172), so one could either make a PR to Zygote or add some piracy to KernelFunctions for now, I guess.

test/basekernels/fbm.jl
theogf: > IMO this is a pretty unfortunate and unintended use of `@test_broken` (which was not introduced in this PR, I know). Instead we should just change `test_ADs` in such a way that instead of specifying the tested ADs we state the broken ADs, i.e.,
> 
> Then `test_ADs` would test all AD backends but expect failing tests for the broken backends and tests should pass for all other backends. In this way it is also much easier to keep track of if some backend is actually working or not (might change, e.g., due to changes in the backends without us noticing otherwise).

Sorry that's my fault. I just wanted to make sure that we could remember where tests were failing and not forget them :)

sharanry:
Sorry for the delay. I am having a hard time defining adjoints which aren't very computationally expensive.

devmotion:
I guess you shouldn't need anything in particular for the Mahalanobis *kernel* but rather (just) a custom adjoint for the distance computations in Distances? In this case [the Matrix cookbook](http://www.math.uwaterloo.ca/~hwolkowi//matrixcookbook.pdf), and in particular equations 72 and 81 are helpful. They show you that `d((x-y)'*Q*(x-y))/dx = (Q + Q') * (x - y)`, `d((x-y)'*Q*(x-y))/dy = - (Q + Q') * (x - y)`, and `d((x-y)'*Q*(x-y))/dQ = (x - y)' * (x - y)` (you should recheck that I didn't make any stupid mistakes :smile:). These expressions aren't too bad, I guess, but in general the Mahalanobis distance isn't computationally cheap so it's not too surprising that the adjoints aren't either.

BTW I just noticed that the docstring of the kernel is incorrect since the distance computation (https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/f14404512f245d0bbd60ad737861b7788d12f95b/src/basekernels/maha.jl#L21) uses `P` but not the inverse of `P` (Distances doesn't use the inverse either, according to their README).

sharanry:
@devmotion Thanks for pointing out the typo in the docstring! 
The problem I facing with the adjoint is for pairwise computation. Doing it efficiently seems quite tricky based on the implmenetation here. https://github.com/JuliaStats/Distances.jl/blob/44036b573ec85287022f4368c4e1e279698bd031/src/mahalanobis.jl#L62

src/basekernels/maha.jl
```diff
@@ -20,4 +20,40 @@ kappa(κ::MahalanobisKernel, d::T) where {T<:Real} = exp(-d)
...
+    dims = Distances.deprecated_dims(dims)
+    Q = metric.qmat
+    Qa = Q * a
+    sa2 = dot_perslice(a, Qa; dims=dims)
+    r = a' * Qa
+    r = sa2 .+ sa2' - 2 * r
+    r = Symmetric(r')
+    return r
+  end
+
```
devmotion, line 58: We should just define the adjoint (and it should actually be moved upstream in the long term), there's no need to redefine `pairwise`. The implementation in Distances should be optimized, so most users want to use it and we should also use it for the forward pass in the reverse mode ADs probably (as done in the adjoints in Zygote for `pairwise(::(Sq)Euclidean, ...)`).

sharanry: Yes I just wanted to check if this is equivalent.

Does it make sense to simplify things and first implement adjoint for evaluate even though it is not necessary and use it for the adjoint of pairwise?

```julia
@adjoint function Distances.evaluate(dist::SqMahalanobis, a, b)
    function back(Δ::Real)
        B_B_inv = dist.qmat + inv(dist.qmat)
        a_b = a - b
        δa = B_B_inv * a_b
        return (qmat = a_b * a_b',), δa, -δa 
    end
  return evaluate(dist::SqMahalanobis, a, b), back
end
```

devmotion: I guess we need both so it's fine to implement the one for `evaluate` first. However, you should define an adjoint for `(dist::SqMahalanobis)(a::AbstractVector, b::AbstractVector)` instead of `evaluate` (`evaluate` just forwards to the functional form and hence one should implement the latter to catch all use cases - I fixed this a while ago in Zygote, and I really recommend having a look at the implementations there: https://github.com/FluxML/Zygote.jl/blob/master/src/lib/distances.jl).

However, ultimately I don't think you should use this adjoint for the adjoint of `pairwise` since I assume it will be quite inefficient (recomputing Q + Q' etc.).

BTW the adjoint you wrote down there is incorrect - you have to take the transpose and not the inverse.

src/zygote_adjoints.jl
```diff
@@ -84,3 +84,13 @@ end
 @adjoint function Base.map(t::Transform, X::RowVecs)
     pullback(_map, t, X)
 end
+
+@adjoint function (dist::Distances.SqMahalanobis)(a, b)
+    function back(Δ::Real)
+        B_B_inv = dist.qmat + transpose(dist.qmat)
```
devmotion, line 90: The name `B_B_inv` is a bit strange :slightly_smiling_face:

src/zygote_adjoints.jl
```diff
@@ -84,3 +84,13 @@ end
...
+        a_b = a - b
+        δa = B_B_inv * a_b
+        return (qmat = a_b * a_b',), δa, -δa 
+    end
+  return evaluate(dist::SqMahalanobis, a, b), back
```
devmotion, line 95:
```suggestion
  return evaluate(dist, a, b), back
```

src/zygote_adjoints.jl
```diff
@@ -94,3 +94,27 @@ end
...
+        B_B_t = dist.qmat + transpose(dist.qmat)
+        a_b = map(
+            x -> (first(last(x)) - last(last(x)))*first(x), 
+            zip(
+                Δ,
+                Iterators.product(eachslice(a, dims=dims), eachslice(b, dims=dims))
+            )
+        )
+        δa = reduce(hcat, sum(map(x -> B_B_t*x, a_b), dims=1))
+        δB = sum(map(x -> x*transpose(x), a_b))
```
devmotion: I would assume it should be possible to vectorize this code? What's the mathematical formula that you use here?

sharanry: It is the same same equations you mentioned earlier. 
`d((x-y)'*Q*(x-y))/dx = (Q + Q') * (x - y)`, `d((x-y)'*Q*(x-y))/dy = - (Q + Q') * (x - y)`, and `d((x-y)'*Q*(x-y))/dQ = (x - y)' * (x - y) `.
But this being done for all pairwise combinations together using `map`. It later sums these differences to get `\deltaB` and others. 
Please note that the current implementation is not correct. I am still debugging it. If you happen to find any obvious mistakes please let me know. I am facing trouble in reducing the results of individual pairwise pullbacks to the final pullback. The way I am summing them is probably wrong.

sharanry: I think even `FiniteDifferences` fails for pairwise as the `Q + δQ` computed while getting adjoints using FiniteDifference may no longer be positively-definite. IMO it is best if (Sq)Mahalanobis distance is actually parameterized by the decomposition of `Q`, i.e, the upper or lower triangular matrix which is not constrained.

devmotion: > IMO it is best if (Sq)Mahalanobis distance is actually parameterized by the decomposition of Q, i.e, the upper or lower triangular matrix which is not constrained.

Yes, that would be the most natural way to ensure that it is always positive semi-definite (if the diagonal is non-negative) and optimization is performed in the correct space. So I guess users would want to use this parameterization even if it is not enforced by KernelFunctions and not directly supported by SqMahalanobis by using something like
```julia
function mykernel(L)
    idxs = diagind(L)
    @inbounds for i in idxs
        L[i] = softplus(L[i])
    end
    return MahalanobisKernel(Array(L * L'))
end
```
Of course, it would be nice if (Sq)Mahalanobis would support specifying e.g. a Cholesky decomposition or PDMat directly (it could even be used for simplifying the computations since `x'*Q*x = (L'*x)'*(L'*x)` in this case), but can't we work around this by checking gradients of the `mykernel` setup instead of computing `Q -> MahalanobisKernel(Q)` directly? That's at least how we do it in DistributionsAD, e.g. in https://github.com/TuringLang/DistributionsAD.jl/blob/a96b159ab25aab67d1a2076726e8b9c392eb6fc7/test/ad/distributions.jl#L18-L34.

sharanry: > but can't we work around this by checking gradients of the `mykernel` setup instead of computing `Q -> MahalanobisKernel(Q)` directly?

Yeah that should work. Will try that out.

Regarding the issue with pairwise implementation which messes up `FiniteDifferences` results, do you suggest I override the implementation for the time being?

devmotion: If you test the suggested parameterization it shouldn't matter.

sharanry: True. Could we also change our side of the parametrization? i.e, the way it is stored in the struct. We could continue to allow initialization using a full matrix. This should allow for seamless AD regardless of how the user decides to initialize them.

devmotion: I'm not sure if we want to do that, I think this deserves some discussion first (and then a separate PR possibly). Ideally, Distances would just support arbitrary matrices and contain optimized implementations for specific array types. We just forward `P` to SqMahalanobis, so ideally we wouldn't perform any transformations or computations. I'm also a bit worried that focusing on a specific parameterization might make it difficult for users who would like to use a different one (but still no dense matrix) or might lead to confusing behaviour.

test/basekernels/maha.jl
```diff
@@ -4,13 +4,27 @@
...
+    M1, M2 = rand(rng,3,2), rand(rng,3,2)
+    fdm = FiniteDifferences.Central(5, 1);
+    
+    
+    FiniteDifferences.to_vec(dist::SqMahalanobis{Float64}) = vec(dist.qmat), x -> SqMahalanobis(reshape(x, size(dist.qmat)...))
+
+    @test_broken j′vp(fdm, x -> MahalanobisKernel(Array(x[1]'*x[1]))(x[2], x[3]), 1, [U, v1, v2]) ≈
+    Zygote.pullback(x -> MahalanobisKernel(Array(x[1]'*x[1]))(x[2], x[3]), [U, v1, v2])[2](1)
+    @test all(j′vp(fdm, x -> SqMahalanobis(Array(x[1]'*x[1]))(x[2], x[3]), 1, [U, v1, v2])[1][1] .≈ 
+    Zygote.pullback(x -> SqMahalanobis(Array(x[1]'*x[1]))(x[2], x[3]), [U, v1, v2])[2](1)[1][1])
```
sharanry, line 26: @devmotion I tried doing what you suggested. The tests still fail. This error probably propagates and causes even the first test to fail. 
```julia
julia> j′vp(fdm, x -> SqMahalanobis(Array(x[1]'*x[1]))(x[2], x[3]), 1, [U, v1, v2])[1][1]
3×3 UpperTriangular{Float64,Array{Float64,2}}:
 0.228808   0.00318764   -0.107503
  ⋅        -0.000391803   0.0132135
  ⋅          ⋅            0.0438772

julia> Zygote.pullback(x -> SqMahalanobis(Array(x[1]'*x[1]))(x[2], x[3]), [U, v1, v2])[2](1)[1][1]
3×3 Array{Float64,2}:
  0.228808    0.00318764   -0.107503
 -0.0281234  -0.000391803   0.0132135
 -0.0933875  -0.00130103    0.0438772
```

test/basekernels/maha.jl
```diff
@@ -4,13 +4,27 @@
...
     @test k(v1, v2) ≈ exp(-sqmahalanobis(v1, v2, P))
     @test kappa(ExponentialKernel(), x) == kappa(k, x)
     @test repr(k) == "Mahalanobis Kernel (size(P) = $(size(P)))"
-    test_ADs(P -> MahalanobisKernel(P), P, ADs=[:Zygote])
+
+    M1, M2 = rand(rng,3,2), rand(rng,3,2)
+    fdm = FiniteDifferences.Central(5, 1);
+    
+    
+    FiniteDifferences.to_vec(dist::SqMahalanobis{Float64}) = vec(dist.qmat), x -> SqMahalanobis(reshape(x, size(dist.qmat)...))
```
devmotion, line 21: Is this needed? If possible, we should avoid this type piracy.

test/basekernels/maha.jl
```diff
@@ -4,13 +4,27 @@
...
+
+    @test_broken j′vp(fdm, x -> MahalanobisKernel(Array(x[1]'*x[1]))(x[2], x[3]), 1, [U, v1, v2]) ≈
```
devmotion, line 23: I guess it might be more readable if the transformations or the whole function is defined separately.

test/basekernels/maha.jl
devmotion: To me your output indicates that it basically works apart from the fact that Zygote incorrectly returns a dense matrix instead of an upper triangular matrix. Since `U` was upper triangular, only the values above and on the diagonal should be returned.

theogf (commented):
Somehow the solution to have to define `kernelmatrix` again for the `NeuralNetworkKernel` seems very hacky, isn't there another solution?

test/basekernels/maha.jl
sharanry: Yes `j′vp` only works when there is a `to_vec` function defined for each argument.

devmotion: I'm wondering since according to the docs `to_vec` is only needed for the inputs `xs...` but not the evaluated function `f` in `j'vp(fdm, f, xs...)`.

src/zygote_adjoints.jl
```diff
@@ -84,3 +84,36 @@ end
 @adjoint function Base.map(t::Transform, X::RowVecs)
     pullback(_map, t, X)
 end
+
+@adjoint function (dist::Distances.SqMahalanobis)(a, b)
+    function back(Δ::Real)
+        B_Bᵀ = dist.qmat + transpose(dist.qmat)
+        a_b = a - b
+        δa = B_Bᵀ * a_b
+        return (qmat = a_b * a_b',), δa, -δa
```
devmotion, line 93: Just noticed that this adjoint is incorrect, all outputs should be multiplied with `Δ`.

test/basekernels/maha.jl
sharanry: `FiniteDifferences` if pretty good in matching the types. Zygote isn't. Do you suggest we manually check if the upper triangular part matches for now?

src/zygote_adjoints.jl
sharanry: Thanks!

test/basekernels/maha.jl
sharanry: From what I understand, tt is also needed for objects like `SqMahalanobis` if they have parameters like `qmat`.

test/basekernels/maha.jl
willtebbutt: Were the call to `UpperTriangular` inside the function, then the adjoint that you would get from `Zygote` would also be `UpperTriangular`. Maybe just do that?

test/basekernels/maha.jl
willtebbutt: That's correct, but actually for some reason we've not made `FiniteDifferences` handle functions-with-data properly yet, so you'll have to build the `SqMaha` object inside of the function that you're differentiating.

src/basekernels/nn.jl
```diff
@@ -23,4 +23,32 @@ function (κ::NeuralNetworkKernel)(x, y)
     return asin(dot(x, y) / sqrt((1 + sum(abs2, x)) * (1 + sum(abs2, y))))
 end
 
+function kernelmatrix(::NeuralNetworkKernel, x::ColVecs, y::ColVecs)
+    validate_inputs(x, y)
+    X_2 = sum(x.X .* x.X, dims=1)
+    Y_2 = sum(y.X .* y.X, dims=1)
```
willtebbutt, line 29: Style:
```suggestion
    X_2 = sum(x.X .* x.X; dims=1)
    Y_2 = sum(y.X .* y.X; dims=1)
```

src/basekernels/nn.jl
```diff
@@ -23,4 +23,32 @@ function (κ::NeuralNetworkKernel)(x, y)
...
+    XY = x.X' * y.X
+    return asin.(XY ./ sqrt.((X_2 .+ 1)' * (Y_2 .+ 1)))
+end
+
+function kernelmatrix(::NeuralNetworkKernel, x::ColVecs)
+    X_2_1 = sum(x.X .* x.X, dims=1) .+ 1
```
willtebbutt, line 35:
```suggestion
    X_2_1 = sum(x.X .* x.X; dims=1) .+ 1
```

src/basekernels/nn.jl
```diff
@@ -23,4 +23,32 @@ function (κ::NeuralNetworkKernel)(x, y)
...
+    XX = x.X' * x.X
+    return asin.(XX ./ sqrt.(X_2_1' * X_2_1))
+end
+
+function kernelmatrix(::NeuralNetworkKernel, x::RowVecs, y::RowVecs)
+    validate_inputs(x, y)
+    X_2 = sum(x.X .* x.X, dims=2)
+    Y_2 = sum(y.X .* y.X, dims=2)
```
willtebbutt, line 43:
```suggestion
    X_2 = sum(x.X .* x.X; dims=2)
    Y_2 = sum(y.X .* y.X; dims=2)
```

src/basekernels/nn.jl
```diff
@@ -23,4 +23,32 @@ function (κ::NeuralNetworkKernel)(x, y)
...
+    XY = x.X * y.X'
+    return asin.(XY ./ sqrt.((X_2 .+ 1)' * (Y_2 .+ 1)))
+end
+
+function kernelmatrix(::NeuralNetworkKernel, x::RowVecs)
+    X_2_1 = sum(x.X .* x.X, dims=2) .+ 1
```
willtebbutt, line 49:
```suggestion
    X_2_1 = sum(x.X .* x.X; dims=2) .+ 1
```

src/zygote_adjoints.jl
```diff
@@ -84,3 +84,58 @@ end
 @adjoint function Base.map(t::Transform, X::RowVecs)
     pullback(_map, t, X)
 end
+
+@adjoint function (dist::Distances.SqMahalanobis)(a, b)
+    function back(Δ::Real)
```
willtebbutt, line 89:
```suggestion
    function SqMahalanobis_pullback(Δ::Real)
```

src/zygote_adjoints.jl
```diff
@@ -84,3 +84,58 @@ end
...
+  return evaluate(dist, a, b), back
+end
+
+
+@adjoint function Distances.pairwise(
+    dist::SqMahalanobis, 
+    a::AbstractMatrix, 
+    b::AbstractMatrix;
+    dims::Union{Nothing,Integer}=nothing
+    )
```
willtebbutt, line 104:
```suggestion
)
```

src/zygote_adjoints.jl
```diff
@@ -84,3 +84,58 @@ end
...
+    function back(Δ::AbstractMatrix)
```
willtebbutt, line 105:
```suggestion
    function pairwise_pullback(Δ::AbstractMatrix)
```

src/zygote_adjoints.jl
```diff
@@ -84,3 +84,58 @@ end
...
+        return (qmat=δB,), δa, -δa
+    end
+    return Distances.pairwise(dist, a, b, dims=dims), back
+end
+
+@adjoint function Distances.pairwise(
+  dist::SqMahalanobis, 
+  a::AbstractMatrix;
+  dims::Union{Nothing,Integer}=nothing
+  )
```
willtebbutt, line 125:
```suggestion
    dist::SqMahalanobis, 
    a::AbstractMatrix;
    dims::Union{Nothing,Integer}=nothing
)
```

src/zygote_adjoints.jl
```diff
@@ -84,3 +84,58 @@ end
...
+  function back(Δ::AbstractMatrix)
```
willtebbutt, line 126:
```suggestion
  function pairwise_pullback(Δ::AbstractMatrix)
```

willtebbutt (commented):
This is looking good. Just some style things.

src/zygote_adjoints.jl
```diff
@@ -84,3 +84,58 @@ end
...
+        a_b = map(
+            x -> (first(last(x)) - last(last(x)))*first(x), 
+            zip(
+                Δ,
+                Iterators.product(eachslice(a, dims=dims), eachslice(b, dims=dims))
+            )
+        )
+        δa = reduce(hcat, sum(map(x -> B_Bᵀ*x, a_b), dims=2))
+        δB = sum(map(x -> x*transpose(x), a_b))
+        return (qmat=δB,), δa, -δa
```
devmotion, line 116: There is som discrepancy between the simple case above and this pullback - intuitively, from the simple case above I would assume that `δB = sum_{i, j} (a_i - b_j) * (a_i - b_j)^T * Δ_{i,j}`. However, here you compute `δB = sum_{i, j} (a_i - b_j) * (a_i - b_j)^T * Δ_{i,j}^2`. Probably one of them is incorrect (table 7 in https://notendur.hi.is/jonasson/greinar/blas-rmd.pdf indicates that the pairwise one is incorrect). Can we add the derivation of the adjoints according to https://www.juliadiff.org/ChainRulesCore.jl/dev/arrays.html as docstrings or comments, or maybe even have a separate PR for the Mahalanobis fixes?

devmotion:
> Somehow the solution to have to define `kernelmatrix` again for the `NeuralNetworkKernel` seems very hacky, isn't there another solution?

I guess one could define a "PreMetric" that evaluates `dot(x, y) / sqrt((1 + sum(abs2, x)) * (1 + sum(abs2, y)))` (or `asin(dot(x, y) / sqrt((1 + sum(abs2, x)) * (1 + sum(abs2, y)))))`), similar to `DotProduct`, and make `NeuralNetworkKernel` a `SimpleKernel`. But even in this case one might want to implement (a) specialized version(s) of `pairwise`, so I'm not sure how much one would gain.

src/zygote_adjoints.jl
sharanry: Thanks for pointing this out. I think a separate PR for mahalanobis fixes makes more sense.

sharanry:
Can we merge this and tackle each of the remaining AD issues in separate PRs? It is getting increasingly tricky to address multiple issues at once. 

Currently this PR does the following:
- Defines `kernelmatrix` function for `NeuralNetworkKernel`.
- Defines Zygote adjoints for Mahalanobis distance metric.
- Zygote tests pass for Exponential, FBM, NN and Gabor kernels.

devmotion:
IMO this PR contains already too many changes, we should just focus on one AD problem at a time.

> Defines Zygote adjoints for Mahalanobis distance metric.

I thought the idea was not include these adjoints since they were missing a clean derivation/documentation and were incorrect?

sharanry:
> I thought the idea was not include these adjoints since they were missing a clean derivation/documentation and were incorrect? Or are you talking about the non-pairwise adjoints only?

I meant only the non-pairwise adjoint . I will be removing the pairwise adjoints for now.

test/basekernels/maha.jl
```diff
@@ -4,14 +4,40 @@
     v1 = rand(rng, 3)
     v2 = rand(rng, 3)
 
-    P = rand(rng, 3, 3)
+    U = UpperTriangular(rand(rng, 3,3))
+    P = Matrix(Cholesky(U, 'U', 0))
+    @assert isposdef(P)
     k = MahalanobisKernel(P=P)
-
+
```
devmotion, line 11:
```suggestion

```

src/zygote_adjoints.jl
```diff
@@ -84,3 +84,13 @@ end
...
 end
+
+@adjoint function (dist::Distances.SqMahalanobis)(a, b)
+    function SqMahalanobis_pullback(Δ::Real)
+        B_Bᵀ = dist.qmat + transpose(dist.qmat)
+        a_b = a - b
+        δa = (B_Bᵀ * a_b) * Δ
+        return (qmat = (a_b * a_b') * Δ,), δa, -δa 
+    end
+  return evaluate(dist, a, b), SqMahalanobis_pullback
```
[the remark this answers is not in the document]
sharanry: This entire file seems to have inconsistent indentation. It alternatively uses 2-space and 4-space indentation. Do you suggest I make it uniformly 4-spaced indentation? I am afraid we will lose its version history.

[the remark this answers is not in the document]
devmotion: Let's just fix this one here and address the general formatting issues in a separate PR. History should be fine but reviewing will be more difficult with unrelated changes.

sharanry:
Any objections to merging this?

test/zygote_adjoints.jl
```diff
@@ -20,6 +23,9 @@
     gzsinus = gradient(:Zygote, [x,y]) do xy
         evaluate(KernelFunctions.Sinus(r), xy[1], xy[2])
     end
+    gzsqmaha = gradient(:Zygote, [Q,x,y]) do xy
```
willtebbutt, line 26:
```suggestion
    gzsqmaha = gradient(:Zygote, [Q, x, y]) do xy
```

test/zygote_adjoints.jl
```diff
@@ -36,11 +42,15 @@
     gfsinus = gradient(:FiniteDiff, [x,y]) do xy
         evaluate(KernelFunctions.Sinus(r), xy[1], xy[2])
     end
+    gfsqmaha = gradient(:FiniteDiff, [Q,x,y]) do xy
```
willtebbutt, line 45:
```suggestion
    gfsqmaha = gradient(:FiniteDiff, [Q, x, y]) do xy
```

willtebbutt (approved):
I have no objections other than these tiny style-related things. This is a great PR.

PR #155, opened 2020-08-20
SelectTransform extended to Symbols

molet:
`SelectTransform` functionality is extended to `Symbol`s making it applicable to `AxisArrays`.

### DIFF
--- a/src/transform/selecttransform.jl
+++ b/src/transform/selecttransform.jl
@@ -1,7 +1,8 @@
 """
-    SelectTransform(dims::AbstractVector{Int})
+    SelectTransform(dims::Union{AbstractVector{Int}, AbstractVector{Symbol}})
 
-Select the dimensions `dims` that the kernel is applied to.
+Select the dimensions `dims` that the kernel is applied to. `dims` can be either all
+integers or symbols.
 ```
     dims = [1,3,5,6,7]
     tr = SelectTransform(dims)
@@ -9,17 +10,17 @@ Select the dimensions `dims` that the kernel is applied to.
     transform(tr,X,obsdim=2) == X[dims,:]
 ```
 """
-struct SelectTransform{T<:AbstractVector{Int}} <: Transform
+struct SelectTransform{T<:Union{AbstractVector{Int}, AbstractVector{Symbol}}} <: Transform
     select::T
-    function SelectTransform{V}(dims::V) where {V<:AbstractVector{Int}}
-        @assert all(dims .> 0) "Selective dimensions should all be positive integers"
-        return new{V}(dims)
-    end
 end
 
-SelectTransform(x::T) where {T<:AbstractVector{Int}} = SelectTransform{T}(x)
+function SelectTransform(dims::T) where {T<:AbstractVector{Int}}
+    @assert all(dims .> 0) "Selective dimensions should all be positive integers"
+    return SelectTransform{T}(dims)
+end
 
 set!(t::SelectTransform{<:AbstractVector{T}}, dims::AbstractVector{T}) where {T<:Int} = t.select .= dims
+set!(t::SelectTransform{<:AbstractVector{T}}, dims::AbstractVector{T}) where {T<:Symbol} = t.select .= dims
 
 duplicate(t::SelectTransform,θ) = t
 
--- a/test/transform/selecttransform.jl
+++ b/test/transform/selecttransform.jl
@@ -8,16 +8,41 @@
     x_cols = ColVecs(randn(rng, maximum(select), 6))
     x_rows = RowVecs(randn(rng, 4, maximum(select)))
 
-    @testset "$(typeof(x))" for x in [x_vecs, x_cols, x_rows]
+    Xs = [x_vecs, x_cols, x_rows]
+
+    @testset "$(typeof(x))" for x in Xs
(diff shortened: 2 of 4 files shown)


src/transform/selecttransform.jl
```diff
@@ -1,25 +1,26 @@
...
+integers or symbols.
 ```
     dims = [1,3,5,6,7]
     tr = SelectTransform(dims)
     X = rand(100,10)
     transform(tr,X,obsdim=2) == X[dims,:]
 ```
 """
-struct SelectTransform{T<:AbstractVector{Int}} <: Transform
+struct SelectTransform{T<:Union{AbstractVector{Int}, AbstractVector{Symbol}}} <: Transform
```
devmotion, line 13: Could we maybe be even more general and just use
```suggestion
struct SelectTransform{T} <: Transform
```
?

src/transform/selecttransform.jl
```diff
@@ -1,25 +1,26 @@
...
-        @assert all(dims .> 0) "Selective dimensions should all be positive integers"
-        return new{V}(dims)
-    end
 end
 
-SelectTransform(x::T) where {T<:AbstractVector{Int}} = SelectTransform{T}(x)
+function SelectTransform(dims::T) where {T<:AbstractVector{Int}}
+    @assert all(dims .> 0) "Selective dimensions should all be positive integers"
+    return SelectTransform{T}(dims)
+end
```
devmotion, line 20: In the more general case, we might want to remove this - it would probably also fail for OffsetArrays I assume.

src/transform/selecttransform.jl
```diff
@@ -1,25 +1,26 @@
...
 
 set!(t::SelectTransform{<:AbstractVector{T}}, dims::AbstractVector{T}) where {T<:Int} = t.select .= dims
+set!(t::SelectTransform{<:AbstractVector{T}}, dims::AbstractVector{T}) where {T<:Symbol} = t.select .= dims
```
devmotion, line 23: Maybe just
```suggestion
set!(t::SelectTransform, dims) = t.select .= dims
```
?

src/transform/selecttransform.jl
molet: Sure, I'm happy to make it even more general!

src/transform/selecttransform.jl
```diff
@@ -1,25 +1,20 @@
 """
-    SelectTransform(dims::AbstractVector{Int})
+    SelectTransform(dims::Union{AbstractVector{Int}, AbstractVector{Symbol}})
```
willtebbutt, line 2:
```suggestion
    SelectTransform(dims)
```
to reflect the updated constraints.

willtebbutt (commented):
Broadly LGTM. Thanks for the PR @molet . Just a couple of requests:

1. Could you try running `test_ADs` on a `SelectTransform` with `Symbol`s, just to ensure that it actually does work as intended.
2. Bump the patch version so we can make a release :)

test/transform/selecttransform.jl
```diff
@@ -8,16 +8,90 @@
...
+    tx_col = transform(SEKernel(), SelectTransform([1,3]))
+    ta_col = transform(SEKernel(), SelectTransform([:x,:z]))
+
+    @test kernelmatrix(tx_row, X, obsdim=2) == kernelmatrix(ta_row, A, obsdim=2)
+    @test kernelmatrix(tx_col, X, obsdim=1) == kernelmatrix(ta_col, A, obsdim=1)
+
+    @test kernelmatrix(tx_row, X, Y, obsdim=2) == kernelmatrix(ta_row, A, B, obsdim=2)
+    @test kernelmatrix(tx_col, X, Z, obsdim=1) == kernelmatrix(ta_col, A, C, obsdim=1)
+
+    @testset "$(AD)" for AD in [:Zygote, :ForwardDiff]
```
willtebbutt, line 67: @molet Did the `test_AD` function not work for this code?

molet:
> 1. Could you try running `test_ADs` on a `SelectTransform` with `Symbol`s, just to ensure that it actually does work as intended.

Unfortunately `test_ADs` function creates the input matrices inside so I cannot use that directly to test  `SelectTransform` with `Symbol`s. Anyway, I added a bunch of tests for `Zygote` and `ForwardDiff`. `ReverseDiff` unfortunately breaks - I'm still investigating why.

willtebbutt:
Oh I see. Yes, that is unfortunately -- our AD testing is in need of a re-vamp generally. Good to know that's a thing we've accidentally been assuming.

willtebbutt:
Don't worry too much about `ReverseDiff` for now if it's being a pain -- I would just add a single test using `@test_broken` so we don't forget about it.

willtebbutt (approved):
Will merge when tests pass.

willtebbutt:
Damn. These tests failed because they take too long. @devmotion @theogf any thoughts on the best way to work around this? Do we just add more `@info` statements or something?

devmotion:
Hmm yes we could try that. I guess AxisArrays in general might not be optimized for the use with Zygote - in my experience the package is quite rough and a bit suboptimal and slightly outdated. Maybe we could use some other custom arrays in the tests such as OffsetArrays or AxisKeys.

PR #157, opened 2020-08-20
Use Functors

devmotion:
This PR replaces the current setup with Requires + Flux and overloading `Flux.trainable`
with an explicit dependency on Functors and overloading `Functors.functor` (mostly by using `Functors.@functor`).

IMO the advantage is that 1) we get rid of one optional dependency (I think the whole Requires stuff is a bit meh,
see e.g. https://github.com/JuliaLang/Pkg.jl/issues/1285#issue-476496324) and 2) `Functors.functor` allows not
only Flux but also other optimization or ML packages to make use of our specifications of the parameters
(although AFAIK Flux is the only package that uses the interface currently).

This change is breaking since the order of parameters of `TransformedKernel` is changed to the order of its fields.

The PR also adds missing parameter definitions for `MahalanobisKernel`, `PiecewisePolynomialKernel`, and `TensorProduct`.


theogf (commented):
That's great! Maybe one last thing would be to add a line in the docs to explain how to make a custom kernel trainable?

devmotion:
Good spot, I completely forgot to update the docs. Should be fixed now.

theogf (approved):
Nice! Looks good to me! LGTM

PR #158, opened 2020-08-30
Fix SqExponential and GammaExponential + style

willtebbutt:
Our implementations of the mentioned kernels were quite non-standard, so I've brought them inline with what is usually used.

I've also fixed some style issues. Apologies for conflating the two in a single PR.

The only changes of importance are those in `src/exponential.jl` -- everything else is either style or modifying tests to respect the new convention.


willtebbutt:
Drop in coverage appears to be to do with style-related changes inflating the line count slightly.

src/basekernels/exponential.jl
```diff
@@ -11,7 +11,7 @@ related form of the kernel or [`GammaExponentialKernel`](@ref) for a generalizat
 """
 struct SqExponentialKernel <: SimpleKernel end
 
-kappa(κ::SqExponentialKernel, d²::Real) = exp(-d²)
+kappa(κ::SqExponentialKernel, d²::Real) = exp(-d² / 2)
```
theogf, line 14: I would disapprove of this change:
- This would be a (very) hard breaking change, as this is the most commonly used kernel
- I have seen all conventions being used and I like the fact that in most kernel we did not introduced these artificial constants (which only really help to simplify differentiations and integrations)

src/basekernels/exponential.jl
```diff
@@ -65,10 +65,12 @@ end
 
 @functor GammaExponentialKernel
 
-kappa(κ::GammaExponentialKernel, d²::Real) = exp(-d²^first(κ.γ))
+kappa(κ::GammaExponentialKernel, d::Real) = exp(-d^first(κ.γ))
 
-metric(::GammaExponentialKernel) = SqEuclidean()
+metric(::GammaExponentialKernel) = Euclidean()
```
theogf, line 70: This one is fine by me, but could you add a reference where it is defined like this?

src/matrix/kernelmatrix.jl
```diff
@@ -74,10 +74,7 @@ function kernelmatrix!(K::AbstractMatrix, κ::SimpleKernel, x::AbstractVector)
 end
 
 function kernelmatrix!(
-    K::AbstractMatrix,
-    κ::SimpleKernel,
-    x::AbstractVector,
-    y::AbstractVector,
+    K::AbstractMatrix, κ::SimpleKernel, x::AbstractVector, y::AbstractVector,
 )
```
theogf, line 78: Here and further shouldn't be just one line then?

src/matrix/kernelpdmat.jl
```diff
@@ -3,24 +3,28 @@ using .PDMats: PDMat
 export kernelpdmat
 
 """
-    Compute a positive-definite matrix in the form of a `PDMat` matrix see [PDMats.jl]() with the cholesky decomposition precomputed
-    The algorithm recursively tries to add recursively a diagonal nugget until positive definiteness is achieved or that the noise is too big
+    Compute a positive-definite matrix in the form of a `PDMat` matrix see [PDMats.jl]()
```
theogf, line 6:
```suggestion
    Compute a positive-definite matrix in the form of a `PDMat` matrix see [PDMats.jl](https://github.com/JuliaStats/PDMats.jl)
```

Somehow the link was missing

src/matrix/kernelpdmat.jl
```diff
@@ -3,24 +3,28 @@ using .PDMats: PDMat
...
+    definiteness is achieved or that the noise is too big.
 """
-function kernelpdmat(
-        κ::Kernel,
-        X::AbstractMatrix;
-        obsdim::Int = defaultobs
-        )
-    K = kernelmatrix(κ,X,obsdim=obsdim)
-    Kmax =maximum(K)
+function kernelpdmat(κ::Kernel, X::AbstractMatrix; obsdim::Int=defaultobs)
```
theogf, line 11: This makes me realize that `kernelpdmat` is only defined for matrices. I will open an issue to solve this

src/basekernels/exponential.jl
willtebbutt: Interesting -- I've never seen this used in practice. Could you point me towards some examples?

src/matrix/kernelmatrix.jl
willtebbutt: Nah. It's specifically in reference to [this](https://github.com/invenia/BlueStyle) section of the style guide... that I thought we had officially adopted, but now looking at the README we appear not to have.

I'll open a separate PR that adds it.

theogf: I think that since this is mostly a style PR, you can just add it here.

src/basekernels/exponential.jl
theogf: Googling it I only found this one
- http://pages.cs.wisc.edu/~matthewb/pages/notes/pdf/svms/RBFKernel.pdf
So to be fair in 95% of the cases it is 1/2. 
Do you have a suggestion on how to smooth out the breaking change? Like a warning or something

willtebbutt: I would just make a breaking release (i.e. bump the minor version) and put a note at the top of the README. I'm reluctant to provide a warning because this is on a very heavily used codepath, so it would probably get quite messy quite quickly, not to mention the issues around performance.

theogf: Maybe a warning when the package loads?

willtebbutt: That seems reasonable.

willtebbutt: @theogf I've done as you suggested.

@devmotion @sharanry does this seem like a reasonable approach to both of you? Given that we're going to see this every time that we load the package for a while, I would like to ensure that we're all on board with it.

devmotion: Alternatively, I've seen packages that display warning messages only during updates (see e.g. the discussion in https://github.com/JuliaLang/Pkg.jl/issues/176). Maybe we could do something similar instead to avoid that the message is displayed every time the package is loaded (and even if it is not a direct dependency).

willtebbutt: Right. I've made this change now. @devmotion could you try it out and resolve this conversation if you're happy with it?

devmotion: Seems to work :+1:

willtebbutt:
@theogf @sharanry the test failures appear to be unrelated to this PR. Any idea what's going on?

theogf:
Looks like Gabor fails randomly...

willtebbutt:
That's a bit sad. I'm reluctant to break the build, because that's really no fun for development. Have we got work ongoing to fix Gabor + AD?

devmotion:
It seems to be fine apart from the tests with Julia nightlies, doesn't it? I assume the test errors there are caused by some Zygote and compiler-related issues that we do not have to worry about (at least I know that these Zygote/compiler problems cause test errors of Turing with Julia nightly quite consistently). Maybe best to just allow test failures on Julia nightly?

willtebbutt:
Oh, weird. It was failing on version 1 before. I guess someone re-ran the builds? Assuming that docs passed then this is good to go.

theogf:
Yeah I reran the failing build

theogf (approved):

PR #159, opened 2020-08-30
test utils revamp

willtebbutt:
Introduces and applies a large number of consistency tests.

The advantage of running these tests is that we can be fairly sure that new kernels are at least self-consistent, and fully implement our interface.

The advantage of producing a new `TestUtils` module is that if someone implements a new kernel outside of `KernelFunctions`, they can straightforwardly test that the kernel obeys our interface.

This can't be merged until the following PRs are merged:

- [x] #154
- [x] #158


sharanry (commented):
I was going through this to understand the new testing strategy. Looks great! But, have a few doubts. :)

src/test_utils.jl
```diff
@@ -0,0 +1,297 @@
...
+    x2::AbstractVector;
+    atol=__ATOL,
+)
+    # TODO: uncomment the tests of ternary kerneldiagmatrix.
+
+    # Ensure that we have the required inputs.
+    @assert length(x0) == length(x1)
+    @assert length(x0) ≠ length(x2)
+    @assert length(x0) == length(x1)
+    @assert length(x0) ≠ length(x2)
```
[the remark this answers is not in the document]
theogf: Aren't the two vectors of different size and therefore should be checked?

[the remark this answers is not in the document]
willtebbutt: Indeed it has!

src/test_utils.jl
```diff
@@ -0,0 +1,297 @@
...
+        RowVecs(randn(rng, T, 2, dim_in));
+        kwargs...,
+    )
+end
+
+function test_interface(k::Kernel, T::Type{<:AbstractVector}; kwargs...)
+    test_interface(Random.GLOBAL_RNG, k, T; kwargs...)
+end
+
+function test_interface(rng::AbstractRNG, k::Kernel, T::Type{<:Real}; kwargs...)
```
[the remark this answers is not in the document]
willtebbutt: Right. So this functionality is meant to just test common cases -- we definitely _can_ test them when the element type is non-real, we just don't know how to automatically generate candidate tests.

So if you wanted to test some arbitrary input type, or some type of `AbstractVector` that's not a `Vector`, `ColVecs`, or `RowVecs`, you would manually construct the inputs and invoke the method on line 30.

My goal here was to make it so that 95% of the time you can just say "test all of the usual things" really easily, but retain the flexibility to do specific things if you need to.

src/test_utils.jl
```diff
@@ -0,0 +1,297 @@
...
+    params_kernel = params(kernel)
+    params_reference = params(reference)
+
+    @test length(params_kernel) == length(params_reference)
+    @test all(p == q for (p, q) in zip(params_kernel, params_reference))
+end
+
+# AD utilities
+
+const FDM = FiniteDifferences.central_fdm(5, 1)
```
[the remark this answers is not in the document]
willtebbutt: Haha I'm not really sure. It generally seems to work well 🤷

src/test_utils.jl
```diff
@@ -0,0 +1,297 @@
...
+function compare_gradient(f, AD::Symbol, args)
+    grad_AD = gradient(f, AD, args)
+    grad_FD = gradient(f, :FiniteDiff, args)
+    @test grad_AD ≈ grad_FD atol=1e-8 rtol=1e-5
+end
+
+testfunction(k, A, B, dim) = sum(kernelmatrix(k, A, B, obsdim = dim))
+testfunction(k, A, dim) = sum(kernelmatrix(k, A, obsdim = dim))
+
+function test_ADs(kernelfunction, args = nothing; ADs = [:Zygote, :ForwardDiff, :ReverseDiff], dims = [3, 3])
```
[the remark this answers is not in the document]
willtebbutt: This seems like a good proposal. It's out of scope in this PR though -- I've just copied the AD tests.

src/transform/lineartransform.jl
```diff
@@ -29,7 +29,7 @@ end
 (t::LinearTransform)(x::Real) = vec(t.A * x)
 (t::LinearTransform)(x::AbstractVector{<:Real}) = t.A * x
 
-_map(t::LinearTransform, x::AbstractVector{<:Real}) = ColVecs(t.A * x')
+_map(t::LinearTransform, x::AbstractVector{<:Real}) = ColVecs(t.A * collect(x'))
```
devmotion, line 32: Why is this needed? For AD?

src/test_utils.jl
```diff
@@ -0,0 +1,295 @@
+module TestUtils
+
+const __ATOL = 1e-9
+
+using LinearAlgebra
```
devmotion, line 5:
```suggestion
using FiniteDifferences
using LinearAlgebra
```

src/transform/lineartransform.jl
willtebbutt: Actually no. I forget precisely which kernel it was that was causing problems, but somewhere this was breaking.

I should probably write a test that properly isolates the issue...

theogf (dismissed):

willtebbutt:
Moved some tests back to `test/test_utils.jl` because they introduce additional deps, and I would like to avoid this.

willtebbutt:
@theogf could you comment on the `GammaRationalQuadraticKernel` implementation? It appears to suffer from the same issues that the `GammaExponentialKernel` had prior to #158 , and was recently mentioned in #170 .

What are your thoughts on how it should be modified? Currently the test produce covariance matrices with negative eigenvalues, so something needs to be modified. Would you mind taking a look?

willtebbutt:
@sharanry the `NeuralNetworkKernel` implementation appears to be inconsistent between various implementations. Would you be able to pull this branch and take a look? I believe that you were the last one to look at that code.

devmotion:
Along these lines, I think it would make sense to check in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/744419e49e278ef74ec127f861ea7f42f18246e8/src/basekernels/exponential.jl#L64 that `0 < γ <= 2` (alternatively, one could use the SqEuclidean distance and force `γ` to `(0, 1]` - but I guess it was changed to Euclidean on purpose in the previous PR to be more in line with common parameterizations; as an additional option, one could just store/use `γ/2` internally).

willtebbutt:
> as an additional option, one could just store/use γ/2 internally

This is what I was thinking to be honest. There definitely seem to be computational / numerical advantages to sticking with `SqEuclidean` over `Euclidean` wherever possible.

theogf:
To be honest I think most of the code legacy is from `MLKernels.jl`. I don't mind them being modified... I think the issue is that with `gamma` up, the function becomes very quickly flat around 0, so you end up with a singular matrix...

devmotion:
It's not only a numerical problem though: for any Hilbertian distance `|.|` (such as the Euclidean norm on standard Euclidean vector spaces), mathematically the function `k(x, y) = exp(-|x-y|^s)` is positive definite if and only if `0 < s <= 2`.

willtebbutt:
Actually @theogf do you have a reference for the Gamma RQ kernel? Thinking about it I've not seen it before.

theogf:
It's also MLKernels legacy https://www.github.com/trthatcher/MLKernels.jl/tree/master/src%2Fkernelfunctions%2Fmercer%2Frationalquadratic.jl

src/basekernels/rationalquad.jl
```diff
@@ -1,48 +1,63 @@
...
     α::Vector{Tα}
     γ::Vector{Tγ}
-    function GammaRationalQuadraticKernel(;alpha::Tα=2.0, gamma::Tγ=2.0, α::Tα=alpha, γ::Tγ=gamma) where {Tα<:Real, Tγ<:Real}
+    function GammaRationalQuadraticKernel(
+        ;alpha::Tα=2.0, gamma::Tγ=2.0, α::Tα=alpha, γ::Tγ=gamma,
+    ) where {Tα<:Real, Tγ<:Real}
         @check_args(GammaRationalQuadraticKernel, α, α > one(Tα), "α > 1")
-        @check_args(GammaRationalQuadraticKernel, γ, γ >= one(Tγ), "γ >= 1")
+        @check_args(GammaRationalQuadraticKernel, γ, γ <= 2, "γ <= 2")
+        @check_args(GammaRationalQuadraticKernel, γ, γ > zero(Tγ), "γ > 0")
```
devmotion, line 48:
```suggestion
        @check_args(GammaRationalQuadraticKernel, γ, zero(γ) < γ <= 2 * one(γ), "0 < γ <= 2")
```

willtebbutt:
Tests are now passing locally.

Assuming that CI passes, all that remains is to assess whether the changes that I made to the rational quadratic and gamma rational quadratic kernels are satisfactory. If so, I propose to bump the patch version and merge.

willtebbutt:
@theogf what are your thoughts on this? Note that I've rejected your previous review because I've made changes to `RationalQuadratic` that need another round of review 🙂

src/test_utils.jl
```diff
@@ -0,0 +1,135 @@
...
+be of different lengths.
+
+    test_interface([rng::AbstractRNG], k::Kernel, T::Type{<:AbstractVector}; atol=__ATOL)
+
+`test_interface` offers certain types of test data generation to make running these tests
+require less code for common input types. For example, `Vector{<:Real}`, `ColVecs{<:Real}`,
+and `RowVecs{<:Real}` are supported. For other input vector types, please provide the data
+manually. 
+"""
+function test_interface(
```
theogf, line 30: I am not a native speaker so I might miss the meaning but I find `test_interface` to be a right name, it sounds to me that the function is going to create a test interface, not run tests.
Shouldn't it be something like `test_kernel` instead?

src/test_utils.jl
```diff
@@ -0,0 +1,135 @@
...
+    # Check that unary pairwise basically works.
+    @test kernelmatrix(k, x0) isa AbstractMatrix
+    @test size(kernelmatrix(k, x0)) == (length(x0), length(x0))
+    @test kernelmatrix(k, x0) ≈ kernelmatrix(k, x0)' atol=atol
+
+    # Check that unary elementwise is consistent with unary pairwise.
+    @test kerneldiagmatrix(k, x0) ≈ diag(kernelmatrix(k, x0)) atol=atol
+
+    # Check that unary pairwise produces a positive definite matrix (approximately).
+    @test all(eigvals(Matrix(kernelmatrix(k, x0))) .> -atol)
```
theogf, line 71: I know that this should be true, but I think that might be a test being too hard, most kernels will need some noise on the diagonal to be properly PD

theogf: Oh now I see that you allow negative values

src/test_utils.jl
willtebbutt: `test` in `test_interface` is  verb in this context, so `test_interface` means "test the interface". Is that clearer?

test/kernels/kernelsum.jl
```diff
@@ -32,40 +32,12 @@
...
-
-            tmp_diag = Vector{Float64}(undef, length(x))
-            @test kerneldiagmatrix!(tmp_diag, k, x) ≈ kerneldiagmatrix(k, x)
-        end
-    end
-    test_ADs(x->KernelSum(SqExponentialKernel(),LinearKernel(c= x[1])), rand(1), ADs = [:ForwardDiff, :ReverseDiff, :Zygote])
+    # Standardised tests.
+    TestUtils.test_interface(k, Float64)
+    test_ADs(
+        x->KernelSum(SqExponentialKernel(),LinearKernel(c= x[1])), rand(1);
```
theogf, line 38:
```suggestion
        x->KernelSum(SqExponentialKernel(), LinearKernel(c=x[1])), rand(1);
```

src/test_utils.jl
devmotion, line 71: An (IMO simpler) alternative:
```suggestion
    @test eigmin(Matrix(kernelmatrix(k, x0))) > -atol
```

theogf (commented):
I just put a few comments (sorry for the single comments, I messed up).

You really did an amazing titan job! That looks really good!

Also should we have tests on `test_utils` itself? Making sure that it goes wrong for a wrongly defined kernel?

src/test_utils.jl
theogf: Ok got it!

src/test_utils.jl
```diff
@@ -0,0 +1,135 @@
...
+"""
+function test_interface(
+    k::Kernel,
+    x0::AbstractVector,
+    x1::AbstractVector,
+    x2::AbstractVector;
+    atol=__ATOL,
+)
+    # TODO: uncomment the tests of ternary kerneldiagmatrix.
+
```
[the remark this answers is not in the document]
willtebbutt: Hmm I'm reluctant to do this in this PR. The user has plenty of feedback when the tests fail in the way that the test sets are printed, and I personally prefer to minimise output during the running of tests.

willtebbutt:
For some reason I can't reply to this directly:

> Has this issue been solved? That would be great to avoid the allocation

No, unfortunately not. Would you be open to me opening a separate issue about it and pressing ahead with getting this PR merged?

willtebbutt:
> Also should we have tests on test_utils itself? Making sure that it goes wrong for a wrongly defined kernel?

Hmmm maybe. As with the above, I wonder whether we could postpone this to a future PR where we introduce some proper test fakes or something? This is already quite large :S

willtebbutt:
> Hmmm maybe. As with the above, I wonder whether we could postpone this to a future PR where we introduce some proper test fakes or something? This is already quite large :S

@theogf thoughts on this? (Sorry for being pushy -- just keen to get this merged so that I can press ahead with my Stheno.jl / TemporalGPs.jl integration plan)

theogf:
Sorry yeah sure, we can do this later! Let's just keep an issue open for this

willtebbutt:
> Sorry yeah sure, we can do this later! Let's just keep an issue open for this

Great. I'll open an issue about those two things. Then we're good to go?

willtebbutt:
If someone is happy to approve, I'll get this merged after a patch bump.

src/basekernels/rationalquad.jl
```diff
@@ -1,48 +1,62 @@
...
 ```
-where `α` is a shape parameter of the Euclidean distance. Check [`GammaRationalQuadraticKernel`](@ref) for a generalization.
+where `α` is a shape parameter of the Euclidean distance. Check
+[`GammaRationalQuadraticKernel`](@ref) for a generalization.
 """
 struct RationalQuadraticKernel{Tα<:Real} <: SimpleKernel
     α::Vector{Tα}
     function RationalQuadraticKernel(;alpha::T=2.0, α::T=alpha) where {T}
-        @check_args(RationalQuadraticKernel, α, α > zero(T), "α > 1")
+        @check_args(RationalQuadraticKernel, α, α > one(T), "α > 1")
```
devmotion, line 14: I don't think `α` has or should be restricted to `α > 1`. Every non-negative `α` is fine, isn't it?

Additionally, we could think about defining it as
```
κ(x, y) = (1 + ||x − y||² / (2 * α))^(-α)
```
such that the Gaussian kernel is retrieved in the limit as `α` goes to infinity (see, e.g., https://www.cs.toronto.edu/~duvenaud/cookbook/).

devmotion (changes requested):

src/basekernels/rationalquad.jl
devmotion: Oh I just saw that you actually changed it to the suggested form - it seems the docstring is outdated then.

devmotion: Maybe for consistency reasons, the same changes should be applied to the Gamma rational quadratic kernel as well? There still the form without the additional 2 is used, and the parameter is also restricted to values larger than 1.

willtebbutt: Docstring and dof-constraints now fixed.

> Maybe for consistency reasons, the same changes should be applied to the Gamma rational quadratic kernel as well?

I thought about doing this, but the `GammaExponential` kernel _doesn't_ use the divide-by-2 convention, because it's not really used anywhere in the literature. Since the Gamma RQ is constructed from that kernel, it feels more appropriate to me to leave this one _without_ the divide-by-2 convention.

> such that the Gaussian kernel is retrieved in the limit as α goes to infinity

One the subject of this, please note the tests -- they now include various dof->infinity consistency checks for both RQ and Gamma-RQ kernels :)

devmotion: Yep, I agree with your opinion on the Gamma rational quadratic kernel, I just assumed it's good to bring it up now.

I like these additional tests!

src/basekernels/rationalquad.jl
```diff
@@ -1,48 +1,62 @@
 """
-    RationalQuadraticKernel(; α = 2.0)
+    RationalQuadraticKernel(; α=2.0)
```
devmotion, line 2: Is it common to use `α = 2`? Something like `α = 1` seems simpler, and is actually used as the default value by e.g. [scikit-learn](https://scikit-learn.org/stable/modules/generated/sklearn.gaussian_process.kernels.RationalQuadratic.html).

src/basekernels/rationalquad.jl
```diff
@@ -1,48 +1,62 @@
...
 where `α` is a shape parameter of the Euclidean distance and `γ` is another shape parameter.
 """
 struct GammaRationalQuadraticKernel{Tα<:Real, Tγ<:Real} <: SimpleKernel
     α::Vector{Tα}
     γ::Vector{Tγ}
-    function GammaRationalQuadraticKernel(;alpha::Tα=2.0, gamma::Tγ=2.0, α::Tα=alpha, γ::Tγ=gamma) where {Tα<:Real, Tγ<:Real}
-        @check_args(GammaRationalQuadraticKernel, α, α > one(Tα), "α > 1")
-        @check_args(GammaRationalQuadraticKernel, γ, γ >= one(Tγ), "γ >= 1")
+    function GammaRationalQuadraticKernel(
+        ;alpha::Tα=2.0, gamma::Tγ=2.0, α::Tα=alpha, γ::Tγ=gamma,
```
devmotion, line 44: I guess
```suggestion
        ;alpha::Tα=2.0, gamma::Tγ=2, α::Tα=alpha, γ::Tγ=gamma,
```
would be more performant in the default case? The question here is also if we should use `alpha = 1` as
default value.

devmotion: Although probably, since we divide gamma by 2 anyway in the computation it doesn't matter.

src/basekernels/rationalquad.jl
willtebbutt: Hmmm good question. I really hadn't thought too much about this and wasn't overly fussed. I worry that 1 will yield a kernel with very long-range correlations (I'm just thinking about what a students-t with dof 1 looks like -- I think they coincide in this case).

I can't say that I'm overly fussed overall, so happy to change to 1 from 2 if you would prefer.

devmotion: I don't have a strong opinion on this, so do whatever you think is more reasonable. Just noticed this when I compared it with the parameter choices in scikit-learn, and was wondering why we use 2 here.

src/basekernels/rationalquad.jl
willtebbutt: Hmmm I wonder whether this is what you want. Generally speaking, if you're using this kernel you probably don't want the `Integer` parameter value, otherwise you would be using the `RationalQuadraticKernel`. My opinion is that making this parameter a float by default probably does make sense as a consequence of this.

devmotion: Makes sense.

src/basekernels/rationalquad.jl
willtebbutt: I think I'm going to leave as-is for the sake of this PR. More than happy for this to be changed in a subsequent PR though.

devmotion (approved):

PR #165, opened 2020-09-10
Use nansafe branch for ForwardDiff

sharanry:
Related issue and comment: https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/116#issuecomment-690285893
Fix proposed here: https://github.com/JuliaDiff/ForwardDiff.jl/pull/451

This uses the `nansafe` branch of ForwardDiff in order to fix issues with FBM kernel.


test/Project.toml
```diff
@@ -18,7 +18,7 @@ AxisArrays = "0.4.3"
 Distances = "0.9"
 FiniteDifferences = "0.10.8"
 Flux = "0.10, 0.11"
-ForwardDiff = "0.10"
+ForwardDiff = "nansafe"
```
sharanry, line 21: This doesn't seem to help as suggested [here](https://discourse.julialang.org/t/specifying-a-branch-of-a-package-in-project-toml/32101/3).

sharanry: The other suggested solution is including this information in the `test/Manifest.toml`. But I am not sure if we want to do that.
https://discourse.julialang.org/t/specifying-a-branch-of-a-package-in-project-toml/32101/3

devmotion: I would be surprised if there would be a way to specify in the Project.toml, and in your links it is explicitly stated that it does not work currently.

An alternative would be to add something like
```julia
import Pkg; Pkg.add(url="https://github.com/JuliaDiff/ForwardDiff.jl", rev="nansafe")
```
to `test/runtests.jl` (or the Travis script, but I guess it would be better to add it to runtests.jl since otherwise tests would fail locally).

sharanry: Looks like `Pkg` is not available in the Travis test environment for `runtests.jl` to use. https://travis-ci.com/github/JuliaGaussianProcesses/KernelFunctions.jl/jobs/383271248#L421

We might be forced to add this in the Travis script.

devmotion: You just have to add it to the test dependencies, then the lines above will work.

sharanry: @devmotion Do you mean we need to add `Pkg` to test dependencies?

devmotion: Yes.

I'm not sure yet though if it's really a good idea to use the nansafe branch for testing.

test/runtests.jl
```diff
@@ -8,6 +8,8 @@ using Random
 using SpecialFunctions
 using Test
 using Flux
+
+using Pkg; Pkg.add(url="https://github.com/JuliaDiff/ForwardDiff.jl", rev="nansafe")
```
devmotion, line 12: Apparently
```suggestion
using Pkg
Pkg.add(PackageSpec(; url="https://github.com/JuliaDiff/ForwardDiff.jl", rev="nansafe"))
```
is needed in older Julia versions.

devmotion:
It seems tests on Travis still yield NaN values even with this branch - it doesn't fix (all of) the AD problems?

sharanry:
> It seems tests on Travis still yield NaN values even with this branch - it doesn't fix (all of) the AD problems?

Locally it seems to fix problems only for FBM, Gabor and Wiener Kernels.

PR #167, opened 2020-09-11
Allow to trigger TagBot manually

devmotion:
~~This PR adds a key so TagBot can trigger the build of the docs every time a new release is tagged.~~ Additionally, this PR allows to run TagBot manually (in case you don't want to wait for it or it failed).

### DIFF
--- a/.github/workflows/TagBot.yml
+++ b/.github/workflows/TagBot.yml
@@ -9,3 +9,4 @@ jobs:
       - uses: JuliaRegistries/TagBot@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
+          ssh: ${{ secrets.DOCUMENTER_KEY }}
--- a/.github/workflows/TagBot.yml
+++ b/.github/workflows/TagBot.yml
@@ -1,7 +1,8 @@
 name: TagBot
 on:
   schedule:
-    - cron: 0 * * * *
+    - cron: 0 0 * * *
+  workflow_dispatch:
 jobs:
   TagBot:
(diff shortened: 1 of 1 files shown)


devmotion:
Oh I missed that docs are currently built with Travis and not Github actions :smile: I'll revert the key change and just keep the manual trigger then.

willtebbutt:
Could you explain how the change works a little? I don't know enough about github actions to know what they're going to do 😂

devmotion:
Sure, so initially I thought the problem with not getting any versioned documentation (whenever a new tag is created) is that the action does not have the rights to trigger building the docs. Only later I realized that actually currently we build the docs on Travis (and I'm not sure what the settings are there and why it doesn't work, I don't have any permissions on Travis it seems). Maybe we could move building the docs to Github actions and add the key (that I removed again in this PR), to check if that resolves the current problems.

The only change in this PR that's left is how (and when) TagBot is triggered: with the new settings it would just run once daily on midnight (instead of every hour) and [one could trigger it manually](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/), in case you, e.g., want to tag the latest release immediately. That's also the default settings suggested in the TagBot repo.

willtebbutt:
> with the new settings it would just run once daily on midnight (instead of every hour) and one could trigger it manually, in case you, e.g., want to tag the latest release immediately.

Oh, nice! Sounds good to me!

willtebbutt (approved):

PR #168, opened 2020-09-15
Make Zygote work for Wiener kernel

sharanry:
#116 

I have defined `kernelmatrix` for wiener kernel. This avoid the "slow method" ~~but gives `NaN`s in the Zygote gradient~~.


src/basekernels/wiener.jl
```diff
@@ -75,4 +75,42 @@ function (::WienerKernel{3})(x, y)
         ( 5 * max(X, Y)^2 + 2 * X * Y + 3 * minXY^2 )
 end
 
+function kernelmatrix(::WienerKernel{I}, x::ColVecs, y::ColVecs) where I
+    validate_inputs(x, y)
+    X = sqrt.(sum(x.X .* x.X; dims=1))
+    Y = sqrt.(sum(y.X .* y.X; dims=1))
+    minXY = min.(permutedims(X), Y)
```
devmotion, line 82: Can't we just use
```suggestion
    X = sum(x.X .* x.X; dims=1)
    Y = sum(y.X .* y.X; dims=1)
    minXY = sqrt.(min.(permutedims(X), Y))
```
here, or maybe even not take the square root at all in cases where we compute (mostly) `minXY.^2` and `minXY.^4` such as `I == 3`?

src/basekernels/wiener.jl
```diff
@@ -75,4 +75,42 @@ function (::WienerKernel{3})(x, y)
...
+            ( 5 .* max.(permutedims(X), Y).^2 .+ 2 .* X .* Y .+ 3 .* minXY.^2 )
+    end
+    return error("Invalid I=$I")
+end
+
+function kernelmatrix(::WienerKernel{I}, x::RowVecs, y::RowVecs) where I
+    validate_inputs(x, y)
+    X = sqrt.(sum(x.X .* x.X; dims=2))
+    Y = sqrt.(sum(y.X .* y.X; dims=2))
+    minXY = min.(permutedims(X), Y)
```
devmotion, line 101: Same here,
```suggestion
    X = sum(x.X .* x.X; dims=2)
    Y = sum(y.X .* y.X; dims=2)
    minXY = sqrt.(min.(permutedims(X), Y))
```
and maybe drop the `sqrt` completely for certain values of `I`?

src/basekernels/wiener.jl
sharanry: We need the squareroot-ed X and Y for I = 2 and 3. We also have odd powers of X and Y for I = 1, 2, 3.

devmotion: Sure but one could specialize on `I` (similar to the scalar definitions which could be changed as well). In any case, taking the square root only once (as done in the suggested change) is definitely more efficient than taking the square root of both `X` and `Y` separately.

sharanry: I am slightly confused. Taking square root for X and Y separate results in `2n` operations whereas taking squareroot on `minXY` would need `n*n` operations. Am I missing something?

devmotion: Oh haha, of course you're right! I completely missed the `permutedims` :sweat_smile: So the comment does not apply to the `kernelmatrix` implementation but rather to the existing one with individual samples.

sharanry:
The `NaN` values in the Zygote gradient seem to be caused by the the `pairwise(Euclidean(), x, y)`. I am not sure why this is happening.

src/basekernels/wiener.jl
```diff
@@ -75,4 +75,42 @@ function (::WienerKernel{3})(x, y)
...
+    if I == 0
+        return minXY
+    elseif I == 1
+        return (1 / 3) .* minXY.^3 .+ (1 / 2) .* minXY.^2 .* Distances.pairwise(Euclidean(), x, y)
```
devmotion, line 105: Why did you switch to `Distances.pairwise`? This will dispatch to Distances anyways.

devmotion:
Can you try the Zygote master branch locally? Maybe it was fixed by https://github.com/FluxML/Zygote.jl/pull/787.

sharanry:
> Can you try the Zygote master branch locally? Maybe it was fixed by [FluxML/Zygote.jl#787](https://github.com/FluxML/Zygote.jl/pull/787).

Thanks! The Zygote and ForwardDiff tests pass on master. However, ReverseDiff fails, probably requires nan checking.

src/basekernels/wiener.jl
sharanry: I have reverted this. I was just trying stuff to get the tests to pass. Left this mistakenly.

src/basekernels/wiener.jl
```diff
@@ -47,15 +47,15 @@ function WienerKernel(;i::Integer=0)
...
+    return sqrt(min(X_2, Y_2))
 end
 
 function (::WienerKernel{1})(x, y)
-    X = sqrt(sum(abs2, x))
-    Y = sqrt(sum(abs2, y))
-    minXY = min(X, Y)
+    X_2 = sum(abs2, x)
+    Y_2 = sum(abs2, y)
+    minXY = sqrt(min(X_2, Y_2))
```
devmotion, line 58: Here one could actually think about not computing the square root and instead computing `minX2Y2^(3/2)` or `sqrt(minX2Y2)^3` in the next line, since the other term is just `minXY^2`.

src/basekernels/wiener.jl
```diff
@@ -47,16 +47,16 @@ function WienerKernel(;i::Integer=0)
...
 
 function (::WienerKernel{1})(x, y)
-    X = sqrt(sum(abs2, x))
-    Y = sqrt(sum(abs2, y))
-    minXY = min(X, Y)
-    return 1 / 3 * minXY^3 + 1 / 2 * minXY^2 * euclidean(x, y)
+    X_2 = sum(abs2, x)
+    Y_2 = sum(abs2, y)
+    minX2Y2 = min(X_2, Y_2)
+    return 1 / 3 * minX2Y2^(3/2) + 1 / 2 * minX2Y2 * euclidean(x, y)
```
devmotion, line 59: Good opportunity to get rid of any unintended promotions as well:
```suggestion
    return minX2Y2^(3/2) / 3 + minX2Y2 * euclidean(x, y) / 2
```

devmotion: BTW I just benchmarked `f(x) = sqrt(x)^3` and `g(x) = x^(3/2)`, and the first one is much faster (due to some compiler optimizations I guess) so you definitely want to use
```suggestion
    return sqrt(minX2Y2)^3 + minX2Y2 * euclidean(x, y) / 2
```

devmotion:
There's a new Zygote version which contains the fix mentioned above.

src/basekernels/wiener.jl
```diff
@@ -47,32 +47,70 @@ function WienerKernel(;i::Integer=0)
...
+    elseif I == 1
+        return minXY.^3 ./ 3 .+ minXY.^2 .* pairwise(Euclidean(), x, y) ./ 2
+    elseif I == 2
+        return minXY.^5 ./ 20 .+ minXY.^3 .* pairwise(Euclidean(), x, y) .*
+            ( X + Y .- minXY ./ 2 ) ./ 12
+    elseif I == 3
+        return minXY.^7 ./ 252 .+ minXY.^4 .* pairwise(Euclidean(), x, y) .*
+            ( 5 .* max.(permutedims(X), Y).^2 .+ 2 .* X .* Y .+ 3 .* minXY.^2 ) ./ 720
+    end
+    return error("Invalid I=$I")
```
devmotion, line 94: The constructor guarantees that this can't be reached. 

Alternatively, (and maybe a bit nicer) might be to dispatch depending on the kernel (similar to `(::WienerKernel)(x, y)`), maybe just for computing the result from `minXY` since this part is always the same.

src/basekernels/wiener.jl
```diff
@@ -47,32 +47,70 @@ function WienerKernel(;i::Integer=0)
...
```
devmotion, line 113: This can't be reached either.

test/basekernels/wiener.jl
```diff
@@ -1,4 +1,5 @@
 @testset "wiener" begin
+    rng = MersenneTwister(123)
```
devmotion, line 2: Alternatively, we could just set a seed in runtests.jl and use the global RNG - the seed is reset in every `@testset`.

test/test_utils.jl
```diff
@@ -10,7 +10,7 @@ end
 
 # AD utilities
 
-const FDM = FiniteDifferences.central_fdm(5, 1)
+const FDM = FiniteDifferences.central_fdm(7, 1)
```
devmotion, line 13: Do tests fail otherwise?

devmotion (commented):
Looks good to me, just had some minor comments/questions left. The only major thing that is missing is the version bump it seems :slightly_smiling_face:

test/basekernels/wiener.jl
sharanry: Currently, the tests are not uniform in this regard. This probably warrants a separate PR.

src/basekernels/wiener.jl
sharanry: Do you suggest dispatching just the if statement part? Problem with that is such a function would need five inputs - the kernel, x, y, X, Y, minXY. I think this will be a very untidy way to do this.

test/test_utils.jl
sharanry: Yes - some times. One of the reasons why I set the seed. I think this should solve some of the AD issues in the other kernels too.

sharanry:
It is weird that Zygote on Gabor kernel tests are failing on Mac without any changes related to it. https://travis-ci.com/github/JuliaGaussianProcesses/KernelFunctions.jl/jobs/386076575#L369

test/test_utils.jl
devmotion: Ah interesting!

test/basekernels/wiener.jl
devmotion: Yep, I agree this would be better in a separate PR. I mainly commented on this since I saw that you changed large parts of the test file and added a specific RNG - I'm not sure if this is actually useful since the default RNG will be reset in every test file.

src/basekernels/wiener.jl
devmotion: I was wondering if one should keep the common part (i.e., the first two or three lines) together but actually as you say that would require inner functions with many arguments. So maybe it would be cleaner to just dispatch the `kernelmatrix` call on `I` (similar to the methods for individual samples).

willtebbutt:
This looks to be stale, so I'm closing.

PR #169, opened 2020-09-18
Add @boundscheck for MOInput and  remove explicit definitions of length and iterate

sharanry:
Removing unneeded definitions and using `@boundscheck` macro based on discussion in https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/pull/30


src/mokernels/moinput.jl
```diff
@@ -8,24 +8,15 @@ struct MOInput{T<:AbstractVector} <: AbstractVector{Tuple{Any,Int}}
...
-Base.firstindex(inp::MOInput) = 1
-
 function Base.getindex(inp::MOInput, ind::Integer)
-    if ind > 0
-        out_dim = ind ÷ length(inp.x) + 1
-        ind = ind % length(inp.x)
-        if ind==0 ind = length(inp.x); out_dim-=1 end
-        return (inp.x[ind], out_dim::Int)
-    else
+    @boundscheck if ind <= 0 || ind > length(inp)
```
devmotion, line 14: As discussed in https://docs.julialang.org/en/v1/devdocs/boundscheck/, this should just be
```suggestion
    @boundscheck checkbounds(inp, ind)
```
and the function has to be inlined for `@inbounds` to have any effect.

devmotion: In particular the custom error in the next lines should be removed.

src/mokernels/moinput.jl
```diff
@@ -8,24 +8,12 @@ struct MOInput{T<:AbstractVector} <: AbstractVector{Tuple{Any,Int}}
...
 
-Base.length(inp::MOInput) = inp.out_dim * length(inp.x)
-
-Base.size(inp::MOInput, d) = d::Integer == 1 ? inp.out_dim * size(inp.x, 1) : 1 
 Base.size(inp::MOInput) = (inp.out_dim * size(inp.x, 1),)
 
-Base.lastindex(inp::MOInput) = length(inp)
-Base.firstindex(inp::MOInput) = 1
-
 function Base.getindex(inp::MOInput, ind::Integer)
```
devmotion, line 13:
```suggestion
@inline function Base.getindex(inp::MOInput, ind::Integer)
```

src/mokernels/moinput.jl
```diff
@@ -8,23 +8,12 @@ struct MOInput{T<:AbstractVector} <: AbstractVector{Tuple{Any,Int}}
...
-        out_dim = ind ÷ length(inp.x) + 1
-        ind = ind % length(inp.x)
-        if ind==0 ind = length(inp.x); out_dim-=1 end
-        return (inp.x[ind], out_dim::Int)
-    else
-        throw(BoundsError(string("Trying to access at ", ind)))
-    end
+    @boundscheck checkbounds(inp, ind)
+    out_dim, ind = fldmod1(ind, length(inp.x))
+    return (inp.x[ind], out_dim::Int)
```
devmotion, line 16:
```suggestion
    return inp.x[ind], out_dim
```

src/mokernels/moinput.jl
devmotion: This is required to force inlining of the function inside of `@inbounds`, according to the [docs](https://docs.julialang.org/en/v1/devdocs/boundscheck/#Eliding-bounds-checks) (and therefore the shown example also contains an `@inline` annotation):

> The @boundscheck(...) macro marks blocks of code that perform bounds checking. When such blocks are inlined into an @inbounds(...) block, the compiler may remove these blocks.

src/mokernels/moinput.jl
```diff
@@ -8,23 +8,12 @@ struct MOInput{T<:AbstractVector} <: AbstractVector{Tuple{Any,Int}}
     out_dim::Integer
 end
 
-Base.length(inp::MOInput) = inp.out_dim * length(inp.x)
-
-Base.size(inp::MOInput, d) = d::Integer == 1 ? inp.out_dim * size(inp.x, 1) : 1 
 Base.size(inp::MOInput) = (inp.out_dim * size(inp.x, 1),)
```
devmotion, line 11: To be consistent with the definition of the indexing operation, this should probably be
```suggestion
Base.size(inp::MOInput) = (inp.out_dim * length(inp.x),)
```

devmotion (changes requested):
IMO `MOInput` requires more changes, similar to those discussed for `MOOutput`, but maybe they should go into another PR (e.g., the parent type should be changed to `AbstractVector{Tuple{T,Int}}` which defines a concrete element type and probably `iterate` should be removed).

In any case, it would be good to address https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/169/files#r494395330 before merging this PR.

src/mokernels/moinput.jl
```diff
@@ -8,23 +8,12 @@ struct MOInput{T<:AbstractVector} <: AbstractVector{Tuple{Any,Int}}
...
-    else
-        throw(BoundsError(string("Trying to access at ", ind)))
-    end
+@inline function Base.getindex(inp::MOInput, ind::Integer)
+    @boundscheck checkbounds(inp, ind)
+    out_dim, ind = fldmod1(ind, length(inp.x))
+    return inp.x[ind], out_dim
 end
 
 Base.iterate(inp::MOInput) = (inp[1], 1)
```
devmotion, line 19: Just noticed that the title of the PR explicitly says "...remove explicit definitions of length and iterate" but `iterate` is still implemented. Is there a reason for it?

sharanry: I had to add it back because the default iterate for `AbstractVector{Tuple{Any,Int}}` is not what we want. It produces a couple of tests fail.

```julia
  Expression: iterate(mgpi) == (mgpi[1], 1)
   Evaluated: (([0.008457582224595583, 0.41464287382643517, 0.40664694052473127, 0.9330052885761637, 0.49389515633005887], 1), (Base.OneTo(12), 1)) == (([0.008457582224595583, 0.41464287382643517, 0.40664694052473127, 0.9330052885761637, 0.49389515633005887], 1), 1)

  Expression: iterate(mgpi, 2) == (mgpi[3], 3)
   Evaluated: (([0.7413131597666685, 0.5768240627991614, 0.6903596228103408, 0.18912000772418436, 0.8404194534198497], 1), (2, nothing)) == (([0.3674929338482975, 0.31972403013388906, 0.2344192068115616, 0.9842115504995159, 0.44655713080439274], 1), 3)

```

devmotion: I think we should just remove the implementation and adjust the tests :slightly_smiling_face: The values are correct and the states are due to the standard implementation in base, which works for linear and cartesian indexing (and according to the comments is fast for linear indexing): https://github.com/JuliaLang/julia/blob/4165fddaeb4c85c6a16ccc2fc47b33376081ed31/base/abstractarray.jl#L1043-L1047

test/mokernels/moinput.jl
```diff
@@ -1,16 +1,16 @@
...
     @test length(mgpi) == 12
     @test size(mgpi) == (12,)
     @test size(mgpi, 1) == 12
     @test size(mgpi, 2) == 1
     @test lastindex(mgpi) == 12
     @test firstindex(mgpi) == 1
-    @test iterate(mgpi) == (mgpi[1], 1)
-    @test iterate(mgpi, 2) == (mgpi[3], 3)
+    @test iterate(mgpi) == (mgpi[1], (Base.OneTo(12), 1))
+    @test iterate(mgpi, 2) == (mgpi[2], (2, nothing))
```
devmotion, line 13: I don't think we should test internals that are implemented in Julia base. Maybe more reasonable would be something like
```
@test [xi for xi in x] == ...
```

sharanry:
@devmotion do you think it is okay to merge this or do you suggest any other changes?

devmotion:
Looks good mainly, just two questions:
- What is the motivation for introducing `moinput`? It seems in contrast to `mooutput` there is no alternative implemented that doesn't return a `MOInput` instance?
- The difference between `MOInput` and `MOOutput` is mainly that the first one's elements are tuples where the second element corresponds to the output dimension, right?

sharanry:
>  What is the motivation for introducing `moinput`? It seems in contrast to `mooutput` there is no alternative implemented that doesn't return a `MOInput` instance?

I wanted it to be similar to `moutput`. Mixed use of capitals just seemed very non-intuitive to me. I want to avoid usage like 
```julia
X, Y = MOInput(ColVecs(x), out_dim), moutput(y)
```

> The difference between MOInput and MOOutput is mainly that the first one's elements are tuples where the second element corresponds to the output dimension, right?

This is one aspect of it. Additionally, one can think of `MOInput` to emulate repetition of inputs for each output dimension. For example N inputs are viewed through `MOInput` as `out_dim x N` inputs without allocating any additional memory. Whereas, `MOutput` emulates flattening of the outputs from size `(out_dim, N)` to vector of length `out_dim x N`.

devmotion:
> I wanted it to be similar to moutput. Mixed use of capitals just seemed very non-intuitive to me.

Seems reasonable. In this case I suggest to not export `MOInput` anymore though.

src/mokernels/moinput.jl
```diff
@@ -1,31 +1,26 @@
...
-    x::T
-    out_dim::Integer
+struct MOInput{T,X} <: AbstractVector{Tuple{T,Int}}
+    x::X
+    out_dim::Int
 end
 
-Base.length(inp::MOInput) = inp.out_dim * length(inp.x)
+"""
+    moinput(x::AbstractVector, out_dim::Int)
```
willtebbutt, line 12: I think this should probably be underscore-separated: `mo_input`.

sharanry:
@devmotion @willtebbutt Is this okay to merge?

src/KernelFunctions.jl
```diff
@@ -46,7 +46,7 @@ export spectral_mixture_kernel, spectral_mixture_product_kernel
 
 export ColVecs, RowVecs
 
-export MOInput
+export MOInput, mo_input
```
devmotion, line 49: I'd suggest to remove `MOInput` from the exports:
```suggestion
export mo_input
```

src/mokernels/mo_input.jl
```diff
@@ -0,0 +1,26 @@
+"""
+    MOInput
+
+A data type to accomodate modelling multi-dimensional output data.
```
devmotion, line 4:
```suggestion
A data type to accomodate modelling multi-dimensional input data.
```

src/mokernels/mo_input.jl
```diff
@@ -0,0 +1,26 @@
...
+"""
+struct MOInput{T,X} <: AbstractVector{Tuple{T,Int}}
+    x::X
+    out_dim::Int
+end
+
+"""
+    mo_input(x::AbstractVector, out_dim::Int)
+
+Return `MOInput` to accomodate modelling multi-dimensional output data.
```
devmotion, line 14: IMO this docstring is not very helpful, since also the one of `MOInput` is not more informative. At some point we should clarify what these methods/types are actually doing.

willtebbutt:
Stale, so closing. Please re-open if updates are provided.

Issue #172, opened 2020-09-22
PeriodicKernel parameters pehaps a little un-intuitive


willtebbutt:
The parameter `r` in our `PeriodicKernel` currently governs the size of the circle onto which the inputs are mapped. Would it not be more intuitive to parametrise it in terms of the frequency or period?

This might be a good opportunity to consider whether we actually _need_ any parameters in the `PeriodicKernel`, since we can get everything that we need via `transform`, although we maybe want to ensure that that's wrapped for users.

PR #173, opened 2020-09-22
Implement periodic transform

willtebbutt:
Allows us to periodicise kernels other than the EQ.

Note that the exact translation from this transformation (which does the canonical thing) and the periodic EQ is a bit convoluted. See #172 for discussion.

### DIFF
--- a/src/transform/periodic_transform.jl
+++ b/src/transform/periodic_transform.jl
@@ -1,7 +1,8 @@
 """
-    PeriodicTransform{Tf<:Ref{<:Real}} <: Transform
+    PeriodicTransform(f)
 
-Makes a kernel periodic by transforming an input to be periodic.
+Makes a kernel periodic by mapping a scalar input onto the unit circle. Samples from a GP
+with a kernel with this transformation applied will produce samples with frequenct `f`.
 """
 struct PeriodicTransform{Tf<:Ref{<:Real}} <: Transform
     f::Tf
@@ -13,7 +14,7 @@ PeriodicTransform(f::Real) = PeriodicTransform(Ref(f))
 
 dim(t::PeriodicTransform) = 2
 
-(t::PeriodicTransform)(x::Real) = [cosp((2 * t.f[]) * x), sinpi((2 * t.f[]) * x)]
+(t::PeriodicTransform)(x::Real) = [cospi((2 * t.f[]) * x), sinpi((2 * t.f[]) * x)]
 
 function _map(t::PeriodicTransform, x::AbstractVector{<:Real})
     return RowVecs(hcat(cospi.((2 * t.f[]) .* x), sinpi.((2 * t.f[]) .* x)))
--- a/src/transform/periodic_transform.jl
+++ b/src/transform/periodic_transform.jl
@@ -4,24 +4,26 @@
 Makes a kernel periodic by mapping a scalar input onto the unit circle. Samples from a GP
 with a kernel with this transformation applied will produce samples with frequenct `f`.
 """
-struct PeriodicTransform{Tf<:Ref{<:Real}} <: Transform
+struct PeriodicTransform{Tf<:AbstractVector{<:Real}} <: Transform
     f::Tf
 end
 
 @functor PeriodicTransform
 
-PeriodicTransform(f::Real) = PeriodicTransform(Ref(f))
+PeriodicTransform(f::Real) = PeriodicTransform([f])
 
 dim(t::PeriodicTransform) = 2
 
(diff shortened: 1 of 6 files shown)


willtebbutt:
@molet this is ready for a review if you're up for it.

src/transform/periodic_transform.jl
```diff
@@ -0,0 +1,27 @@
...
+    f::Tf
+end
+
+@functor PeriodicTransform
+
+PeriodicTransform(f::Real) = PeriodicTransform(Ref(f))
+
+dim(t::PeriodicTransform) = 2
+
+(t::PeriodicTransform)(x::Real) = [cospi((2 * t.f[]) * x), sinpi((2 * t.f[]) * x)]
```
devmotion, line 17: On Julia >= 1.6, we should probably use `sincospi` instead. Maybe change the order here to avoid breaking changes?

src/transform/periodic_transform.jl
```diff
@@ -0,0 +1,27 @@
...
+
+function _map(t::PeriodicTransform, x::AbstractVector{<:Real})
+    return RowVecs(hcat(cospi.((2 * t.f[]) .* x), sinpi.((2 * t.f[]) .* x)))
+end
+
+Base.isequal(t1::PeriodicTransform, t2::PeriodicTransform) = isequal(t1.f[], t2.f[])
```
devmotion, line 23: Is this needed? I'm always confused about whether to define `==` or `isequal` - according to the docstring
```julia
  Implementation
  ≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡

  The default implementation of isequal calls ==, so a type that does not involve
  floating-point values generally only needs to define ==.
```
and hence we implement `isequal` since `f[]` probably is a floating point number?

src/transform/periodic_transform.jl
```diff
@@ -0,0 +1,27 @@
...
+Makes a kernel periodic by mapping a scalar input onto the unit circle. Samples from a GP
+with a kernel with this transformation applied will produce samples with frequenct `f`.
+"""
+struct PeriodicTransform{Tf<:Ref{<:Real}} <: Transform
+    f::Tf
+end
+
+@functor PeriodicTransform
+
+PeriodicTransform(f::Real) = PeriodicTransform(Ref(f))
```
theogf, line 13: f should be a Vector and not a Ref to be compatible with Functors, Zygote etc (I think)

src/transform/periodic_transform.jl
```diff
@@ -0,0 +1,27 @@
+"""
+    PeriodicTransform(f)
+
+Makes a kernel periodic by mapping a scalar input onto the unit circle. Samples from a GP
+with a kernel with this transformation applied will produce samples with frequenct `f`.
```
molet, line 5:
```suggestion
with a kernel with this transformation applied will produce samples with frequency `f`.
```

src/transform/periodic_transform.jl
```diff
@@ -0,0 +1,27 @@
...
+"""
+struct PeriodicTransform{Tf<:Ref{<:Real}} <: Transform
+    f::Tf
```
molet, line 8: I suggest to check if the specified frequency is positive.

theogf: We actually don't have any checks of this type for any Transform and we leave this responsibility to the user. 
If we change this policy I would be for having a general PR solving this problem

willtebbutt: Additionally, does the frequency even need to be positive? Sure, you might _want_ to constrain it to be positive, but I don't think that a negative frequency causes any conceptual or practical problems as far as this library is concerned.

src/transform/periodic_transform.jl
willtebbutt: I'm really not sure -- I'm just mirroring what's already there. I'll leave as-is unless you object.

devmotion: It's fine, I was just curious.

src/transform/periodic_transform.jl
molet: Indeed, conceptually any real frequency would work (well, even zero frequency) and also if it is the user's responsibility to make sure the parameters are reasonable then I'm fine with it.

willtebbutt:
AFAICT everyone is now happy with this PR. Subject to approval, I can go ahead and merge -- testing will improve once #159 is available.

molet (approved):

devmotion (approved):

devmotion:
Bump the version number I guess?

PR #181, opened 2020-10-01
SelectTransform with single index

willtebbutt:
The output `SelectTransform(5)(some_vector)` should really be a scalar, however, this currently doesn't work. It is important that this works so that (for example) a single dimension of a vector-valued input can be passed to kernels that can't handle vector-valued inputs, such as kernels constructed via `PeriodicTransform`.

This PR addresses this issue by making this possible.

### DIFF
--- a/test/transform/selecttransform.jl
+++ b/test/transform/selecttransform.jl
@@ -1,114 +1,114 @@
 @testset "selecttransform" begin
-    # rng = MersenneTwister(123456)
+    rng = MersenneTwister(123456)
 
-    # select = [1, 3, 5]
-    # t = SelectTransform(select)
+    select = [1, 3, 5]
+    t = SelectTransform(select)
 
-    # x_vecs = [randn(rng, maximum(select)) for _ in 1:3]
-    # x_cols = ColVecs(randn(rng, maximum(select), 6))
-    # x_rows = RowVecs(randn(rng, 4, maximum(select)))
+    x_vecs = [randn(rng, maximum(select)) for _ in 1:3]
+    x_cols = ColVecs(randn(rng, maximum(select), 6))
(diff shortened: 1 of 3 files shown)


test/transform/selecttransform.jl
```diff
@@ -109,4 +109,14 @@
             testfunction(ta_col, a, C, 1)
         end
     end
+
+    @testset "single-index" begin
```
theogf, line 113: Could you also check what happens to a Vector of Vectors?

molet (approved):
LGTM

test/transform/selecttransform.jl
```diff
@@ -109,4 +109,15 @@
...
+        t = SelectTransform(4)
+        @testset "$(name)" for (name, x) in [
+            ("Vector{<:Vector}", [randn(6) for _ in 1:3]),
+            ("ColVecs", ColVecs(randn(5, 10))),
+            ("RowVecs", RowVecs(randn(11, 4))),
+        ]
+            @test KernelFunctions._map(t, x) isa AbstractVector{Float64}
```
theogf, line 120: I think you need to relax the test as `AbstractVector{<:Real}`

willtebbutt: Ah. I've just not changed the evaluation definition. Will do that now.

willtebbutt: Hmmm `view` is inconsistent with `getindex` for integer-indexing. `getindex` has the behaviour that we want. Need to find a work-around. Shouldn't be too hard.

willtebbutt:
@theogf anything else you need?

theogf:
Nope it's all good!

theogf (approved):

PR #182, opened 2020-10-04
Improve testing

willtebbutt:
This PR makes use of a lot more data points during testing, which makes the tests somewhat harder to pass.

However, it also relaxes some tolerances for specific kernels that appear to struggle for accuracy. For example, the exponential kernel appears to be fairly inaccurate, so the tolerance has been relaxed a bit there.

The extra data points highlighted that the `Gabor` kernel cannot be used in conjunction with multi-dimensional inputs, which is something that we were previously doing in our tests. That has now been resolved.

@sharanry if possible, could you see if this fixes your problem, or are you only seeing test failures in CI?


willtebbutt:
These failures are weird, and I'm really struggling to reproduce them locally. Not really sure what's going on :grimace:

willtebbutt:
Tests now pass. This is ready for review. CI drop is related to test utils.

devmotion (approved):

PR #185, opened 2020-10-05
Update docstring corresponding to #158

johannesgiersdorf:
Edited the doc sting according to the new Update 0.8.0


src/basekernels/exponential.jl
```diff
@@ -3,7 +3,7 @@
 
 The squared exponential kernel is a Mercer kernel given by the formula:
 ```
-    κ(x,y) = exp(-‖x-y‖²)
+    κ(x,y) = exp(-‖x-y‖²/2)
```
willtebbutt, line 6:
```suggestion
    κ(x, y) = exp(-‖x - y‖² / 2)
```

devmotion (approved):

willtebbutt (approved):
LGTM modulo style fix (that I figure we might as well do while we're here).

Also, could you bump the patch version so that we can release 0.8.1 with this fixed docstring?

src/basekernels/exponential.jl
```diff
@@ -3,7 +3,7 @@
 
 The squared exponential kernel is a Mercer kernel given by the formula:
 ```
-    κ(x,y) = exp(-‖x-y‖²)
+    κ(x, y) = exp(-‖x - y‖² / 2)
 ```
 Can also be called via `SEKernel`, `GaussianKernel` or `SEKernel`.
```
devmotion, line 8: While you're at it: This should actually be changed to
```suggestion
Can also be called via `RBFKernel`, `GaussianKernel` or `SEKernel`.
```

devmotion (approved):

PR #207, opened 2020-12-09
Link to stable documentation

devmotion:
I noticed that currently http://juliagaussianprocesses.github.io/KernelFunctions.jl/ links to the unstable docs but I assume it might be better to link to the stable version.


theogf:
Won't that be recreated after a new build?

devmotion:
The `index.html` file? I don't think so, at least in AbstractGPs the file wasn't updated or changed in a while: https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/tree/gh-pages

theogf (approved):

Issue #203, opened 2020-12-07
Slow mode for kerneldiagmatrix


theogf:
Zygote currently fails to differentiate through `kerneldiagmatrix` when given a `RowVecs` or a `ColVecs`

MWE : 
```julia
using KernelFunctions, Zygote

X = KernelFunctions.RowVecs(rand(3, 3))
k = transform(SqExponentialKernel(), 2.0)
Zygote.gradient(k) do k
	sum(kerneldiagmatrix(k, X))
end

ERROR: In slow method
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] (::KernelFunctions.var"#back#186")(::Array{Array{Float64,1},1}) at /home/theo/.julia/packages/KernelFunctions/V02nz/src/zygote_adjoints.jl:75
 [3] (::KernelFunctions.var"#171#back#187"{KernelFunctions.var"#back#186"})(::Array{Array{Float64,1},1}) at /home/theo/.julia/packages/ZygoteRules/6nssF/src/adjoint.jl:49
 [4] _map at /home/theo/.julia/packages/KernelFunctions/V02nz/src/transform/scaletransform.jl:26 [inlined]
 [5] (::typeof(∂(_map)))(::Array{Array{Float64,1},1}) at /home/theo/.julia/packages/Zygote/nK6sg/src/compiler/interface2.jl:0
 [6] kerneldiagmatrix at /home/theo/.julia/packages/KernelFunctions/V02nz/src/kernels/transformedkernel.jl:85 [inlined]
 [7] (::typeof(∂(kerneldiagmatrix)))(::FillArrays.Fill{Float64,1,Tuple{Base.OneTo{Int64}}}) at /home/theo/.julia/packages/Zygote/nK6sg/src/compiler/interface2.jl:0

```

willtebbutt:
Right. So what's happening is on the reverse-pass, the gradient w.r.t. the output of
```julia
_map(κ.transform, x)
```
[...]

PR #208, opened 2020-12-09
Fix gradient issues with kernelmatrix_diag and use ChainRulesCore

theogf:
This PR aims at solving the AD issues coming from `kernelmatrix_diag`, add tests for it and incidentally solve AD issue for when `x==y`.
This solves the issue on #203


willtebbutt:
This makes me a little nervous. The benefit of calling `kerneldiagmatrix` from inside `kerneldiagmatrix`, is that if there's a specialised method in the base kernel for `kerneldiagmatrix` (e.g. return all zeros), that still gets called.

I would have thought that a better solution would be to require `_map(κ.transform, x)` to be defined for each transform, so that it happens efficiently.

willtebbutt:
I'm actually surprised that this problem is happening at all. Looking into it now.

test/utils_AD.jl
```diff
@@ -0,0 +1,180 @@
+
+const FDM = FiniteDifferences.central_fdm(5, 1)
```
devmotion, line 2: This file seems to contain a lot of code that already exists in `test_utils.jl`?

theogf: That's possible there is some overlapping. @willtebbutt wrote the Test_Utils module, could you comment on it?

theogf:
@willtebbutt It seems your fix on defining methods for SimpleKernel fixed the issue. I added AD test for kerneldiagmatrix now

test/utils_AD.jl
devmotion: It's basically exactly the same apart from the `testdiagfunction`, isn't it? https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/master/test/test_utils.jl

theogf: Yeah and in `test/test_utils.jl` there is more AD stuff.
I think it needs a different PR to solve this though

devmotion: Hmm I don't understand - can't you just add the `testdiagfunction` parts to `test/test_utils.jl`? The `test/utils_AD.jl` file in this PR does not exist and seems to duplicate quite many things.

devmotion: You also did not include or load the new `test/utils_AD.jl` file in the tests.

theogf: Oh you're right I got confused. I did not see that utils_AD.jl was not used...

test/test_utils.jl
```diff
@@ -105,6 +107,21 @@ function test_FiniteDiff(kernelfunction, args = nothing, dims = [3, 3])
...
+            @test_nowarn gradient(:FiniteDiff, A) do a
+                testdiagfunction(k, a, dim)
+            end
+            @test_nowarn gradient(:FiniteDiff , A) do a
+                testdiagfunction(k, a, B, dim)
+            end
+            @test_nowarn gradient(:FiniteDiff, B) do b
+                testdiagfunction(k, A, b, dim)
+            end
+            if !(args === nothing)
```
devmotion, line 120: It doesn't really matter but it feels a bit more natural to write:
```suggestion
            if args !== nothing
```

test/test_utils.jl
```diff
@@ -157,6 +174,21 @@ function test_AD(AD::Symbol, kernelfunction, args = nothing, dims = [3, 3])
...
+            compare_gradient(AD, A) do a
+                testdiagfunction(k, a, dim)
+            end
+            compare_gradient(AD, A) do a
+                testdiagfunction(k, a, B, dim)
+            end
+            compare_gradient(AD, B) do b
+                testdiagfunction(k, A, b, dim)
+            end
+            if !(args === nothing)
```
devmotion, line 187:
```suggestion
            if args !== nothing
```

devmotion:
It seems the error is not fixed?

theogf:
> It seems the error is not fixed?

Yep, I tried @willtebbutt suggestion but it does not solve the problem...

src/matrix/kernelmatrix.jl
```diff
@@ -103,6 +103,14 @@ function kernelmatrix(κ::SimpleKernel, x::AbstractVector, y::AbstractVector)
     return map(d -> kappa(κ, d), pairwise(metric(κ), x, y))
 end
 
+function kerneldiagmatrix(κ::SimpleKernel, x::AbstractVector)
+    return map(x -> κ(x, x), x)
```
willtebbutt, line 107:
```suggestion
    return kerneldiagmatrix(κ, x, x)
```

src/matrix/kernelmatrix.jl
```diff
@@ -103,6 +103,14 @@ function kernelmatrix(κ::SimpleKernel, x::AbstractVector, y::AbstractVector)
...
+end
+
+function kerneldiagmatrix(κ::SimpleKernel, x::AbstractVector, y::AbstractVector)
+    return map(d -> kappa(κ, d), map(metric(κ), x, y))
```
willtebbutt, line 111:
```suggestion
    return map(d -> kappa(κ, d), colwise(metric(κ), x, y))
```
The solution isn't currently working because `map` utilises `getindex` on `x` and `y` to compute the distances. You would need to use an implementation that doesn't do this -- `Distances.colwise` would satisfy this.

theogf: Oh I see, where `Distances.colwise` is not defined for vectors right now, but we could have our own implementation.

willtebbutt: Yeah, we'll probably want to do what we did with `pairwise` and roll our own to avoid type-pirating as you say.

theogf: This still does not work, or maybe my implementation of `colwise` is not what you thought about

src/distances/pairwise.jl
```diff
@@ -39,3 +39,15 @@ function pairwise!(
...
+# Also defines the colwise method for abstractvectors
+
+
+function colwise(::PreMetric, x::AbstractVector)
+    zeros(length(x))
+end
+
+function colwise(d::PreMetric, x::AbstractVector, y::AbstractVector)
+    broadcast(d, x, y)
+end
```
willtebbutt, line 53: Yup, you'll need to define `colwise` specifically for `RowVecs` and `ColVecs`

theogf: `Distances.colwise` is mutating arrays :cry:

theogf:
Back to this annoying problem!
So I seem to have found a solution (for Zygote at least), but it has a small touch of piracy :D.
`Distances.jl` accepts now iterators/generator for `pairwise` which makes our task easier. However it's not the case for `colwise` which only takes matrices... I dispatched correctly for `ColVecs` and `RowVecs` using `x.X` and `x.X'` but then for the generic case of `AbstractVector` I had to implement our own version -> type piracy... I will anyway open an issue on Distances.jl to see if it is possible to add the option.

devmotion:
I guess the correct way would be to define our own `colwise` function, analogously to `pairwise`. And then fall back to the definitions in Distances where it is possible.

src/distances/pairwise.jl
```diff
@@ -29,3 +29,31 @@ function pairwise!(
...
+function colwise(d::PreMetric, x::ColVecs)
+    Distances.colwise(d, x.X, x.X)
+end
+
+function colwise(d::PreMetric, x::RowVecs)
+    Distances.colwise(d, x.X', x.X')
+end
+
+function colwise(d::PreMetric, x::AbstractVector)
+    d.(x, x)
```
devmotion, line 46: I think
```suggestion
    return map(d, x, x)
```
would be simpler here - there's no need for the broadcasting machinery since the inputs have the same dimension.

src/matrix/kernelmatrix.jl
```diff
@@ -100,6 +100,16 @@ function kernelmatrix(κ::SimpleKernel, x::AbstractVector, y::AbstractVector)
     return map(d -> kappa(κ, d), pairwise(metric(κ), x, y))
 end
 
+function kernelmatrix_diag(κ::SimpleKernel, x::AbstractVector)
+    return map(d -> kappa(κ, d), colwise(metric(κ), x))
```
devmotion, line 104: BTW one could use `Base.Fix1` here instead of the custom closure:
```suggestion
    return map(Base.Fix1(kappa, κ), colwise(metric(κ), x))
```

src/zygote_adjoints.jl
```diff
@@ -23,6 +23,11 @@ end
     end
 end
 
+@adjoint function Distances.colwise(d::Delta, X::AbstractMatrix, Y::AbstractMatrix)
+    return Distances.colwise(d, X, Y), function (Δ::AbstractVector)
+        return (nothing, nothing, nothing)
+    end
+end
```
devmotion, line 30: Could we add ChainRules-definitions instead of rules for Zygote? ChainRules also avoids the Zygote-weirdness with `nothing`.

src/matrix/kernelmatrix.jl
theogf: When was `Fix1` introduced? I thought it was a recent addition

src/zygote_adjoints.jl
theogf: Completely agree with this, but should it be in this PR or should it be done in another one?

src/matrix/kernelmatrix.jl
devmotion: It exists since Julia 0.7.0-alpha: https://github.com/JuliaLang/julia/pull/26708

theogf: Haha while watching the Julia PR I saw this PR reference appearing :p

src/zygote_adjoints.jl
theogf: I ended up rewriting `rrule` wherever it was possible

theogf:
Oh man this starts to be exhausting :tired_face: 

So now moving to `rrule` should solve most of the issues, **EXCEPT** that as pointed by @willtebbutt in https://github.com/JuliaDiff/ChainRulesCore.jl/issues/239#issuecomment-713560336, since `Zygote` defines already some `@adjoint` for `Distances.pairwise` and therefore the `rrule` we define for our own types are completely ignored...

@willtebbutt should I make a PR to Zygote to pass the `Distances.jl` rule under the `ChainRules` syntax?

willtebbutt:
Eurgh, yeah, this _is_ turning into a pain. Thanks for persevering.

I would be pro- that and would be willing to both review a PR and push for it to be merged.

theogf:
My bad, removing adjoints from Zygote does not help. @willtebbutt Could you have a quick look at `chainrules.jl` (previously `zygote_adjoints.jl`) ? It seems my `rrule` are completely ignored by Zygote...

theogf:
Yep I mostly restart the REPL everytime. I am evaluating `test/basekernels/constant.jl` because it needs the `Delta` `rrule`

theogf:
OMG I am so stupid, I found the problem...
Since I was doing `using ChainRulesCore`, `rrule` was not overloaded but just created as a new local function...

willtebbutt:
Classic -- happens to me all of the time!

src/chainrules.jl
```diff
@@ -0,0 +1,142 @@
...
+
+function rrule(
+    ::typeof(Distances.pairwise),
+    d::DotProduct,
+    X::AbstractMatrix,
+    Y::AbstractMatrix;
+    dims=2,
+)
+    P = Distances.pairwise(d, X, Y; dims=dims)
+    if dims == 1
```
devmotion, line 56: Does this check lead to any type inference problems?

src/chainrules.jl
```diff
@@ -0,0 +1,142 @@
...
+        return P, pairwise_pullback_rows
+    end
+end
+
+function rrule(
+    ::typeof(Distances.colwise), d::DotProduct, X::AbstractMatrix, Y::AbstractMatrix
+)
+    C = Distances.colwise(d, X, Y)
+    function colwise_pullback(Δ::AbstractVector)
+        return (nothing, Δ' .* Y, Δ' .* X)
```
devmotion, line 89:
```suggestion
        return NO_FIELDS, Δ' .* Y, Δ' .* X
```

src/chainrules.jl
```diff
@@ -0,0 +1,142 @@
...
+end
+
+## Reverse Rules Sinus
+function rrule(::typeof(Distances.evaluate), s::Sinus, x::AbstractVector, y::AbstractVector)
+    d = (x - y)
+    sind = sinpi.(d)
+    val = sum(abs2, sind ./ s.r)
+    gradx = 2π .* cospi.(d) .* sind ./ (s.r .^ 2)
+    function evaluate_pullback(Δ)
+        return (r=-2Δ .* abs2.(sind) ./ s.r,), Δ * gradx, -Δ * gradx
```
devmotion, line 101: It seems `-2 .* abs2.(sind) ./ s.r` could be precomputed? I'm not completely familiar with the latest updates in ChainRules but AFAIK one should use `Composite` here?

src/chainrules.jl
```diff
@@ -0,0 +1,142 @@
...
+    end
+    return C, colwise_pullback
+end
+
+## Reverse Rules Sinus
+function rrule(::typeof(Distances.evaluate), s::Sinus, x::AbstractVector, y::AbstractVector)
+    d = (x - y)
+    sind = sinpi.(d)
+    val = sum(abs2, sind ./ s.r)
+    gradx = 2π .* cospi.(d) .* sind ./ (s.r .^ 2)
```
devmotion, line 99: Do we depend on StatsFuns? If yes, one could also use
```suggestion
    gradx = twoπ .* cospi.(d) .* sind ./ (s.r .^ 2)
```

src/chainrules.jl
```diff
@@ -0,0 +1,142 @@
...
+    return val, evaluate_pullback
+end
+
+## Reverse Rules for matrix wrappers
+
+function rrule(::ColVecs, X::AbstractMatrix)
+    ColVecs_pullback(Δ::NamedTuple) = (Δ.X,)
+    ColVecs_pullback(Δ::AbstractMatrix) = (Δ,)
+    function ColVecs_pullback(Δ::AbstractVector{<:AbstractVector{<:Real}})
+        return throw(error("In slow method"))
```
devmotion, line 112: What's the advantage over just
```suggestion
        return error("in slow method")
```
?

src/chainrules.jl
```diff
@@ -0,0 +1,142 @@
...
+    C = Distances.colwise(d, X, Y)
+    function colwise_pullback(Δ::AbstractVector)
+        return (nothing, Δ' .* Y, Δ' .* X)
+    end
+    return C, colwise_pullback
+end
+
+## Reverse Rules Sinus
+function rrule(::typeof(Distances.evaluate), s::Sinus, x::AbstractVector, y::AbstractVector)
+    d = (x - y)
```
devmotion, line 96:
```suggestion
    d = x - y
```

src/chainrules.jl
theogf: Do you mean the `if dims == 1` ?

devmotion: Yes. Mainly because you return two different functions depending on the value of `dim`, so I wonder if it messes up type inference and if it would be better to move the check inside the pullback function.

src/chainrules.jl
theogf: That's legacy code from Stheno.jl so I don't know but I guess you're right.

theogf:
Ok things look a bit better :)
@willtebbut Can you have a look at the `rrule` for the `Row/ColVecs` constructors? I had to define a new pullback for `Composite` types and it looked a bit different. Tests seem to like it but I am not sure it's entirely correct.

src/chainrules.jl
```diff
@@ -0,0 +1,158 @@
...
+
+function ChainRulesCore.frule((_, Δx, Δy), d::Distances.Euclidean, x, y)
+    Δ = x - y
+    D = sqrt(sum(abs2, Δ))
+    if iszero(D)
+        return D, Zero()
+    else
+        Δ ./= D
+        return D, dot(Δ, Δx) - dot(Δ, Δy)
+    end
```
devmotion, line 11: This is definitely type unstable, isn't it? Maybe we could write instead:
```suggestion
    if !iszero(D)
        Δ ./= D
    end
    return D, dot(Δ, Δx) - dot(Δ, Δy)
```

src/chainrules.jl
```diff
@@ -0,0 +1,158 @@
...
+end
+
+## Reverse Rules Delta
+
+function ChainRulesCore.rrule(::typeof(Distances.evaluate), s::Delta, x::AbstractVector, y::AbstractVector)
```
devmotion, line 16: Didn't Distances.jl switch to callable structs and forwards `evaluate(dist, x, y)` to `dist(x, y)`? In this case it would be better to define adjoints for `dist(x, y)` instead of `evaluate` (IIRC I changed this in the implementation of Zygote a while ago to fix a bug).

src/chainrules.jl
theogf: Ah yeah that's a good idea. But anyway it's not used right now.
Both ForwardDiff and ReverseDiff return `NaN` for the `Euclidean` metric, which is correct but highly undesirable. I can fix it for ReverseDiff but any idea how I can do it for ForwardDiff?

src/chainrules.jl
```diff
@@ -0,0 +1,158 @@
...
+    function colwise_pullback(::AbstractVector)
+        return NO_FIELDS, Zero(), Zero()
+    end
+    return C, colwise_pullback
+end
+
+## Reverse Rules DotProduct
+function ChainRulesCore.rrule(
+    ::typeof(Distances.evaluate), s::DotProduct, x::AbstractVector, y::AbstractVector
+)
```
devmotion, line 52: Same here, maybe it would be better to define an adjoint for
```suggestion
function ChainRulesCore.rrule(
    s::DotProduct, x::AbstractVector, y::AbstractVector
)
```
(if we overload `DotProduct`)

src/chainrules.jl
theogf: That's because we use `evaluate` here : https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/ae78b7337eced196f8f90e656252cf12b287391b/src/generic.jl#L9

But this could be changed as well.

And this is also because `DotProduct` is not a metric so we would need to add all the same machinery once #194 is finished.

theogf: Ah yeah you're right, I will also change it in the rest of the code.

theogf:
:tada:  **ALL TESTS ARE PASSING LOCALLY!** :tada:

Ok so for the Euclidean problem I solved it by having:

`colwise(::PreMetric, x) = zeros(length(x))` because that's the definition

while defining a different one for `DotProduct` and `Delta` (which are not proper distance, this should be solved in #194

theogf:
@devmotion The reviewdog is not happy with the EOF... Is that anything that can be solved?

theogf:
image

devmotion:
Ah I see.

I just ran JuliaFormatter locally and it fails since the files do not end with a newline character (it updates the two files that reviewdog complains about). So indeed the format check fails correctly. The only problem seems to be that Github suggestions do not work properly for these suggestions with an additional newline character. Therefore I guess in this case unfortunately one has to fix the files manually, either locally (with or without JuliaFormatter) or directly on Github.

src/chainrules.jl
```diff
@@ -0,0 +1,170 @@
...
+            "Pullback on AbstractVector{<:AbstractVector}.\n" *
+            "This might happen if you try to use gradients on the generic `kernelmatrix` or `kernelmatrix_diag`.\n" *
+            "To solve this issue overload `kernelmatrix(_diag)` for your kernel for `ColVecs`",
+        )
+    end
+    return ColVecs(X), ColVecs_pullback
+end
+
+function ChainRulesCore.rrule(::Type{<:RowVecs}, X::AbstractMatrix)
+    RowVecs_pullback(Δ::Composite) = (NO_FIELDS, Δ.X)
```
devmotion, line 140:
```suggestion
    RowVecs_pullback(Δ::Composite{<:RowVecs}) = (NO_FIELDS, Δ.X)
```

theogf:
Current fails are just random, it's ready to review/merge

src/chainrules.jl
```diff
@@ -0,0 +1,170 @@
...
+    end
+    return val, evaluate_pullback
+end
+
+## Reverse Rules for matrix wrappers
+
+function ChainRulesCore.rrule(::Type{<:ColVecs}, X::AbstractMatrix)
+    ColVecs_pullback(Δ::Composite) = (NO_FIELDS, Δ.X)
+    ColVecs_pullback(Δ::NamedTuple) = (Δ.X,)
+    ColVecs_pullback(Δ::AbstractMatrix) = (Δ,)
```
devmotion, line 128: When do these occur? And maybe one could restrict it to `NamedTuple{(:X,)}`?

src/chainrules.jl
```diff
@@ -0,0 +1,170 @@
...
+            "To solve this issue overload `kernelmatrix(_diag)` for your kernel for `ColVecs`",
+        )
+    end
+    return ColVecs(X), ColVecs_pullback
+end
+
+function ChainRulesCore.rrule(::Type{<:RowVecs}, X::AbstractMatrix)
+    RowVecs_pullback(Δ::Composite) = (NO_FIELDS, Δ.X)
+    RowVecs_pullback(Δ::NamedTuple) = (Δ.X,)
+    RowVecs_pullback(Δ::AbstractMatrix) = (Δ,)
```
devmotion, line 142: Same question here :slightly_smiling_face:

src/chainrules.jl
```diff
@@ -0,0 +1,170 @@
...
+    return RowVecs(X), RowVecs_pullback
+end
+
+ZygoteRules.@adjoint function Base.map(t::Transform, X::ColVecs)
+    return ZygoteRules.pullback(_map, t, X)
+end
+
+ZygoteRules.@adjoint function Base.map(t::Transform, X::RowVecs)
+    return ZygoteRules.pullback(_map, t, X)
+end
```
devmotion, line 159: Maybe keep these in a separate `zygoterules.jl` file?

src/chainrules.jl
```diff
@@ -0,0 +1,170 @@
...
+
+function ChainRulesCore.rrule(dist::Distances.SqMahalanobis, a, b)
+    d = dist(a, b)
+    function SqMahalanobis_pullback(Δ::Real)
+        B_Bᵀ = dist.qmat + transpose(dist.qmat)
+        a_b = a - b
+        δa = (B_Bᵀ * a_b) * Δ
```
devmotion, line 166: Maybe it would make sense to use a `Thunk` for `δa` (and therefore implicitly `B_Bᵀ`)?

src/chainrules.jl
```diff
@@ -0,0 +1,170 @@
...
+        return (qmat=(a_b * a_b') * Δ,), δa, -δa
```
devmotion, line 167: Should this again be a `Composite` instead?

src/distances/pairwise.jl
```diff
@@ -29,3 +29,34 @@ function pairwise!(
 )
     return Distances.pairwise!(out, d, reshape(x, :, 1), reshape(y, :, 1); dims=1)
 end
+
+# Also defines the colwise method for abstractvectors
+
+function colwise(::PreMetric, x::AbstractVector)
+    return zeros(length(x)) # Valid since d(x,x) == 0 by definition
+end
```
devmotion, line 37: It's a bit unfortunate that the return type might be different from `colwise(metric, x, x)`.

test/chainrules.jl
```diff
@@ -0,0 +1,27 @@
+@testset "Chain Rules" begin
```
devmotion, line 1: We should use ChainRulesTestUtils for testing ChainRules definitions.

devmotion (changes requested):
Looks good overall! I had a few questions, and I think we should use ChainRulesTestUtils to properly test the ChainRules definitions.

src/chainrules.jl
theogf: But I don't think the `dims` argument passed to the pullback ?

src/chainrules.jl
theogf: I have no idea... There is no documentation about this...

test/chainrules.jl
theogf: I am trying to add tests using `test_rrule` and `test_frule` but I get tons of errors (I think because of the structs). I would leave it aside for now...

src/chainrules.jl
devmotion: I guess @willtebbutt might know this?

test/chainrules.jl
devmotion: If it's too much work (and I mean there are already many changes in this PR) I would be fine with addressing it in a separate PR. Just make sure to open an issue that reminds us about it :slightly_smiling_face:

src/chainrules.jl
theogf: Also it's legacy and I don't think we even have a kernel with the SqMahalanobis distance :laughing:

devmotion: IIRC there was another definition where I thought one could/should use Composite.

test/chainrules.jl
theogf: Yeah I would prefer to finally finish with this. Note that these rules are anyway tested through the kernels (so we are kinda safe). But of course we should add them in a later state

src/chainrules.jl
```diff
@@ -0,0 +1,160 @@
...
+
+function ChainRulesCore.rrule(
+    dist::Distances.SqMahalanobis, a::AbstractVector, b::AbstractVector
+)
+    d = dist(a, b)
+    function SqMahalanobis_pullback(Δ::Real)
+        B_Bᵀ = dist.qmat + transpose(dist.qmat)
+        a_b = a - b
+        δa = @thunk((B_Bᵀ * a_b) * Δ)
+        return (qmat=(a_b * a_b') * Δ,), δa, @thunk(-δa)
```
devmotion, line 131: I think there might not be any benefit from putting the gradient of `b` in a thunk since the construction of the closure might be too expensive. Moreover, it seems one can safely assume that `qmat` is symmetric since the definition of `pairwise!` in Distances uses this property (see https://github.com/JuliaStats/Distances.jl/issues/174). After reading https://juliadiff.org/ChainRulesCore.jl/dev/gradient_accumulation.html, my suggestion would be:
```suggestion
        a_b = a - b
        ∂qmat = InplaceableThunk(
            @thunk((a_b * a_b') * Δ),
            X̄ -> mul!(X̄, a_b, a_b', true, Δ),
        )
        ∂a = InplaceableThunk(
            @thunk((2 * Δ) * dist.qmat * a_b),
            X̄ -> mul!(X̄, dist.qmat, a_b, true, 2 * Δ),
        )
        ∂b = InplaceableThunk(
            @thunk((-2 * Δ) * dist.qmat * a_b),
            X̄ -> mul!(X̄, dist.qmat, a_b, true, -2 * Δ),
        )
        return Composite{typeof(dist)}(qmat = ∂qmat), ∂a, ∂b
```

theogf:
Should we make a patch version bump followed by a minor version bump where we remove the deprecation warnings?

willtebbutt:
Sorry @theogf , which deprecations are you referring to?

theogf:
`kerneldiagmatrix` vs `kernelmatrix_diag` and some others like `degree` vs `d` etc...

willtebbutt:
Oh I see. Yeah, patch followed by minor release seems reasonable to me.

devmotion (approved):
Looks good to me :+1:

PR #212, opened 2020-12-31
Clarify how to set up the lengthscale in the docs

theogf:


docs/src/userguide.md
```diff
@@ -7,14 +7,15 @@ For example to create a square exponential kernel
 ```julia
   k = SqExponentialKernel()
 ```
+!!! tip "How do I set the lengthscale?" Instead of having lengthscale(s) for each kernel we use `Transform` objects (see [Transform](@ref)) which are directly going to act on the inputs before passing them to the kernel. 
+For example, if you want to to premultiply the input by 2.0, you can create your kernel with the following options:
```
devmotion, line 11:
```suggestion
For example, if you want to premultiply the input by 2.0, you can create your kernel with the following options:
```

docs/src/userguide.md
```diff
@@ -7,14 +7,15 @@ For example to create a square exponential kernel
...
 ```julia
-  k = transform(SqExponentialKernel(),ScaleTransform(2.0)) # returns a TransformedKernel
-  k = @kernel SqExponentialKernel() l=2.0 # Will be available soon
-  k = TransformedKernel(SqExponentialKernel(),ScaleTransform(2.0))
+  k = transform(SqExponentialKernel(), 2.0)) # returns a TransformedKernel
+  k = TransformedKernel(SqExponentialKernel(), ScaleTransform(2.0))
 ```
+In the example of the [SqExponentialKernel](@ref), you can reproduce the usual definition, $$\exp\left(-\frac{\|x-x'\|^2}{\rho^2}$$, by using the following `Transform` : `transform(SqExponentialKernel(), 1 / ρ`.
```
devmotion, line 16:
```suggestion
In the example of the [SqExponentialKernel](@ref), you can reproduce the usual definition, $$\exp\left(-\frac{\|x-x'\|^2}{\rho^2}\left)$$, by using `transform(SqExponentialKernel(), 1 / ρ)`. 
```

docs/src/userguide.md
```diff
@@ -7,14 +7,15 @@ For example to create a square exponential kernel
...
-Instead of having lengthscale(s) for each kernel we use `Transform` objects (see [Transform](@ref)) which are directly going to act on the inputs before passing them to the kernel.
-For example to premultiply the input by 2.0 we create the kernel the following options are possible
+!!! tip "How do I set the lengthscale?" Instead of having lengthscale(s) for each kernel we use `Transform` objects (see [Transform](@ref)) which are directly going to act on the inputs before passing them to the kernel. 
+For example, if you want to to premultiply the input by 2.0, you can create your kernel with the following options:
 ```julia
-  k = transform(SqExponentialKernel(),ScaleTransform(2.0)) # returns a TransformedKernel
-  k = @kernel SqExponentialKernel() l=2.0 # Will be available soon
-  k = TransformedKernel(SqExponentialKernel(),ScaleTransform(2.0))
+  k = transform(SqExponentialKernel(), 2.0)) # returns a TransformedKernel
+  k = TransformedKernel(SqExponentialKernel(), ScaleTransform(2.0))
```
devmotion, line 14: I'm curious - is it useful to export `TransformedKernel` or would it be sufficient to only recommend `transform`?

theogf: I know some people use it if they are more comfortable with having the right constructor name.

docs/src/userguide.md
```diff
@@ -79,9 +80,12 @@ For example :
...
 What if you want to differentiate through the kernel parameters? Even in a highly nested structure such as :
 ```julia
-  k = transform(0.5*SqExponentialKernel()*MaternKernel()+0.2*(transform(LinearKernel(),2.0)+PolynomialKernel()),[0.1,0.5])
+  k = transform(
+        0.5 * SqExponentialKernel() * MaternKernel()
+      + 0.2 * (transform(LinearKernel(), 2.0) + PolynomialKernel()),
+      [0.1, 0.5])
 ```
-One can get the array of parameters to optimize via `params` from `Flux.jl`
+One can access the array of trainable parameters via `params` from `Flux.jl`
```
devmotion, line 88: Does `params` return a flattened vector of parameters or a named tuple? Alternatively, one could use `first(Functors.functor(k))` to obtain a named tuple of parameters (Functors.functor is explained in more detail in the "Custom kernel" section).

docs/src/userguide.md
devmotion: I don't mind to be honest. I was just curious why the documentation mentions two different ways here. In general, it is always possible to import `TransformedKernel` with `using KernelFunctions: TransformedKernel`, so it's more a design decision if we want to encourage users to use `transform` rather than `TransformedKernel`.

docs/src/userguide.md
theogf: Well it actually would return a tuple of arrays to be exact. I think giving the Flux approach is probably more appropriate for first-timers. Functors is nice and light-weight but you kind of need to know what you are doing :D

devmotion: Sure, most people will be unfamiliar with Functors and how to implement `Functors.functor` manually. But here it would be only about interpreting the output of `first(Functors.functor(k))` which is something such as
```julia
(x = 1, y = [1, 2, 3])
```
in the example in the [Functors README](https://github.com/FluxML/Functors.jl) and seems quite nice.

docs/src/userguide.md
theogf: That's fair, I will remove this line, anyway the docs are more complete in the `Transform` section

docs/src/userguide.md
theogf: Ok I will change it accordingly. Probably the best thing would be to give a straight example where the parameters can be optimized.

docs/src/userguide.md
```diff
@@ -7,14 +7,15 @@ For example to create a square exponential kernel
...
-Instead of having lengthscale(s) for each kernel we use `Transform` objects (see [Transform](@ref)) which are directly going to act on the inputs before passing them to the kernel.
-For example to premultiply the input by 2.0 we create the kernel the following options are possible
+!!! tip "How do I set the lengthscale?" Instead of having lengthscale(s) for each kernel we use `Transform` objects (see [Transform](@ref)) which are directly going to act on the inputs before passing them to the kernel. 
+For example, if you want to premultiply the input by 2.0, you can create your kernel with the following options:
 ```julia
-  k = transform(SqExponentialKernel(),ScaleTransform(2.0)) # returns a TransformedKernel
-  k = @kernel SqExponentialKernel() l=2.0 # Will be available soon
-  k = TransformedKernel(SqExponentialKernel(),ScaleTransform(2.0))
+  k = transform(SqExponentialKernel(), 2.0)) # returns a TransformedKernel
+  k = TransformedKernel(SqExponentialKernel(), ScaleTransform(2.0))
```
theogf, line 14:
```suggestion
```

docs/src/userguide.md
theogf, line 88:
```suggestion
One can access the named tuple of trainable parameters via `Functors.functor` from `Functors.jl`.
This mean that in practice you can implicitly optimize the kernel parameters by calling:
```julia
using Flux
hp = params(k)
gradient(hp) do
	... some loss function on the kernel ....
end
```

```

devmotion:
Errors caused by https://github.com/FluxML/Zygote.jl/issues/873.

st--:
This PR's changes have been incorporated in #213.

Issue #163, opened 2020-09-07
Exporting ColVecs, RowVecs and vec_of_vecs


sharanry:
Shouldn't ColVecs, RowVecs and vec_of_vecs be available to the users? If we are to promote the use of these abstractions, I think it is prudent we export them.

willtebbutt:
Absolutely. We've also not properly documented their use. A PR that both exports them and makes documentation on the matter visible to users would be great.

codeboy5:
Hey I would like to take up this issue .
Any tips on how should I proceed ?

theogf:
Hi @codeboy5 , thanks for contributing!
For solving this issue you should open a PR where you add the export to the forementioned types/methods and write detailed docs about what they are doing/how to use them.

st--:
Should also address this comment: https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/213#discussion_r553543552

st--:
Should also add a docstring for `defaultobs` and remove the hard-coded constant from the docs (see https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/232/files#r559398795)

devmotion:
No, as I wrote in the comment, I disagree. In my opinion there is no need for a docstring since it is an internal detail that is irrelevant for users. It is only relevant that by default columns are observations.

theogf:
Solved in #159

PR #213, opened 2021-01-06
Documentation improvements (inc. lengthscale explanation) and Matern12Kernel alias

st--:
Just some minor edits for clarity and typos while going through the docs, and adds Matern12Kernel as alias for ExponentialKernel (in line with the explicitly defined Matern32Kernel and Matern52Kernel). Incorporates the lengthscales explanation from #212.


docs/src/kernels.md
```diff
@@ -110,15 +110,23 @@ The [`MaternKernel`](@ref) is defined as
...
 
-### Matern 3/2 Kernel
+### Matérn 1/2 Kernel
+
+The Matérn 1/2 kernel is defined as
+```math
+  k(x,x') = \exp\left(-|x-x'|\right).
+```
+It is equivalent to the [`ExponentialKernel`](@ref).
+
```
theogf, line 120: I would probably mention that there is no `Matern12Kernel`, and that one should just call `ExponentialKernel`

theogf (commented):
It looks good, I don't see any issue. You don't necessarily have to correct spacing in latex since it is going to be ignored anyway.

docs/src/kernels.md
st--: Yeah, that's why I didn't mention `Matern12Kernel`, but there _is_ a Matérn-1/2 kernel.:)

st--: I've clarified this. Though a `Matern12Kernel` function that constructs an ExponentialKernel might be a helpful alias for users ? Similar to how there's a `LaplacianKernel`?

docs/src/kernels.md
```diff
@@ -223,43 +231,41 @@ where $i\in\{-1,0,1,2,3\}$ and coefficients $a_i$, $b_i$ are fixed and residuals
 
 ### Transformed Kernel
 
-The [`TransformedKernel`](@ref) is a kernel where input are transformed via a function `f`
+The [`TransformedKernel`](@ref) is a kernel where inputs are transformed via a function `f`:
 
 ```math
   k(x,x';f,\widetile{k}) = \widetilde{k}(f(x),f(x')),
```
devmotion, line 237:
```suggestion
  k(x,x';f,\widetilde{k}) = \widetilde{k}(f(x),f(x')),
```

docs/src/kernels.md
```diff
@@ -223,43 +231,41 @@ where $i\in\{-1,0,1,2,3\}$ and coefficients $a_i$, $b_i$ are fixed and residuals
...
-Where $\widetilde{k}$ is another kernel and $\sigma^2 > 0$.
+where $\widetilde{k}$ is another kernel and $\sigma^2 > 0$.
 
 ### Kernel Sum
 
-The [`KernelSum`](@ref) is defined as a sum of kernels
+The [`KernelSum`](@ref) is defined as a sum of kernels:
 
 ```math
   k(x, x'; \{k_i\}) = \sum_i k_i(x, x').
```
devmotion, line 255:
```suggestion
  k(x, x'; \{k_i\}) = \sum_i k_i(x, x')
```

docs/src/kernels.md
```diff
@@ -223,43 +231,41 @@ where $i\in\{-1,0,1,2,3\}$ and coefficients $a_i$, $b_i$ are fixed and residuals
...
 ```
 
-### KernelProduct
+### Kernel Product
 
-The [`KernelProduct`](@ref) is defined as a product of kernels
+The [`KernelProduct`](@ref) is defined as a product of kernels:
 
 ```math
   k(x,x';\{k_i\}) = \prod_i k_i(x,x').
```
devmotion, line 263:
```suggestion
  k(x,x';\{k_i\}) = \prod_i k_i(x,x')
```

docs/src/metrics.md
```diff
@@ -1,16 +1,17 @@
 # Metrics
 
 KernelFunctions.jl relies on [Distances.jl](https://github.com/JuliaStats/Distances.jl) for computing the pairwise matrix.
-To do so a distance measure is needed for each kernel. Two very common ones can already be used : `SqEuclidean` and `Euclidean`.
-However all kernels do not rely on distances metrics respecting all the definitions. That's why  additional metrics come with the package such as `DotProduct` (`<x,y>`) and `Delta` (`δ(x,y)`).
-Note that every `SimpleKernel` must have a defined metric defined as :
+To do so a distance measure is needed for each kernel. Two very common ones can already be used: `SqEuclidean` and `Euclidean`.
+However, not all kernels rely on distance metrics respecting all the definitions as in Distances.jl. For this reason, KernelFunctions.jl provides additional "metrics" such as `DotProduct` ($\langle x, y \rangle$) and `Delta` ($\delta(x,y)$).
+
+Note that every `SimpleKernel` must have a metric, specified as
```
devmotion, line 7: The whole page only applies to `SimpleKernel`. It is not correct that "a distance measure is needed for each kernel".

docs/src/transform.md
```diff
@@ -1,9 +1,9 @@
...
 `Transform` is the object that takes care of transforming the input data before distances are being computed. It can be as standard as `IdentityTransform` returning the same input, or multiplying the data by a scalar with `ScaleTransform` or by a vector with `ARDTransform`.
-There is a more general `Transform`: `FunctionTransform` that uses a function and apply it on each vector via `mapslices`.
-You can also create a pipeline of `Transform` via `TransformChain`. For example `LowRankTransform(rand(10,5))∘ScaleTransform(2.0)`.
+There is a more general `Transform`: `FunctionTransform` that uses a function and applies it on each vector via `mapslices`.
+You can also create a pipeline of `Transform` via `TransformChain`. For example, `LowRankTransform(rand(10,5))∘ScaleTransform(2.0)`.
 
-One apply a transformation on a matrix or a vector via `KernelFunctions.apply(t::Transform,v::AbstractVecOrMat)`
+A transformation can be applied to a matrix or a vector via `KernelFunctions.apply(t::Transform, v::AbstractVecOrMat)`
 
 Check the list on the [API page](@ref Transforms)
```
devmotion, line 9:
```suggestion
Check the list on the [API page](@ref Transforms).
```

docs/src/transform.md
```diff
@@ -1,9 +1,9 @@
...
+# Input Transforms
 
 `Transform` is the object that takes care of transforming the input data before distances are being computed. It can be as standard as `IdentityTransform` returning the same input, or multiplying the data by a scalar with `ScaleTransform` or by a vector with `ARDTransform`.
-There is a more general `Transform`: `FunctionTransform` that uses a function and apply it on each vector via `mapslices`.
-You can also create a pipeline of `Transform` via `TransformChain`. For example `LowRankTransform(rand(10,5))∘ScaleTransform(2.0)`.
+There is a more general `Transform`: `FunctionTransform` that uses a function and applies it on each vector via `mapslices`.
+You can also create a pipeline of `Transform` via `TransformChain`. For example, `LowRankTransform(rand(10,5))∘ScaleTransform(2.0)`.
 
-One apply a transformation on a matrix or a vector via `KernelFunctions.apply(t::Transform,v::AbstractVecOrMat)`
+A transformation can be applied to a matrix or a vector via `KernelFunctions.apply(t::Transform, v::AbstractVecOrMat)`
```
devmotion, line 7:
```suggestion
A transformation `t` can be applied to a matrix or a vector `v` via `KernelFunctions.apply(t, v)`.
```

docs/src/kernels.md
devmotion: In my opinion, the documentation should only mention kernels that are implemented explicitly. I.e., I think a section about Matern 1/2 kernels should only be included if an alias such as `Matern12Kernel` exists.

Personally, I prefer if the amount of aliases is minimized to keep the API clean and therefore I am in favour of not including one for the exponential kernel (and therefore I would also not include a section about the Matern 1/2 kernel in the documentation). I think the term "Laplacian kernel" is much more common and (maybe) justifies an alias.

devmotion:
Errors are caused by https://github.com/FluxML/Zygote.jl/issues/873.

docs/src/kernels.md
```diff
@@ -153,7 +161,7 @@ The [`PiecewisePolynomialKernel`](@ref) is defined as
   r =& x^\top P x',\\
   j =& \lfloor \frac{D}{2}\rfloor + V + 1,
 ```
-where $x\in \mathbb{R}^D$, $V \in \{0,1,2,3\} and $P$ is a positive definite matrix.
+where $x\in \mathbb{R}^D$, $V \in \{0,1,2,3\} and $P$ is a positive-definite matrix.
 $f$ is a piecewise polynomial (see source code).
```
devmotion, line 165: "(see source code)" should link to the code (if needed).

docs/src/kernels.md
```diff
@@ -153,7 +161,7 @@ The [`PiecewisePolynomialKernel`](@ref) is defined as
   r =& x^\top P x',\\
   j =& \lfloor \frac{D}{2}\rfloor + V + 1,
 ```
```
devmotion, line 163: This is missing a
```suggestion
\end{aligned}
```
```
and a corresponding `\begin{aligned}`. However, maybe it would be better to separate the definition of `r` and `j` in a separate equation or sentence. Maybe one could also make the dependence of `r` on `x` and `x'` more explicit (not sure about the parameters `P` and `V` in `r` and `j`, it might become too verbose at some point).

File: docs/src/kernels.md [hunk 10]
```diff
@@ -96,11 +96,11 @@ The [`GaborKernel`](@ref) is defined as
   k(x,x'; l,p) =& h(x-x';l,p)\\
   h(u;l,p) =& \exp\left(-\cos\left(\pi \sum_i \frac{u_i}{p_i}\right)\sum_i \frac{u_i^2}{l_i^2}\right),
```
devmotion, line 97: Either
```suggestion
\begin{aligned}
  k(x,x'; l,p) =& h(x-x';l,p)\\
  h(u;l,p) =& \exp\left(-\cos\left(\pi \sum_i \frac{u_i}{p_i}\right)\sum_i \frac{u_i^2}{l_i^2}\right),
\end{aligned}
```
or just
```suggestion
  k(x,x'; l,p) = \exp\left(-\cos\left(\pi \sum_i \frac{{(x - x')}_i}{p_i}\right)\sum_i \frac{{(x - x')}_i^2}{l_i^2}\right),
```

src/basekernels/exponential.jl
```diff
@@ -42,8 +42,9 @@ iskroncompatible(::ExponentialKernel) = true
 
 Base.show(io::IO, ::ExponentialKernel) = print(io, "Exponential Kernel")
 
-## Alias ##
+## Aliases ##
 const LaplacianKernel = ExponentialKernel
+const Matern12Kernel = ExponentialKernel
```
devmotion, line 47: As mentioned in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/213#discussion_r553016003, I would add neither the alias nor the documentation section for Matern 1/2.

docs/src/kernels.md
willtebbutt: FWIW (and to increase the sample size by 1) I commonly use the name `Matern-1/2` to refer to the exponential / laplacian kernel. Incidentally, I would say that I see Matern-1/2 and Exponential used similarly often, while I see Laplacian used less often 🤷

willtebbutt: While I broadly agree that aliases are a bit annoying from a maintenance perspective, we do currently do them. If we were to introduce any extra aliases in this PR, I would want them to have a docstring that explicitly says "this is an alias for kernel X` so that it's easy for the user to figure out what is an alias for what based on the auto-generated documentation / typing `? kernel_name` at the REPL.

devmotion: Interesting, it's the exact opposite for me, usually I only encounter Exponential and Laplacian. But a sample size of 1 is sufficient here to convince me that an alias might be useful :slightly_smiling_face: 

I strongly agree that we should document all aliases, not only on the webpage but also with a docstring.

st--: I added the Matern12Kernel alias because in my (personal) experience Matern-1/2 is used ~two thirds of the time, exponential kernel less than a third of the time, and I'd not even come across "Laplacian kernel" before. Also, as a user when I see "Matern32Kernel" and "Matern52Kernel" but there isn't a "Matern12Kernel" that's rather confusing - and if you end up using MaternKernel (whose default is \nu=1/2), you end up with a rather less efficient implementation.

I was thinking the aliases should be documented, and had silently assumed that consts couldn't get a docstring as the other ones don't have one! Happy to add the docstrings. Is it simply a matter of having a string in the line immediately preceding each const?

docs/src/kernels.md
st--: Yes, that'd be good! Can you suggest what to link to, and how you would do the link so it renders nicely in the docs?

docs/src/kernels.md
st--: Everywhere else uses correct punctuation in equations (e.g. commas or full stops depending on context) - why would you want to change it here ?

docs/src/kernels.md
st--: done

docs/src/kernels.md
st--: tried to clean this up, let me know what you think.

docs/src/kernels.md
devmotion: I am used to use punctuation in equations only if they are part of the sentence but not if they are separated by, e.g., a colon as in this case.

However, I would be fine with using punctuation in such cases as well if we agree on it and apply it consistently. Currently, usually in these cases no punctuation is used but it is inconsistent.

docs/src/kernels.md
theogf: Either that or you can also write a separate docstring like this :
```
"""
	Matern12Kernel()

This is a MaternKernel
"""
Matern12Kernel
```

docs/src/metrics.md
st--: tried to address this, let me know what you think

docs/src/kernels.md
devmotion: One can link to e.g. docstrings (see https://juliadocs.github.io/Documenter.jl/stable/man/guide/#Cross-Referencing) but unfortunately the docstring of `PiecewisePolynomialKernel` does not explain it. So probably the docstring should be fixed and list the polynomials in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/d6eab2bfbf5c772bd293752665ea42dd087866b4/src/basekernels/piecewisepolynomial.jl#L37-L41.

docs/src/kernels.md
devmotion: E.g., we don't use punctuation in the documentation of TensorProduct (both on master and in this PR).

docs/src/kernels.md
st--: Ok I've added docstrings for all the aliases. Happy with that?

docs/src/kernels.md
st--: There's probably more inconsistency in the punctuation; if you feel strongly about it would you be willing to go and sort it out yourself (maybe in a separate PR)? I hope you find this PR helpful and an improvement as it is but I'm loathe to spend more time nitpicking on dots and commas right now!

docs/src/kernels.md
```diff
@@ -142,18 +149,16 @@ The [`PeriodicKernel`](@ref) is defined as
...
 ## Piecewise Polynomial Kernel
 
-The [`PiecewisePolynomialKernel`](@ref) is defined as
-
+The [`PiecewisePolynomialKernel`](@ref) is defined for $x\in \mathbb{R}^D$ and $V \in \{0,1,2,3\}$ as
 ```math
-  k(x,x'; P, V) =& \max(1 - r, 0)^{j + V} f(r, j),\\
-  r =& x^\top P x',\\
-  j =& \lfloor \frac{D}{2}\rfloor + V + 1,
+  k(x,x'; P, V) &= \max(1 - r, 0)^{j + V} f(r, j),
```
devmotion, line 158: IMO it is not really beneficial to introduce `r` and hide `x` and `x'` on the RHS. What do you think about
```suggestion
  k(x,x'; P, V) = \max(1 - x^\mathsf{T} P x', 0\right)^{j + V} f_V\left(x^\mathsf{T} P x', j\right),
```
to make it a bit more explicit and highlight the dependency of `f` on `V`?

docs/src/kernels.md
```diff
@@ -142,18 +149,16 @@ The [`PeriodicKernel`](@ref) is defined as
...
 ```
 
-where $r$ has the same dimension as $x$ and $r_i >0$.
+where $r$ has the same dimension as $x$ and $r_i > 0$.
 
 ## Piecewise Polynomial Kernel
 
-The [`PiecewisePolynomialKernel`](@ref) is defined as
-
+The [`PiecewisePolynomialKernel`](@ref) is defined for $x\in \mathbb{R}^D$ and $V \in \{0,1,2,3\}$ as
```
devmotion, line 156:
```suggestion
The [`PiecewisePolynomialKernel`](@ref) is defined for $x, x' \in \mathbb{R}^D$ and $V \in \{0,1,2,3\}$ as
```

devmotion: Or maybe mention `P` as well since it is also one of the arguments?
```suggestion
The [`PiecewisePolynomialKernel`](@ref) is defined for $x, x' \in \mathbb{R}^D$, a positive-definite matrix $P \in \mathbb{R}^{D \times D}$, and $V \in \{0,1,2,3\}$ as
```

docs/src/kernels.md
devmotion: :+1: I added some additional comments.

docs/src/kernels.md
devmotion: Sure, I am fine with postponing this :slightly_smiling_face:

docs/src/metrics.md
devmotion: Sounds much better, thanks a lot!

docs/src/kernels.md
st--: I think the maths make more sense in the docs than in the docstring; I've updated it accordingly.

devmotion:
BTW should we merge https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/214 first?

docs/src/kernels.md
st--: done

docs/src/kernels.md
st--: done

docs/src/kernels.md
devmotion: It looks really good now!

src/basekernels/piecewisepolynomial.jl
```diff
@@ -2,13 +2,12 @@
     PiecewisePolynomialKernel{V}(maha::AbstractMatrix)
 
 Piecewise Polynomial covariance function with compact support, V = 0,1,2,3.
-The kernel functions are 2v times continuously differentiable and the corresponding
-processes are hence v times  mean-square differentiable. The kernel function is:
+The kernel functions are 2V times continuously differentiable and the corresponding
+processes are hence V times mean-square differentiable. The kernel function is:
```
devmotion, line 6: :+1:

Maybe you can just copy the equation and text in the documentation to the docstring so that the additional information is shown to users in the REPL as well?

docs/src/kernels.md
willtebbutt: Looks great, thanks!

docs/src/kernels.md
```diff
@@ -142,19 +149,23 @@ The [`PeriodicKernel`](@ref) is defined as
...
-
+The [`PiecewisePolynomialKernel`](@ref) is defined for $x, x'\in \mathbb{R}^D$, a positive-definite matrix $P \in \mathbb{R}^{D \times D}$, and $V \in \{0,1,2,3\}$ as
 ```math
-  k(x,x'; P, V) =& \max(1 - r, 0)^{j + V} f(r, j),\\
-  r =& x^\top P x',\\
-  j =& \lfloor \frac{D}{2}\rfloor + V + 1,
+  k(x,x'; P, V) = \max(1 - x^\top P x', 0)^{j + V} f_V(x^\top P x', j),
+```
+where $j = \lfloor \frac{D}{2}\rfloor + V + 1$, and $f_V$ are polynomials defined as follows:
+````math
```
devmotion, line 161: There seems to be some problem with the LaTeX code in the preview: https://juliagaussianprocesses.github.io/KernelFunctions.jl/previews/PR213/kernels/:
```suggestion
```math
```

st--: woops! good catch

src/basekernels/piecewisepolynomial.jl
st--: I think the exact definition may be interesting/helpful in the full docs but when I think of what I want in the REPL I'd rather not have the docstring contain what's effectively the code implementation. A bit over-redundant to have the same equations three times in three separate places?

st--:
I realised I hadn't actually touched userguide.md ... have a look through that before you merge!

docs/src/userguide.md
```diff
@@ -2,24 +2,23 @@
...
 ```julia
   k = SqExponentialKernel()
 ```
+Instead of having lengthscale(s) for each kernel we use [`Transform`](@ref) objects which act on the inputs before passing them to the kernel.
+For example, to premultiply the input by 2.0 (equivalent to a lengthscale of 0.5) we can use the following options:
 ```julia
-  k = transform(SqExponentialKernel(),ScaleTransform(2.0)) # returns a TransformedKernel
-  k = @kernel SqExponentialKernel() l=2.0 # Will be available soon
```
st--, line 14: I think it's confusing to have something in the docs that isn't actually implemented. This example should be added back in by #38 if and when that's ready to go in.

docs/src/userguide.md
```diff
@@ -2,24 +2,23 @@
...
 ```
-Check the [`Transform`](@ref) page to see the other options.
-To premultiply the kernel by a variance, you can use `*` or create a `ScaledKernel`
+Check the [`Transform`](@ref) page to see all available transforms.
+
+To premultiply the kernel by a variance, you can use `*` or create a `ScaledKernel`:
 ```julia
   k = 3.0*SqExponentialKernel()
-  k = ScaledKernel(SqExponentialKernel(),3.0)
-  @kernel 3.0*SqExponentialKernel()
```
st--, line 22: Removed as not currently implemented, to be added by #38 (see above).

docs/src/userguide.md
```diff
@@ -29,60 +28,59 @@ To compute the kernel function on two vectors you can call
...
   x2 = rand(3)
-  k(x1,x2)
+  k(x1, x2)
 ```
 
 ## Creating a kernel matrix
 
 Kernel matrices can be created via the `kernelmatrix` function or `kerneldiagmatrix` for only the diagonal.
-An important argument to give is the dimensionality of the input `obsdim`. It tells if the matrix is of the type `# samples X # features` (`obsdim`=1) or `# features X # samples`(`obsdim`=2) (similarly to [Distances.jl](https://github.com/JuliaStats/Distances.jl))
+An important argument to give is the data layout of the input `obsdim`. It specifies whether the number of observed data points is along the first dimension (`obsdim=1`, i.e. the matrix shape is number of samples times number of features) or along the second dimension (`obsdim=2`, i.e. the matrix shape is number of features times number of samples), similarly to [Distances.jl](https://github.com/JuliaStats/Distances.jl). If not given explicitly, `obsdim` defaults to [`defaultobs`](@ref).
```
st--, line 37: I found this paragraph particularly confusing, and it's important for new users to get right; I've tried to clarify the explanation, but happy to iterate on this!

docs/src/userguide.md
```diff
@@ -29,60 +28,59 @@ To compute the kernel function on two vectors you can call
...
-Make sure that `k` is a vector compatible with such constructions (with `iskroncompatible`). Both method will return a . For those methods to work in your code you need to include `using Kronecker` first
-- For a Nystrom approximation : `kernelmatrix(nystrom(k, X, ρ, obsdim = 1))` where `ρ` is the proportion of sampled used.
+Make sure that `k` is a kernel compatible with such constructions (with `iskroncompatible(k)`). Both methods will return a Kronecker matrix. For those methods to work in your code you need to include `using Kronecker` first.
+- For a Nystrom approximation: `kernelmatrix(nystrom(k, X, ρ, obsdim=1))` where `ρ` is the fraction of data samples used in the approximation.
 
 ## Composite kernels
 
-One can create combinations of kernels via `KernelSum` and `KernelProduct` or using simple operators `+` and `*`.
-For example :
+Sums and products of kernels are also valid kernels. They can be created via `KernelSum` and `KernelProduct` or using simple operators `+` and `*`.
```
st--, line 68: This could probably do with adding an explanation of TensorProduct as well. Or just refer to the "Composite Kernels" section in kernels.md: could one of you help me out with how to get the cross-reference to a specific section in another file right?

docs/src/userguide.md
```diff
@@ -2,24 +2,23 @@
...
-For example to create a square exponential kernel
+To create a kernel object, choose one of the pre-implemented kernels, see [Base Kernels](@ref), or create your own, see [Creating your own kernel](@ref).
+For example, a square exponential kernel is created by
 ```julia
   k = SqExponentialKernel()
 ```
+Instead of having lengthscale(s) for each kernel we use [`Transform`](@ref) objects which act on the inputs before passing them to the kernel.
+For example, to premultiply the input by 2.0 (equivalent to a lengthscale of 0.5) we can use the following options:
```
st--, line 11: I belatedly remembered that #212 is touching upon this as well.... @theogf should I just incorporate your changes from there in this PR ?

docs/src/userguide.md
st--:
```suggestion
Sums and products of kernels are also valid kernels. They can be created via `KernelSum` and `KernelProduct` or using the equivalent operators `+` and `*`, or `TensorProduct`. See [Composite Kernels](@ref) for the detailed definitions.
```
?
(What happens if the section heading is ambiguous?)

docs/src/userguide.md
devmotion: I agree, IIRC it was one of the comments I had in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/212 has well which addresses this section.

devmotion: Maybe you should sync the PR with https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/212?

docs/src/userguide.md
willtebbutt: I think this section actually needs to also talk about the `ColVecs` / `RowVecs` way of doing this as well -- it's arguably the primary way in which `KernelFunctions`s handles multi-dimensional inputs, with the `obsdim` kwarg just wrapping the input matrix in either a `ColVecs` (`obsdim=2`) vs `RowVecs` (`obsdim=1`).

src/basekernels/piecewisepolynomial.jl
devmotion: It would only be two places, wouldn't it? Users shouldn't have to search through the source code on Github, so it would only be the docstring and the webpage. IMO the docstring should contain the mathematical expressions of the kernels and this is not the case here. It's just a coincidence in Julia that the code is very similar to the mathematical expressions :wink: 

What's your opinion, @willtebbutt @theogf?

docs/src/userguide.md
st--: ok, should we leave that (related to #163) for a separate PR ?

st--:
Thinking about https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/213#discussion_r553540255 made me wonder about whether `[Transform](@ref)` goes to the definition in the API or to transform.md. I renamed the primary heading of transform.md to "Input Transforms" which seems to make a lot more sense to me, so presumably we also need to update the pages dict in make.jl?

src/basekernels/piecewisepolynomial.jl
theogf: I think having it in two places is not a problem!

docs/src/userguide.md
theogf: By syncing you mean merging its current status to this PR? Sure why not!

docs/src/userguide.md
theogf: Yeah that seems like the most practical option!

theogf:
> Thinking about [#213 (comment)](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/213#discussion_r553540255) made me wonder about whether `[Transform](@ref)` goes to the definition in the API or to transform.md. I renamed the primary heading of transform.md to "Input Transforms" which seems to make a lot more sense to me, so presumably we also need to update the pages dict in make.jl?

I think : https://juliadocs.github.io/Documenter.jl/stable/man/syntax/#Duplicate-Headers would be the correct approach.

src/basekernels/piecewisepolynomial.jl
willtebbutt: My position is that the maths should be written out separately in this case (because while the code is reasonably concise, the maths is clearer), and that it should be in the docstring and autogenerated into the docs.

@st--  can I ask: is your concern that the REPL will have too much information if the maths is in the docstring, or is it more the duplication that's the problem?

docs/src/userguide.md
willtebbutt: Good shout.

docs/src/transform.md
```diff
@@ -1,9 +1,9 @@
...
-There is a more general `Transform`: `FunctionTransform` that uses a function and apply it on each vector via `mapslices`.
-You can also create a pipeline of `Transform` via `TransformChain`. For example `LowRankTransform(rand(10,5))∘ScaleTransform(2.0)`.
+There is a more general `Transform`: `FunctionTransform` that uses a function and applies it on each vector via `mapslices`.
+You can also create a pipeline of `Transform` via `TransformChain`. For example, `LowRankTransform(rand(10,5))∘ScaleTransform(2.0)`.
 
-One apply a transformation on a matrix or a vector via `KernelFunctions.apply(t::Transform,v::AbstractVecOrMat)`
+A transformation `t` can be applied to a matrix or a vector `v` via `KernelFunctions.apply(t, v)`.
 
-Check the list on the [API page](@ref Transforms)
+Check the full list of provided transforms on the [API page](@ref Transforms).
```
devmotion, line 9: Maybe one could also just move the list of transforms here (the full API section is a bit overwhelming to be honest). But if we want to do this, it might best to change it in a separate PR and maybe for some other sections as well.

src/basekernels/piecewisepolynomial.jl
st--: @willtebbutt A bit of both, really. If we ensure all kernel types have the exact kernel function definition in the docstring, what's the point of kernel.md - why not just point to the api.md instead then? I think that's a) something that requires a bit more discussion first and b) is beyond the scope of this PR and should probably be done in its own PR.

docs/src/userguide.md
st--: done

docs/src/userguide.md
st--: done

docs/src/transform.md
st--: Yeah. Is it possible to split the API page into individual pages for the "submodules"?

devmotion: Sure, docstrings can be placed everywhere with `@docs` blocks: https://juliadocs.github.io/Documenter.jl/stable/man/syntax/#@docs-block

docs/src/kernels.md
```diff
@@ -142,19 +149,23 @@ The [`PeriodicKernel`](@ref) is defined as
...
 ## Piecewise Polynomial Kernel
 
-The [`PiecewisePolynomialKernel`](@ref) is defined as
-
+The [`PiecewisePolynomialKernel`](@ref) is defined for $x, x'\in \mathbb{R}^D$, a positive-definite matrix $P \in \mathbb{R}^{D \times D}$, and $V \in \{0,1,2,3\}$ as
 ```math
-  k(x,x'; P, V) =& \max(1 - r, 0)^{j + V} f(r, j),\\
-  r =& x^\top P x',\\
-  j =& \lfloor \frac{D}{2}\rfloor + V + 1,
+  k(x,x'; P, V) = \max(1 - x^\top P x', 0)^{j + V} f_V(x^\top P x', j),
```
devmotion, line 158: Actually, I just noticed that the documentation was incorrect: the code uses `r = sqrt(x^T P x')`.

st--: fixed

docs/src/userguide.md
```diff
@@ -2,88 +2,95 @@
...
+
+!!! tip "How do I set the lengthscale?"
+Instead of having lengthscale(s) for each kernel we use [`Transform`](@ref) objects which act on the inputs before passing them to the kernel. Note that the transforms such as [`ScaleTransform`](@ref) and [`ARDTransform`](@ref) _multiply_ the input by a scale factor, which corresponds to the _inverse_ of the lengthscale.
+For example, a lengthscale of 0.5 is equivalent to premultiplying the input by 2.0, and you can create the corresponding kernel as follows:
 ```julia
-  k = transform(SqExponentialKernel(),ScaleTransform(2.0)) # returns a TransformedKernel
-  k = @kernel SqExponentialKernel() l=2.0 # Will be available soon
-  k = TransformedKernel(SqExponentialKernel(),ScaleTransform(2.0))
+  k = transform(SqExponentialKernel(), 2.0)
+  k = TransformedKernel(SqExponentialKernel(), ScaleTransform(2.0))  # equivalent explicit construction
```
st--, line 16: can remove this again if you prefer @devmotion

docs/src/userguide.md
```diff
@@ -2,88 +2,95 @@
...
 ```julia
-  k = transform(0.5*SqExponentialKernel()*MaternKernel()+0.2*(transform(LinearKernel(),2.0)+PolynomialKernel()),[0.1,0.5])
+  k = transform(
+        0.5 * SqExponentialKernel() * Matern12Kernel()
+      + 0.2 * (transform(LinearKernel(), 2.0) + PolynomialKernel()),
+      [0.1, 0.5])
 ```
-One can get the array of parameters to optimize via `params` from `Flux.jl`
-
+One can access the named tuple of trainable parameters via `Functors.functor` from `Functors.jl`.
```
st--, line 88: Should we explicitly mention using `first(Functors.functor(k))` as an example (see https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/212#discussion_r550477655)?

docs/src/userguide.md
devmotion: I think it is better to state only the more compact and recommended form `k = transform(....)` in the documentation.

st--: Done. There is probably some matter of "getting used to it" to this, but I have to say I find having to specify the inverse lengthscale to be deeply unintuitive compared to other GP packages. At least `transform(k, ScaleTransform(2.0))` is still explicitly saying what it is (and there could be a `LengthscaleTransform()` that divides instead of multiplying...), but I think it is not very helpful to new users that `transform(k, 2.0)` will give you a kernel with lengthscale 0.5. What are your thoughts on that ?

st--:
@theogf @willtebbutt @devmotion I've addressed all review comments as far as I can tell, though I've got a few questions left open. Other than that, what else is missing for you to approve this PR?

docs/src/userguide.md
willtebbutt: I think I agree that we could do more here to signpost what's going on here / offer a set-the-lengthscale option. I also think we could probably do with providing a bit more of an explanation for the rationale behind the transform approach to setting length scales, but that's a job for a different PR.

docs/src/userguide.md
```diff
@@ -2,88 +2,94 @@
...
-For example to create a square exponential kernel
+To create a kernel object, choose one of the pre-implemented kernels, see [Base Kernels](@ref), or create your own, see [Creating your own kernel](@ref).
+For example, a squared exponential kernel is created by
 ```julia
   k = SqExponentialKernel()
 ```
+
+!!! tip "How do I set the lengthscale?"
```
willtebbutt, line 11: It looks like this isn't rendering quite as expected in the docs -- the accompanying text appears outside of the box.

willtebbutt:
Thanks for the contribution @st-- . I'm happy for this to be merged as-is, but I'm also happy to discuss / address the following in this PR or another.

I'm concerned is that we're duplicating documentation a bit at the minute by having both docstrings and having the entire `Kernel`s section written out manually. This creates an opportunity for the docs to decouple from the docstrings and increases the maintenance burden. I wonder whether it would make sense, therefore, to consolidate this documentation -- if we were to move the concise mathematical description of the kernels from the docs to the docstrings, and just move /duplicate bits of the `API` section in the `Kernel Functions` section we could get the best of both worlds?

If we decide that this is out of scope for this PR, I'll open a separate issue to discuss.

docs/src/userguide.md
st--: Yeah I think that'd be great. It'd also be good to demonstrate how to set bijectors on parameters in this framework! But probably better to turn it into a separate issue/PR.

docs/src/userguide.md
st--: Thanks, I didn't actually check how the `!!! tip` syntax works (the content of a tip/note needs to be indented), and didn't actually check the generated docs locally. Fixed now!

src/basekernels/piecewisepolynomial.jl
st--: Covered by #218

st--:
I've opened #217 and #218 to discuss/address some of the points that were brought up by this PR. I'd be in favour of keeping PRs small (this one went from just fixing a few typos to taking several days already :sweat_smile: ), and split out these larger discussion/change points.

Two open questions:
1. https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/213#discussion_r553540255: Is that a good(-enough) way of incorporating TensorProduct for now? (And/)Or should we open another issue for that?
2. https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/213#discussion_r553882886: Should we explicitly mention using `first(Functors.functor(k))` as an example (see [#212 (comment)](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/212#discussion_r550477655))?

devmotion:
I'm fine with the PR, I think further improvements (in particular, updating the docstrings and using them on the webpage) should be done in other PRs. I also think 1 and 2 should not be addressed in this PR - I think it is better to not mention `TensorProduct` there for now instead of forcing it into this section which only mentions sums and products, and in my opinion most users won't call or look into Functors.functor, so we might not need an example there (currently).

docs/src/userguide.md
willtebbutt: This looks like a good change to me.

docs/src/userguide.md
willtebbutt: This seems like a good idea to me -- a brief sentence would probably suffice. @devmotion @theogf any thoughts?

willtebbutt (approved):
Looks like I forgot to hit approve.

devmotion (approved):
Ah I also forgot it :smile:

docs/src/userguide.md
theogf: So I checked and this is not completely true unfortunately:
```julia
using Functors
k = 2.5 * transform(SqExponentialKernel(), 2.0)
first(functor(k)) == ((kernel = Squared Exponential Kernel
	- Scale Transform (s = 2.0), σ² = [2.5]))
```

st--:
Yay! Happy to contribute my first PR on here. I've added points 1. and 2. to #221 to keep track of them.

PR #215, opened 2021-01-07
Fix tests

devmotion:
This PR fixes the test errors caused by https://github.com/FluxML/Zygote.jl/issues/873. It replaces `MaternKernel` (of order 3/2) with `Matern32Kernel` which is evaluated without the Bessel function.


willtebbutt (approved):
LGTM. Just needs a patch bump.

PR #220, opened 2021-01-09
Introduce JuliaFormatter with style=blue

### LINKED ISSUE #219: CI: introduce automatic code formatting (e.g. JuliaFormatter)?
In Python projects I've made very good experiences with automatic code formatting using `black`, and I was delighted to find Julia having similar packages such as https://github.com/domluna/JuliaFormatter.jl. Using an automatic code formatter tends to make code rather more readable (automatically breaking long statements etc., nice indentation, etc.), consistent, whilst at the same time removing any discussions around how to format code in PRs. What are your thoughts on introducing that in here? For that I would a) add a check to the CI that the code is consistent with the formatter, and b) make it easy (Makefile/script?) to format the code before pushing. It would also require a one-off PR to get the currently existing code formatted.

st--:
Introduces a format check (using JuliaFormatter / https://github.com/julia-actions/julia-format) with `style = "blue"`. Resolves #219.


willtebbutt:
I like this a lot 👍

src/KernelFunctions.jl
```diff
@@ -113,10 +119,10 @@ include("zygote_adjoints.jl")
 include("test_utils.jl")
 
 function __init__()
-    @require Kronecker="2c470bb0-bcc8-11e8-3dad-c9649493f05e" begin
+    @require Kronecker = "2c470bb0-bcc8-11e8-3dad-c9649493f05e" begin
```
devmotion, line 122: It seems a bit inconsistent to add whitespace here, compared with the conventions for keyword arguments - is this according to BlueStyle or a JuliaFormatter thing?

devmotion:
Did the merge commit revert any changes from JuliaFormatter? I guess it would be good to add the Github action for checking the format in this PR as well to be sure it passes.

.github/workflows/ci.yml
```diff
@@ -47,3 +47,10 @@ jobs:
         with:
             github-token: ${{ secrets.GITHUB_TOKEN }}
             path-to-lcov: ./lcov.info
+  format:
```
st--, line 50: Separate from the others as format-testing should be independent of the host/version matrix and it seems more sensible to only run it once (and then it's easier to distinguish format failure from the "regular" tests)

.github/workflows/ci.yml
```diff
@@ -47,3 +47,10 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+    - uses: julia-actions/julia-format@master
+      with:
+        args: -v .
```
devmotion, line 56: To me it seems this approach should not be used anymore. The julia-format repo says that the legacy Docker version that is used here is deprecated and it is "advised you use the format check example above". So instead we should adapt https://github.com/julia-actions/julia-format/blob/master/workflows/format_check.yml, as suggested in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/219#issuecomment-757319168.

devmotion (changes requested):

.github/workflows/ci.yml
st--: Ah, I had misunderstood the README. Updated accordingly.

st--: A bit unfortunate that's not made clearer; the https://github.com/domluna/JuliaFormatter.jl README might as well point directly to the format_check.yml then ...

willtebbutt: Hmm yeah, you do have to dig around a bit to figure out what the right thing to do is, and I'm still not really sure I understand what's going on -- it seems like it's otherwise a well-documented package. Would be great if one of you could open an issue about it.

.github/workflows/ci.yml
```diff
@@ -47,3 +47,34 @@ jobs:
...
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        julia-version: [1.3]
+        julia-arch: [x64]
+        os: [ubuntu-latest]
+    steps:
+      - uses: julia-actions/setup-julia@latest
+        with:
+          version: ${{ matrix.julia-version }}
```
devmotion, line 61: This seems unnecessarily complicated and uses an outdated Julia version, I suggest:
```suggestion
    runs-on: ubuntu-latest
    steps:
      - uses: julia-actions/setup-julia@latest
        with:
          version: 1
```

.github/workflows/ci.yml
```diff
@@ -47,3 +47,34 @@ jobs:
...
+
```
devmotion, line 62:
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -47,3 +47,34 @@ jobs:
...
+      - uses: actions/checkout@v1
+      - name: Install JuliaFormatter and format
+        # The version of JuliaFormatter used is just to show how to specify the version. The latest
+        # version would be preferrable.
+        run: |
+          julia  -e 'using Pkg; Pkg.add(PackageSpec(name="JuliaFormatter", version="0.12.2"))'
```
devmotion, line 68: I don't like that the version number has to be updated manually if there is a new version of JuliaFortmatter. We could just install the latest version but this means we would end up with breaking changes without realizing. Ideally, we would add a Project.toml file in some directory and be notified about breaking releases by CompatHelper.

.github/workflows/ci.yml
st--: https://github.com/julia-actions/julia-format/issues/15

.github/workflows/ci.yml
st--: I agree, that'd be nice. I'm not sure how you'd go about sorting it out with the Project.toml (feel free to make those changes on this branch or a separate PR). I do think it's overall better to pin the version of the formatter, as otherwise it can be hard to figure out how exactly to get the local formatter be in sync with the cloud format-checker..

.github/workflows/ci.yml
```diff
@@ -47,3 +47,29 @@ jobs:
         with:
             github-token: ${{ secrets.GITHUB_TOKEN }}
             path-to-lcov: ./lcov.info
+
+  format:
```
st--, line 51:
```suggestion
  format-check:
```

.github/workflows/ci.yml
devmotion: > I'm not sure how you'd go about sorting it out with the Project.toml

Similar to the package, tests, and documentation, we can specify the SemVer-compatible versions in a separate project environment. CompatHelper will automatically create a PR on Github if a new breaking release is available (we are not notified about non-breaking releases).

> it can be hard to figure out how exactly to get the local formatter be in sync with the cloud format-checker.

I think specifying the version that we use in a project environment is the best we can do. We can't control which version of JuliaFormatter developers use in their editors or install globally but if they load our project environment it will always be compatible.

src/KernelFunctions.jl
```diff
@@ -12,7 +12,7 @@ if !isfile(joinpath(@__DIR__, "update_v0.8.0"))
         This kernel now divides the squared distance by 2 to align with standard practice.
         This warning will be removed in 0.9.0.
         """;
-        color = Base.info_color(),
+        color=Base.info_color(),
```
devmotion, line 15: Break formatting on purpose to test action:
```suggestion
        color = Base.info_color(),
```

devmotion: Seems to work, tests failed correctly.

devmotion:
I simplified the Github action and pushed some changes, but since I got involved now someone else has to approve the PR.

src/KernelFunctions.jl
nickrobinson251: I agree. I suspect this is a JuliaFormatter (or CSTParser) issue

theogf (approved):

theogf:
@devmotion I think you still need to indicate that your change requests have been satisfied.

devmotion (approved):

Issue #177, opened 2020-09-25
Piecewise Polynomial Kernel


willtebbutt:
The `PiecewisePolynomialKernel` is something of an oddity in that our current implementation accepts an argument that parametrises a distance. While the dimensionality of this argument is needed so that we can ensure positive definiteness, the warping of the input space that it performs can be achieved via `transform`.

It would be great if someone could refactor this kernel for our next minor release (we might want to think about getting into the habit of maintaining a dev branch or something)

devmotion:
It seems, similarly maybe the Mahalanobis kernel could be removed completely?

PR #225, opened 2021-01-10
Deprecate use of Mahalanobis distance

devmotion:
Fixes https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/177 and corrects the documentation of PiecewisePolynomialKernel.


src/basekernels/piecewisepolynomial.jl
```diff
@@ -1,48 +1,60 @@
...
+function PiecewisePolynomialKernel(; v::Int=0, maha=nothing, d::Int=-1)
+    if maha !== nothing
+        Base.depwarn("keyword argument `maha` is deprecated", :PiecewisePolynomialKernel)
+        d = size(maha, 1)
+        return transform(PiecewisePolynomialKernel{v}(d), LinearTransform(cholesky(maha).U))
+    else
+        return PiecewisePolynomialKernel{v}(d)
     end
-    return (maha = x.maha,), reconstruct_kernel
 end
```
theogf, line 44: Should a check on the positivity of `d` be made?

src/basekernels/piecewisepolynomial.jl
```diff
@@ -1,48 +1,60 @@
...
-# See also https://github.com/FluxML/Functors.jl/issues/3#issuecomment-626747663
-function Functors.functor(::Type{<:PiecewisePolynomialKernel{V}}, x) where V
-    function reconstruct_kernel(xs)
-        return PiecewisePolynomialKernel{V}(xs.maha)
+# TODO: remove `maha` keyword argument in next breaking release
+function PiecewisePolynomialKernel(; v::Int=0, maha=nothing, d::Int=-1)
+    if maha !== nothing
+        Base.depwarn("keyword argument `maha` is deprecated", :PiecewisePolynomialKernel)
+        d = size(maha, 1)
+        return transform(PiecewisePolynomialKernel{v}(d), LinearTransform(cholesky(maha).U))
```
theogf, line 40: Is this needed since you created the `deprecated.jl` file?

src/basekernels/piecewisepolynomial.jl
devmotion: It is checked in the inner constructor.

src/basekernels/piecewisepolynomial.jl
devmotion: The other deprecation only covers the inner constructor.

devmotion: Since we can't dispatch on keyword arguments, we can't add both the deprecated and the new version of the constructor with keyword arguments but instead have to move the deprecation in the function body.

theogf:
Maybe it would be worth noting somewhere in the docs how to create a "MahalonobisKernel" (and similarly use the Maha.. distance for PieceWise..), there might be people looking for this exactly without realizing that one can use `LinearTransform` instead.

docs/src/kernels.md
```diff
@@ -153,17 +153,19 @@ where $r$ has the same dimension as $x$ and $r_i > 0$.
 
 ## Piecewise Polynomial Kernel
 
-The [`PiecewisePolynomialKernel`](@ref) is defined for $x, x'\in \mathbb{R}^D$, a positive-definite matrix $P \in \mathbb{R}^{D \times D}$, and $V \in \{0,1,2,3\}$ as
+The [`PiecewisePolynomialKernel`](@ref) is defined for $x, x' \in \mathbb{R}^d$ and
+$v \in \{0,1,2,3\}$ as
 ```math
-  k(x,x'; P, V) = \max(1 - \sqrt{x^\top P x'}, 0)^{j + V} f_V(\sqrt{x^\top P x'}, j),
```
st--, line 158: It looks like it's the old documentation that was wrong but just wanted to flag up that this is a change from xᵀ P x' to (x - x')ᵀ P (x - x')

docs/src/kernels.md
```diff
@@ -153,17 +153,19 @@ where $r$ has the same dimension as $x$ and $r_i > 0$.
...
+k(x, x'; v) = \max(1 - \|x - x'\|, 0)^{j + V} f_v(\|x - x'\|, j),
```
st--, line 159:
```suggestion
k(x, x'; v) = \max(1 - \|x - x'\|, 0)^{j + v} f_v(\|x - x'\|, j),
```

src/basekernels/maha.jl
```diff
@@ -1,25 +0,0 @@
...
-    MahalanobisKernel(P::AbstractMatrix)
-
-Mahalanobis distance-based kernel given by
-```math
-    κ(x,y) =  exp(-r^2), r^2 = maha(x,P,y) = (x-y)'* P *(x-y)
-```
-where the matrix P is the metric.
-
-"""
-struct MahalanobisKernel{T<:Real, A<:AbstractMatrix{T}} <: SimpleKernel
```
st--, line 11: Should there be a `deprecated.jl` version of this as well? (It could just construct an ExponentialKernel with a different metric?)

docs/src/kernels.md
devmotion: Yes, currently the docs are wrong.

docs/src/kernels.md
```diff
@@ -153,17 +153,19 @@ where $r$ has the same dimension as $x$ and $r_i > 0$.
 
 ## Piecewise Polynomial Kernel
 
-The [`PiecewisePolynomialKernel`](@ref) is defined for $x, x'\in \mathbb{R}^D$, a positive-definite matrix $P \in \mathbb{R}^{D \times D}$, and $V \in \{0,1,2,3\}$ as
+The [`PiecewisePolynomialKernel`](@ref) is defined for $x, x' \in \mathbb{R}^d$ and
```
st--, line 156: Might just be a question of style but I've usually seen upper-case $D$ for dimensionality (as it's a kind of constant... and then $d$ can be used to iterate over dimensions e.g. in tensor product kernel; and similarly for N number of data points and n iterating from 1 to N, etc.).

src/basekernels/maha.jl
st--: Woops, missed that that was already done. Sorry!

docs/src/kernels.md
devmotion: The lower-case `d` is used due to the constructor with keyword arguments. I try to use upper-case names only for modules, types, and constructors. The documentation on the webpage is just copied from the docstring (and should eventually be replaced with the docstring). However, maybe the name of the keyword argument should be more verbose and one should use something like `dim` instead?

st--: Yes, that sounds good! Do you want it to be `dim` or `dims`? Would be good if this was consistent across the codebase (and with wider Julia). `dims` is what's used in PeriodicKernel, though `dims` makes me think of a tuple (d1, d2, d3) whereas here it's just a scalar Int, so maybe `dim` everywhere?

I'd generally be in favour of more explicit and preferably ascii (keyword/struct member) names, was thinking of opening an issue for discussing that (e.g. `period` instead of `r` for PeriodicKernel, `weights`, `means`, `vars` for spectral_mixture_kernel)...

devmotion: Actually, `PeriodKernel` only uses `r` as parameter and keyword argument name (which I agree is not very descriptive). There exist some strange outer constructor with positional arguments of name `dims` but the name is not user-facing.

Regardless, I think the name `dims` is more common in the Julia ecosystem, e.g., as keyword argument to reducing functions such as `sum`. In these functions usually one can specify multiple dimensions (I assume therefore the name `dims` is used instead of `dim`). So the use of "dimensions" is a bit different here, but maybe `dims` would still be a better choice.

devmotion:
I guess I should set up JuliaFormatter support for my editors but I don't want to mess with all the other projects that don't use it... Maybe calling `format` manually is the easiest option?

src/basekernels/piecewisepolynomial.jl
```diff
@@ -1,56 +1,67 @@
...
+_f(::PiecewisePolynomialKernel{1}, r, j) = 1 + (j + 1) * r
+_f(::PiecewisePolynomialKernel{2}, r, j) = 1 + (j + 2) * r + (j^2 + 4 * j + 3) / 3 * r^2
+function _f(::PiecewisePolynomialKernel{3}, r, j)
     return 1 +
-           (j + 3) * r +
-           (6 * j^2 + 36j + 45) / 15 * r .^ 2 +
-           (j^3 + 9 * j^2 + 23j + 15) / 15 * r .^ 3
+        (j + 3) * r +
+        (6 * j^2 + 36j + 45) / 15 * r ^ 2 +
+        (j^3 + 9 * j^2 + 23j + 15) / 15 * r ^ 3
```
devmotion, line 53: JuliaFormatter wants this to be formatted
```julia
    return 1 +
           (j + 3) * r +
           (6 * j^2 + 36j + 45) / 15 * r ^ 2 +
           (j^3 + 9 * j^2 + 23j + 15) / 15 * r ^ 3
```
This feels weird since usually we just increase the indentation level but do not align operators or lines. Maybe I missed it but I couldn't find any information about similar cases in BlueStyle - is this a bug in JuliaFormatter?

docs/src/kernels.md
st--: If it strictly expects a single scalar, I'd prefer to be explicit about that and call it just `dim`.

src/basekernels/piecewisepolynomial.jl
```diff
@@ -1,56 +1,65 @@
 """
```
st--, line 1:
```suggestion
@doc raw"""
```
and then not having to escape all the LaTeX backslashes

src/basekernels/piecewisepolynomial.jl
```diff
@@ -1,56 +1,101 @@
...
+        return quote
+            return @evalpoly(r, $((:(coeffs[$i]) for i in 1:N)...))
+        end
+    end
+else
+    _evalpoly(r, coeffs) = evalpoly(r, coeffs)
 end
 
-metric(κ::PiecewisePolynomialKernel) = Mahalanobis(κ.maha)
+kappa(κ::PiecewisePolynomialKernel, r) = max(1 - r, 0)^κ.alpha * _evalpoly(r, κ.coeffs)
```
st--, line 88: What was your motivation behind changing all of this from explicit functions to storing coefficients inside the type struct? (I found the previous version of the code much easier to understand and read.)

devmotion: It seems wasteful to re-compute the fixed coefficients every time the kernel function is evaluated. Moreover, `evalpoly` uses Horner's method and hence the evaluation of the polynomials is more stable and more efficient than the naive implementation before.

theogf (approved):

devmotion:
@willtebbutt Do you have any comments or objections to this PR? Since everything should be deprecated properly, it is not breaking.

willtebbutt (approved):
None whatsoever. This is an excellent change.

Project.toml
```diff
@@ -17,7 +17,7 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 ZygoteRules = "700de1a5-db45-46bc-99cf-38207098b444"
 
 [compat]
-Compat = "2.2, 3"
+Compat = "3.7"
```
devmotion, line 20: I forgot that we already depend on Compat which contains an implementation of `evalpoly` for Julia < 1.4. It was introduced in Compat 3.7. Is it OK to drop support for older versions?

@theogf @willtebbutt @st--

devmotion: Latest version of Compat is 3.25.

willtebbutt: No objections here.

PR #232, opened 2021-01-17
Rename `TensorProduct` and implement `TensorCore.tensor`

devmotion:
This PR renames `TensorProduct` to `KernelTensorProduct` (analogously to `KernelSum` and `KernelProduct`) and implements and exports `TensorCore.tensor === TensorCore.:⊗` for tensor product kernels (similar to the overloads for `KernelSum` and `KernelProduct`).

IMO the name `TensorProduct` is too generic to be owned and exported by KernelFunctions. The overload is nice to have, I think, especially with the more verbose `KernelTensorProduct`. [TensorCore](https://github.com/JuliaMath/TensorCore.jl) is a lightweight package that contains default implementations for tensor operations. Originally, the code was merged in https://github.com/JuliaLang/julia/pull/35150 which was later reverted in https://github.com/JuliaLang/julia/pull/35744 and moved to its own package. It seems other packages with tensor operations such as Kronecker.jl should also overload `TensorCore.tensor` or rename their functions if they are type piracy to avoid conflicts.


docs/make.jl
```diff
@@ -23,6 +28,8 @@ makedocs(;
         "Custom Kernels" => "create_kernel.md",
         "API" => "api.md",
     ],
+    strict=true,
+    checkdocs=:exports,
```
devmotion, line 32: Initially, there was an error in the implementation of the doctests but tests did not fail. This setting ensures that failing doctests, broken links, and docstrings of exported functions that are not included in the documentation will cause the build to fail. This required some additional fixes.

docs/src/api.md
```diff
@@ -74,7 +83,6 @@ kernelmatrix!
 kerneldiagmatrix
 kerneldiagmatrix!
 kernelpdmat
-kernelkronmat
```
devmotion, line 77: There is no docstring for `kernelkronmat`, therefore the build failed even after loading Kronecker.jl. I removed it for now.

docs/src/userguide.md
```diff
@@ -52,7 +52,7 @@ To evaluate the kernel function on two vectors you simply call the kernel object
 ## Creating a kernel matrix
 
 Kernel matrices can be created via the `kernelmatrix` function or `kerneldiagmatrix` for only the diagonal.
-An important argument to give is the data layout of the input `obsdim`. It specifies whether the number of observed data points is along the first dimension (`obsdim=1`, i.e. the matrix shape is number of samples times number of features) or along the second dimension (`obsdim=2`, i.e. the matrix shape is number of features times number of samples), similarly to [Distances.jl](https://github.com/JuliaStats/Distances.jl). If not given explicitly, `obsdim` defaults to [`defaultobs`](@ref).
+An important argument to give is the data layout of the input `obsdim`. It specifies whether the number of observed data points is along the first dimension (`obsdim=1`, i.e. the matrix shape is number of samples times number of features) or along the second dimension (`obsdim=2`, i.e. the matrix shape is number of features times number of samples), similarly to [Distances.jl](https://github.com/JuliaStats/Distances.jl). If not given explicitly, `obsdim` defaults to `2`.
```
devmotion, line 55: The link [`defaultobs`](@ref) was broken. There exists no docstring for `defaultobs` either, so I just changed it to `2`.

src/kernels/kerneltensorproduct.jl
```diff
@@ -0,0 +1,145 @@
+"""
```
devmotion, line 1: I only changed `TensorProduct` to `KernelTensorProduct` and added a new docstring. I renamed the file in a separate PR but somehow, in contrast to the tests, Github does not display the changes nicely.

src/matrix/kernelkroneckermat.jl
```diff
@@ -1,19 +1,19 @@
...
+    return Kronecker.kronecker(k, dims)
 end
 
 function kernelkronmat(
     κ::Kernel, X::AbstractVector{<:AbstractVector}; obsdim::Int=defaultobs
 )
     @assert iskroncompatible(κ) "The chosen kernel is not compatible for Kronecker matrices"
     Ks = kernelmatrix.(κ, X)
-    return K = reduce(⊗, Ks)
+    return K = reduce(Kronecker.:⊗, Ks)
```
devmotion, line 16: Since Kronecker does not implement `TensorCore.:⊗` but instead exports its own function `Kronecker.:⊗`, only the module is imported and `Kronecker.:⊗` and `Kronecker.kronecker` are called explicitly.

src/utils.jl
```diff
@@ -1,5 +1,3 @@
-hadamard(x, y) = x .* y
```
devmotion, line 1: TensorCore contains an implementation of `hadamard`.

test/kernels/kernelproduct.jl
```diff
@@ -1,43 +1,18 @@
...
-    @test (KernelProduct((k1, k2)) * k3).kernels == (k1, k2, k3)
-    @test (k3 * KernelProduct((k1, k2))).kernels == (k3, k1, k2)
 
     # Standardised tests.
     TestUtils.test_interface(k, Float64)
     test_ADs(
-        x -> SqExponentialKernel() * LinearKernel(; c=x[1]),
-        rand(1);
-        ADs=[:ForwardDiff, :ReverseDiff, :Zygote],
+        x -> KernelProduct(SqExponentialKernel(), LinearKernel(; c=exp(x[1]))), rand(1)
```
devmotion, line 15: It seems AD errors are fixed by enforcing non-negativity of the constant offset. Same below for `KernelSum` and `KernelTensorProduct`.

test/kernels/overloads.jl
```diff
@@ -0,0 +1,36 @@
+@testset "overloads" begin
```
devmotion, line 1: Consistent with the current test setup, I created a separate file that tests only the implementations of `+`, `*`, and `⊗`.

test/runtests.jl
```diff
@@ -1,7 +1,7 @@
 using KernelFunctions
 using AxisArrays
 using Distances
-using Kronecker
+using Kronecker: Kronecker
```
devmotion, line 4: Again this avoids conflicts between `Kronecker.:⊗` and `TensorCore.:⊗` while still loading `kernelkronmat`.

docs/src/userguide.md
theogf, line 55: Maybe we can add some details if your data is not in a matrix format?

src/kernels/overloads.jl
```diff
@@ -0,0 +1,22 @@
+for (M, op, T) in (
+    (:Base, :+, :KernelSum),
+    (:Base, :*, :KernelProduct),
+    (:TensorCore, :tensor, :KernelTensorProduct),
+)
+    @eval begin
+        $M.$op(k1::Kernel, k2::Kernel) = $T(k1, k2)
+
+        $M.$op(k1::$T, k2::$T) = $T(k1.kernels..., k2.kernels...)
```
theogf, line 9: The default behavior is/was that if we have one kernel `<:Tuple` and one kernel `<:AbstractVector`, one will get a  `<:Tuple`. Is it really a sensible choice?

docs/src/userguide.md
devmotion: I would like to avoid any additional unrelated changes in this PR. I think it would be best to address https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/163 and in particular https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/163#issuecomment-756693126 in a separate PR.

src/matrix/kernelkroneckermat.jl
st--: I'd think it'd be helpful to add your PR comment directly as a comment to the code (presumably if at some point Kronecker would reuse TensorCore's definitions, this ought to be changed?).

src/kernels/overloads.jl
devmotion: The nice thing about this default is that you don't end up with abstract types unexpectedly but, of course, it might not be friendly to the compiler. I haven't seen any instances of it in our examples and tests, so I am not sure how often such a combination happens in practice.

src/matrix/kernelkroneckermat.jl
```diff
@@ -1,19 +1,19 @@
-using .Kronecker
+using .Kronecker: Kronecker
```
devmotion, line 1:
```suggestion
# Since Kronecker does not implement `TensorCore.:⊗` but instead exports its own function
# `Kronecker.:⊗`, only the module is imported and Kronecker.:⊗ and Kronecker.kronecker are
# called explicitly.
using .Kronecker: Kronecker
```

src/matrix/kernelkroneckermat.jl
devmotion: I added a comment :+1:

devmotion:
AD errors on Julia 1.3 are caused by `GaborKernel` and seem unrelated.

test/runtests.jl
devmotion: I also opened an issue: https://github.com/MichielStock/Kronecker.jl/issues/80

docs/src/userguide.md
st--: Could it not be fixed in a better way by adding a docstring do `defaultobs` instead?

docs/make.jl
st--: Might make for easier reviewing to separate this into a separate PR? (Maybe not worth splitting this PR up at this point, but to keep in mind in the future?)

docs/src/userguide.md
devmotion: No, `defaultobs` is an internal variable to ensure consistency in our implementation but it is only relevant for users that the default dimension is 2 (therefore also the signatures in all docstrings mentions 2 explicitly instead of just copying the implementation with `defaultobs`). Additionally, as discussed in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/232#discussion_r559475324 and the issues linked therein, this section has to be changed and users should be pointed to `ColVecs` and `RowVecs` instead.

docs/make.jl
devmotion: The PR includes only the minimal set of changes that are required for the tests to pass.

devmotion:
Is there anything left that should be addressed?

devmotion:
This is ready for another round of review, most (all?) unrelated changes are included in the master branch.

src/kernels/kerneltensorproduct.jl
```diff
@@ -139,7 +139,6 @@ function printshifted(io::IO, kernel::KernelTensorProduct, shift::Int)
         for _ in 1:(shift + 1)
             print(io, "\t")
         end
-        print(io, "- ")
```
theogf: What the reason for removing the '- ' ?

devmotion: It is inconsistent, neither KernelSum nor KernelProduct use it.

theogf: Oh okay, then it would be nice to have it for `KernelSum` and `KernelProduct`

devmotion: I don't want to touch them here. In particular, since there are more inconsistencies in the `show` implementations in general, and in fact we don't implement them correctly. `show(io, obj)` should return a compact one-line representation, for verbose output one has to implement `show(io, MIME("text/plain"), obj)` instead.

theogf (approved):

PR #234, opened 2021-01-18
Example notebooks

st--:
I've played around a bit with the example notebooks. This isn't intended for merging as-is (if anything, it probably makes sense to split it into one PR per notebook), just to see where I've gone so far - happy to hear your thoughts! Heavily builds on top of @theogf's [examples branch](https://github.com/theogf/KernelFunctions.jl/tree/examples).

Following #303 and #314, examples are now split into separate PRs:
- Gaussian process priors -> #316
- training kernel parameters -> #317 
- kernel ridge regression -> #320
- support vector machine -> #321
- deep kernel learning -> #322 (maybe move this into AbstractGPs.jl?)


docs/make.jl
```diff
@@ -1,6 +1,22 @@
...
+end
+
+for filename in readdir(EXAMPLES_SRC)
+    endswith(filename, ".jl") || continue
+    name = splitext(filename)[1]
+    Literate.markdown(
+        joinpath(EXAMPLES_SRC, filename), EXAMPLES_OUT; name=name, documenter=true
+    )
+end
+
```
devmotion, line 19: I would suggest copying the setup in AbstractGPs (also below).

devmotion:
Regarding the general structure, I think we should copy the configuration in AbstractGPs. I assume it would also make sense to move examples that are based or require AbstractGPs to the AbstractGPs repo.

devmotion:
Generally, I think it would be nice to use a separate environment and hence a separate Project.toml + Manifest.toml for probably each of the examples such that they are easily reproducible and the dependencies are separated clearly. One would have to change this in AbstractGPs as well.

examples/deepkernellearning.jl
```diff
@@ -1,40 +1,78 @@
...
+# We use a couple of useful packages to plot and optimize
+# the different hyper-parameters
 using KernelFunctions
-using MLDataUtils
-using Zygote
 using Flux
 using Distributions, LinearAlgebra
 using Plots
+using ProgressMeter
+using AbstractGPs
```
devmotion, line 10: IMO the example should be moved to AbstractGPs.

examples/gaussianprocesspriors.jl
```diff
@@ -0,0 +1,90 @@
+# -*- coding: utf-8 -*-
```
devmotion, line 1:
```suggestion
```

examples/gaussianprocesspriors.jl
```diff
@@ -0,0 +1,90 @@
...
+# # Gaussian process prior samples
+#
+# The kernels defined in this package can also be used to specify the
+# covariance of a Gaussian process prior.
+# A Gaussian process (GP) is defined by its mean function $m(\cdot)$ and its covariance function or kernel $k(\cdot, \cdot')$:
```
devmotion, line 6: I have never seen the notation `\cdot'`, IMO `k(\cdot, \cdot)` already indicates that `k` is a function of two distinct arguments.

examples/gaussianprocesspriors.jl
```diff
@@ -0,0 +1,90 @@
...
+# $$
+#   f \sim \mathcal{GP}(m(\cdot), k(\cdot, \cdot')
+# $$
```
devmotion, line 9:
```suggestion
# ```math
#   f \sim \mathcal{GP}\big(m(\cdot), k(\cdot, \cdot')\big)
# ```
```

examples/gaussianprocesspriors.jl
```diff
@@ -0,0 +1,90 @@
...
+# The function values of the GP at a finite number of points $X = \{x_n\}_{n=1}^N$ follow a multivariate normal distribution with mean vector $\mathbf{m}$ and covariance matrix $\mathrm{K}$, where
+# $$
+# \begin{align}
+#   \mathbf{m}_i &= m(x_i) \\
+#   \mathrm{K}_{i,j} &= k(x_i, x_j)
+# \end{align}
+# $$
```
devmotion, line 16:
```suggestion
# ```math
# \begin{aligned}
#   \mathbf{m}_i &= m(x_i) \\
#   \mathrm{K}_{i,j} &= k(x_i, x_j)
# \end{aligned}
# ```
```

examples/kernel_ridge_regression.jl
```diff
@@ -0,0 +1,170 @@
+# -*- coding: utf-8 -*-
```
devmotion, line 1:
```suggestion
```

examples/kernel_ridge_regression.jl
```diff
@@ -0,0 +1,170 @@
...
+y_train = f_truth.(x_train) + noise
+y_test = f_truth.(x_test)
+
+plot(x_test, y_test; label=raw"$f(x)$")
+scatter!(x_train, y_train; label="observations")
+
+# For training inputs $\mathrm{X}=(\mathbf{x}_n)_{n=1}^N$ and observations $\mathbf{y}=(y_n)_{n=1}^N$, the linear regression weights $\mathbf{w}$ using the least-squares estimator are given by
+# $$
+# \mathbf{w} = (\mathrm{X}^\top \mathrm{X})^{-1} \mathrm{X}^\top \mathbf{y}
+# $$
```
devmotion, line 36:
```suggestion
# ```math
# \mathbf{w} = (\mathrm{X}^\top \mathrm{X})^{-1} \mathrm{X}^\top \mathbf{y}
# ```
```

examples/kernel_ridge_regression.jl
```diff
@@ -0,0 +1,170 @@
...
+# We predict at test inputs $\mathbf{x}_*$ using
+# $$
+# \hat{y}_* = \mathbf{x}_*^\top \mathbf{w}
+# $$
```
devmotion, line 40:
```suggestion
# ```math
# \hat{y}_* = \mathbf{x}_*^\top \mathbf{w}
# ```
```

examples/kernel_ridge_regression.jl
```diff
@@ -0,0 +1,170 @@
...
+y_pred = linear_regression(x_train, y_train, x_test)
+scatter(x_train, y_train; label="observations")
+plot!(x_test, y_pred; label="linear fit")
+
+# We can improve the fit by including additional features, i.e. generalizing to $\mathrm{X} = (\phi(x_n))_{n=1}^N$, where $\phi(x)$ constructs a feature vector for each input $x$. Here we include powers of the input, $\phi(x) = (1, x, x^2, \dots, x^d)$:
+
+function featurize_poly(x; degree=1)
+    xcols = [x .^ d for d in 0:degree]
+    return hcat(xcols...)
+end
```
devmotion, line 59:
```suggestion
featurize_poly(x; degree=1) = x .^ (0:degree)'
```
However, the function is so simple that it might be easier and more compact to just use `x .^ (0:degree)'` wherever it is needed.

examples/kernel_ridge_regression.jl
```diff
@@ -0,0 +1,170 @@
...
+
+function featurized_fit_and_plot(degree)
+    X = featurize_poly(x_train; degree=degree)
+    Xstar = featurize_poly(x_test; degree=degree)
+    y_pred = linear_regression(X, y_train, Xstar)
+    scatter(x_train, y_train; label=nothing)
+    p = plot!(x_test, y_pred; label=nothing, title="fit of order $degree")
+    return p
```
devmotion, line 67: IIRC the same can be achieved by
```suggestion
    scatter(x_train, y_train; legend=false, title="fit of order $degree")
    return plot!(x_test, y_pred)
```

examples/kernel_ridge_regression.jl
```diff
@@ -0,0 +1,170 @@
...
+# $$
+# We predict at test inputs $\mathbf{x}_*$ using
+# $$
+# \hat{y}_* = \mathbf{x}_*^\top \mathbf{w}
+# $$
+# This is implemented by `linear_regression`:
+
+function linear_regression(X::AbstractVecOrMat, y::AbstractVector, Xstar::AbstractVecOrMat)
+    weights = inv(X' * X) * X' * y
+    return Xstar * weights
```
devmotion, line 45: I think we should use the more accurate and preferred algorithm (see e.g. https://discourse.julialang.org/t/efficient-way-of-doing-linear-regression/31232/33?u=devmotion) and write
```suggestion
    weights = X \ y
    return Xstar * weights
```

examples/kernel_ridge_regression.jl
```diff
@@ -0,0 +1,170 @@
...
+# $$
+# \mathbf{w} = (\mathrm{X}^\top \mathrm{X})^{-1} \mathrm{X}^\top \mathbf{y}
+# $$
+# We predict at test inputs $\mathbf{x}_*$ using
+# $$
+# \hat{y}_* = \mathbf{x}_*^\top \mathbf{w}
+# $$
+# This is implemented by `linear_regression`:
+
+function linear_regression(X::AbstractVecOrMat, y::AbstractVector, Xstar::AbstractVecOrMat)
```
devmotion, line 43: Since we don't dispatch on the arguments anywhere the types are not helpful:
```suggestion
function linear_regression(X, y, Xstar)
```

examples/kernel_ridge_regression.jl
```diff
@@ -0,0 +1,170 @@
...
+# $$
+# \hat{y}_* = \mathbf{x}_*^\top \mathbf{w}
+# $$
+# This is implemented by `ridge_regression`:
+
+function ridge_regression(
+    X::AbstractVecOrMat, y::AbstractVector, Xstar::AbstractVecOrMat, lambda::Float64
+)
+    weights = inv(X' * X + lambda * I) * X' * y
+    return Xstar * weights
```
devmotion, line 97:
```suggestion
function ridge_regression(X, y, Xstar, lambda)
    weights = (X' * X + lambda * I) \ (X' * y)
    return Xstar * weights
```

examples/kernel_ridge_regression.jl
```diff
@@ -0,0 +1,170 @@
...
+end
+
+function regularized_fit_and_plot(degree, lambda)
+    X = featurize_poly(x_train; degree=degree)
+    Xstar = featurize_poly(x_test; degree=degree)
+    y_pred = ridge_regression(X, y_train, Xstar, lambda)
+    scatter(x_train, y_train; label=nothing)
+    p = plot!(x_test, y_pred; label=nothing, title=string(raw"$\lambda=", lambda, raw"$"))
+    return p
```
devmotion, line 106:
```suggestion
    scatter(x_train, y_train; legend=false, title="\$\\lambda=$lambda\$")
    return plot!(x_test, y_pred)
```

examples/kernel_ridge_regression.jl
```diff
@@ -0,0 +1,170 @@
...
+    k::Kernel,
+    X::AbstractVecOrMat,
+    y::AbstractVector,
+    Xstar::AbstractVecOrMat,
+    lambda::Float64,
+)
+    K = kernelmatrix(k, X)
+    kstar = kernelmatrix(k, Xstar, X)
+    return kstar * inv(K + lambda * I) * y
+end
```
devmotion, line 144:
```suggestion
function kernel_ridge_regression(k, X, y, Xstar, lambda)
    K = kernelmatrix(k, X)
    kstar = kernelmatrix(k, Xstar, X)
    return kstar * ((K + lambda * I) \ y)
end
```

examples/svm.jl
```diff
@@ -1,26 +1,75 @@
+# -*- coding: utf-8 -*-
```
devmotion, line 1:
```suggestion
```

examples/train_kernel_parameters.jl
```diff
@@ -0,0 +1,87 @@
+# -*- coding: utf-8 -*-
```
devmotion, line 1:
```suggestion
```

examples/train_kernel_parameters.jl
```diff
@@ -0,0 +1,87 @@
...
+scatter(x_train, y_train; lab="data")
+plot!(x_test, sinc; lab="true function")
+plot!(x_test, ŷ; lab="prediction")
+
+# We define the loss based on the L2 norm both
+# for the loss and the regularization
+
+function loss(θ)
+    ŷ = f(x_train, x_train, y_train, θ)
+    return sum(abs2, y_train - ŷ) + exp(θ[4]) * norm(ŷ)
```
devmotion, line 64: Usually one should avoid to close over non-constant global variables since it breaks type inference and therefore is slow.

examples/train_kernel_parameters.jl
```diff
@@ -0,0 +1,87 @@
...
+
+loss(log.(ones(4)))
+
+# ## Training the model
+
+θ = vcat(log.([1.0, 0.0, 0.01]), log(0.001)) # Initial vector
+anim = Animation()
+opt = Optimise.ADAGrad(0.5)
+for i in 1:30
+    grads = ForwardDiff.gradient(x -> loss(x), θ) # We compute the gradients given the kernel parameters and regularization
```
devmotion, line 77:
```suggestion
    grads = ForwardDiff.gradient(loss, θ) # We compute the gradients given the kernel parameters and regularization
```
I am wondering if Zygote or some other AD backend would be faster/slower.

examples/train_kernel_parameters.jl
```diff
@@ -0,0 +1,87 @@
...
+    Δ = Optimise.apply!(opt, θ, grads)
+    θ .-= Δ # We apply a simple Gradient descent algorithm
```
devmotion, line 79:
```suggestion
    Optimise.update!(opt, θ, grads)
```

examples/train_kernel_parameters.jl
```diff
@@ -0,0 +1,87 @@
...
+    p = scatter(
```
devmotion, line 80:
```suggestion
    scatter(
```

examples/train_kernel_parameters.jl
```diff
@@ -0,0 +1,87 @@
...
+    return sum(abs2, y_train - ŷ) + exp(θ[4]) * norm(ŷ)
+end
+
+# The loss with our starting point :
+
+loss(log.(ones(4)))
+
+# ## Training the model
+
+θ = vcat(log.([1.0, 0.0, 0.01]), log(0.001)) # Initial vector
```
devmotion, line 73:
```suggestion
θ = log.([1.0, 0.0, 0.01, 0.001]) # Initial vector
```

examples/train_kernel_parameters.jl
```diff
@@ -0,0 +1,87 @@
...
+)
+
+# From theory we know the prediction for a test set x given
+# the kernel parameters and normalization constant
+
+function f(x, x_train, y_train, θ)
+    k = kernelcall(θ[1:3])
+    return kernelmatrix(k, x, x_train) *
+           inv(kernelmatrix(k, x_train) + exp(θ[4]) * I) *
+           y_train
```
devmotion, line 48:
```suggestion
           ((kernelmatrix(k, x_train) + exp(θ[4]) * I) \ y_train)
```

examples/deepkernellearning.jl
st--: Happy to move this one, I think it's a bit less about kernels themselves. @theogf as this was yours, would like to hear what you think before removing it here ?

examples/gaussianprocesspriors.jl
st--: It's from jupytext. I'd like to see if it's possible to get Literate.jl and jupytext to play along 100% instead of just 98%... https://github.com/fredrikekre/Literate.jl/issues/131

devmotion: Even if it is supported, I would prefer the standard Literate syntax. Otherwise one either ends up with a mix of different styles or has to pay attention to use the non-standard jupytext formatting (at least non-standard in this environment). To me it seems this would add more confusion but not provide any benefits. If you want to play around and tune the code, I can recommend VSCode or the outdated Juno environment which support executing single blocks and lines of literate files automatically in the REPL.

docs/make.jl
devmotion: More precisely, better use (copied from AbstractGPs):
```suggestion
# Print `@debug` statements (https://github.com/JuliaDocs/Documenter.jl/issues/955)
if haskey(ENV, "GITHUB_ACTIONS")
    ENV["JULIA_DEBUG"] = "Documenter"
end

EXAMPLES = joinpath(@__DIR__, "..", "examples")
OUTPUT = joinpath(@__DIR__, "src", "examples")

ispath(OUTPUT) && rm(OUTPUT; recursive=true)

# Add links to binder and nbviewer below the first heading of level 1
function preprocess(content)
    sub = s"""
        \0
        #
        # [![](https://mybinder.org/badge_logo.svg)](@__BINDER_ROOT_URL__/examples/@__NAME__.ipynb)
        # [![](https://img.shields.io/badge/show-nbviewer-579ACA.svg)](@__NBVIEWER_ROOT_URL__/examples/@__NAME__.ipynb)
    """
    return replace(content, r"^# # [^\n]*"m => sub; count=1)
end

for file in readdir(EXAMPLES; join=true)
    endswith(file, ".jl") || continue
    Literate.markdown(file, OUTPUT; documenter=true, preprocess=preprocess)
    Literate.notebook(file, OUTPUT)
end

```

docs/make.jl
```diff
@@ -15,7 +31,13 @@ makedocs(;
...
         "Home" => "index.md",
         "User Guide" => "userguide.md",
-        "Examples" => "example.md",
+        "Examples" => [
+            "Kernel Ridge Regression" => "examples/kernel_ridge_regression.md",
+            "Training kernel parameters" => "examples/train_kernel_parameters.md",
+            "Gaussian process priors" => "examples/gaussianprocesspriors.md",
+            "SVM" => "examples/svm.md",
+            "Deep Kernel Learning" => "examples/deepkernellearning.md",
+        ],
```
devmotion, line 40: I think I would prefer if as in AbstractGPs the examples are listed after the regular documentation such that kernels, transforms, etc. are explained properly first. In addition, again the setup in AbstractGPs seems a bit better. I can't comment below, so I put the suggestion here but they should be moved below "API" in a second step:
```suggestion
        "Examples" => joinpath.("examples", filter(x -> endswith(x, ".md"), readdir(OUTPUT))),
```

examples/kernel_ridge_regression.jl
st--: My intention here was to illustrate how you end up with _kernelized_ ridge regression. That requires having `X * X'`. Happy to add a comment along the lines of "if you actually want to do linear regression, use <X\y / using GLM>"...

examples/kernel_ridge_regression.jl
st--: you don't think it'd be helpful for the reader to know what types it should be ?

theogf:
About the GP example(s) here are my thoughts.
It would be nice to have the following:
- One being Gaussian Process regression but where we do not use `AbstractGPs` at all (although we could mention it of course), we just reformulate the theory, maybe we could follow exactly the Rasmussen and Williams algorithm. We could additionally merge this example with the kernel parameters optimization
- One with the automated statistician where the kernel selection is made automatically. I think it's really fine to depend on `AbstractGPs` when the focus is on the kernels really.
- Same thing with the deep kernel learning, the GP layer is not the interesting one, I don't think depending on `AbstractGPs` make it necessary to move it to other docs

examples/train_kernel_parameters.jl
theogf: `ForwardDiff` is for me the AD the most bug-free and reliable, that's why I started with it

examples/deepkernellearning.jl
theogf: I replied more generally on the main comments.

examples/kernel_ridge_regression.jl
devmotion: I think it would be good to remove `inv`, see also the comments below. You basically never want to use it. So with the explicit normal equations you should still write
```suggestion
    weights = (X' * X) \ (X' * y)
    return Xstar * weights
```

examples/deepkernellearning.jl
```diff
@@ -1,40 +1,78 @@
...
-loss(k, λ)
+xmax = 3; # Limits
+N = 150
+noise = 0.01
+x_train = collect(eachrow(rand(Uniform(xmin, xmax), N))) # Training dataset
+target_f(x) = sinc(abs(x)^abs(x)) # We use sinc with a highly varying value
+target_f(x::AbstractArray) = target_f(first(x))
+y_train = target_f.(x_train) + randn(N) * noise
+x_test = collect(eachrow(range(xmin, xmax; length=200))) # Testing dataset
+spectral_mixture_kernel()
```
theogf, line 25: Should this be removed?

examples/kernel_ridge_regression.jl
devmotion: Personally, I don't think `AbstractVecOrMat` is particularly helpful since it does not even provide guarantees about the shape and dimensions. In general, I would put more focus on good explanations, consistent and descriptive variable names, and comments in the code if needed.

examples/kernel_ridge_regression.jl
theogf: Adding a prior on the weights would solve the problem. As it would be `inv(X' * X + lambda * I) * X' * y` or even `(X' * X + lambda * Y) \ X' * y`

theogf: Oh I see it's done later

examples/kernel_ridge_regression.jl
```diff
@@ -0,0 +1,170 @@
...
+# \mathbf{w} = (\mathrm{X}^\top \mathrm{X})^{-1} \mathrm{X}^\top \mathbf{y}
+# $$
+# we introduce the ridge parameter $\lambda$:
+# $$
+# \mathbf{w} = (\mathrm{X}^\top \mathrm{X} + \lambda \mathbb{1})^{-1} \mathrm{X}^\top \mathbf{y}
+# $$
+# As before, we predict at test inputs $\mathbf{x}_*$ using
+# $$
+# \hat{y}_* = \mathbf{x}_*^\top \mathbf{w}
+# $$
```
theogf, line 90:
```suggestion
# ```math
# \hat{y}_* = \mathbf{x}_*^\top \mathbf{w}
# ```
```

examples/kernel_ridge_regression.jl
```diff
@@ -0,0 +1,170 @@
...
+# \mathbf{w} = \mathrm{X}^\top (\mathrm{X} \mathrm{X}^\top + \lambda \mathbb{1})^{-1} \mathbf{y}
+# $$
+# where we can now replace the inner product with the kernel matrix,
+# $$
+# \mathbf{w} = \mathrm{X}^\top (\mathrm{K} + \lambda \mathbb{1})^{-1} \mathbf{y}
+# $$
+# And the prediction yields another inner product,
+# $$
+# \hat{y}_* = \mathbf{x}_*^\top \mathbf{w} = \langle \mathbf{x}_*, \mathbf{w} \rangle = \mathbf{k}_* (\mathrm{K} + \lambda \mathbb{1})^{-1} \mathbf{y}
+# $$
```
theogf, line 129:
```suggestion
# ```math
# \hat{y}_* = \mathbf{x}_*^\top \mathbf{w} = \langle \mathbf{x}_*, \mathbf{w} \rangle = \mathbf{k}_* (\mathrm{K} + \lambda \mathbb{1})^{-1} \mathbf{y}
# ```
```

docs/make.jl
st--: Could you clarify what you prefer about the AbstractGP.jl/docs/make.jl? The only functional difference I can see is the preprocess() - is there anything else I missed?

examples/gaussianprocesspriors.jl
st--: > but not provide any benefits

We can discuss trade-offs, but I don't think this claim is fair; I've explained my motivation on the Literate.jl issue I linked. Having the example notebooks in a format that can directly be edited interactively in jupyter notebooks makes it easier to contribute notebooks for people who are comfortable with that workflow (like me, but it's a widely popular workflow), without requiring people to use VSCode (which is a lot harder to get into than firing up a jupyter notebook).

docs/make.jl
devmotion: It is simpler and makes it easier to write and add examples :slightly_smiling_face: There's e.g. no need to manually try to split and join filenames, the `preprocess` function allows to write tutorials without thinking about the specific documentation setup and the additional links to the different outputs, and if the examples are added automatically (see below) one can just drop an additional file into `examples/` and it will be added automatically.

examples/gaussianprocesspriors.jl
devmotion: My impression is that in the Julia community the common workflow for writing Literate or Weave documents is to work with Juno or VSCode, to copy and run code in a REPL on the side, or rebuild the output with Literate and Weave intermittently. I think it's fine to support additional workflows but I think
- it should be officially support by Literate and Weave
- it should not impact the other apparently more common scenarios.
 
From the issue I get the impression that there won't be an official support for jupytext in the near future. And even if it is added, the second point seems still problematic. It seems jupytext requires a different syntax that deviates from the official syntax for Literate and Weave - if someone edits the document with jupytext the syntax of the document would have to change and so we might end up with a mix of the regular and the jupytext syntax. This seems confusing.

devmotion:
Regarding the AbstractGPs discussion, I still think that examples that use AbstractGPs should be added to the other repo or not use AbstractGPs if it is not the main focus. Mainly, because examples that use AbstractGPs would be interesting for users of AbstractGPs but would not show up when they go to the documentation of the package. While probably kernels are most commonly used with GPs, it seems more natural and easier to find for users if the examples in KernelFunctions are focused on, well, kernel functions and the ones in AbstractGPs on GPs.

theogf:
> Regarding the AbstractGPs discussion, I still think that examples that use AbstractGPs should be added to the other repo or not use AbstractGPs if it is not the main focus. Mainly, because examples that use AbstractGPs would be interesting for users of AbstractGPs but would not show up when they go to the documentation of the package. While probably kernels are most commonly used with GPs, it seems more natural and easier to find for users if the examples in KernelFunctions are focused on, well, kernel functions and the ones in AbstractGPs on GPs.

I get your point but it seems also absurd to restrict ourselves to only non-GP examples.
For example for the automated kernel selection, the fact that a GP is used is a minor point.

If this is a problem this is where the general website we set up could be useful. We could have these more complex examples there and mention them in the docs of each package.

st--:
Hi, it's been a while but I'd still like to get more & better examples into the KernelFunctions docs :) what do you think would be the best way forward ?

theogf:
> Hi, it's been a while but I'd still like to get more & better examples into the KernelFunctions docs :) what do you think would be the best way forward ?

Nice to see you back! The main problem is the discussion on which examples should belong here, or AbstractGPs. TBH I think it's a bit pointless to discuss this before the tutorials are even finished :laughing: . We should first finish them and then decide where they should go

I guess you can start by adding the review suggestions and try to get the Documentation action to pass so we can have previews :)

docs/make.jl
```diff
@@ -27,6 +27,7 @@ isempty(processes) || success(processes) || error("some examples were not run su
 # Build documentation
 using KernelFunctions
 using Documenter
+using Literate
```
st--, line 30:
```suggestion
```

st--:
Closing this as it's now split up into individual branches and PRs for each example (#316, #317, #320, #321, #321).

Issue #233, opened 2021-01-18
Rename kerneldiagmatrix ?


st--:
`kerneldiagmatrix` seems a somewhat confusing choice of name - intuitively, I would expect this to still return a (`Diagonal`) matrix, not the vector of the diagonal. `kernelmatrixdiag` or `kernelmatrix_diag` would seem more intuitive: then `kernelmatrix(...) |> diag === kernelmatrixdiag(...)` and it would also be in line with `cov_diag` and `mean_and_cov_diag` in AbstractGP.jl. A backwards-compatible deprecation should be easy. I'd be happy to make a PR with these changes if you're generally in agreement. Thoughts?

theogf:
`kernelmatrix_diag` sounds good to me.

PR #235, opened 2021-01-18
Rename kerneldiagmatrix[!] -> kernelmatrix_diag[!]

st--:
Addresses #233


devmotion (changes requested):
Looks good roughly but it seems you forgot to add the deprecations.

st--:
> Looks good roughly but it seems you forgot to add the deprecations.

So I did! I added the _tests_ of the deprecation and apparently was so used to writing code first I then didn't think anymore about actually deprecating them. Thanks for pointing that out :laughing:

src/deprecated.jl
```diff
@@ -7,3 +7,17 @@
...
+end
+@deprecate function kerneldiagmatrix(κ::Kernel, X, Y; obsdim::Int=2)
+    return kernelmatrix_diag(κ, X, Y; obsdim=obsdim)
+end
+@deprecate function kerneldiagmatrix!(K::AbstractVector, κ::Kernel, X; obsdim::Int=2)
+    return kernelmatrix_diag!(K, κ, X; obsdim=obsdim)
+end
+@deprecate function kerneldiagmatrix!(K::AbstractVector, κ::Kernel, X, Y; obsdim::Int=2)
+    return kernelmatrix_diag!(K, κ, X, Y; obsdim=obsdim)
+end
```
devmotion, line 23: `@deprecate` expects a different format. Also the deprecations are not correct and will error since the vector-based methods (the default) do not take any keyword arguments.
```suggestion
@deprecate kerneldiagmatrix(k::Kernel, x::AbstractVector) kernelmatrix_diag(k, x)
@deprecate kerneldiagmatrix(k::Kernel, x::AbstractVector, y::AbstractVector) kernelmatrix_diag(k, x, y)
@deprecate kerneldiagmatrix!(out::AbstractVector, k::Kernel, x::AbstractVector) kernelmatrix_diag!(out, k, x)
@deprecate kerneldiagmatrix!(out::AbstractVector, k::Kernel, x::AbstractVector, y::AbstractVector) kernelmatrix_diag!(out, k, x, y)

@deprecate kerneldiagmatrix(k::Kernel, X::AbstractMatrix; obsdim::Int=defaultobs) kernelmatrix_diag(k, X; obsdim=obsdim)
@deprecate kerneldiagmatrix(k::Kernel, X::AbstractMatrix, Y::AbstractMatrix; obsdim::Int=defaultobs) kernelmatrix_diag(k, X, Y; obsdim=obsdim)
@deprecate kerneldiagmatrix!(out::AbstractVector, k::Kernel, X::AbstractMatrix; obsdim::Int=defaultobs) kernelmatrix_diag!(out, k, X; obsdim=obsdim)
@deprecate kerneldiagmatrix!(out::AbstractVector, k::Kernel, X::AbstractMatrix, Y::AbstractMatrix; obsdim::Int=defaultobs) kernelmatrix_diag!(out, k, X, Y; obsdim=obsdim)
```

src/deprecated.jl
```diff
@@ -9,15 +9,15 @@
...
+)
+@deprecate kerneldiagmatrix(κ::Kernel, X, Y; obsdim::Int=2) kernelmatrix_diag(
+    κ, X, Y; obsdim=obsdim
+)
+@deprecate kerneldiagmatrix!(K::AbstractVector, κ::Kernel, X; obsdim::Int=2) kernelmatrix_diag!(
+    K, κ, X; obsdim=obsdim
+)
+@deprecate kerneldiagmatrix!(K::AbstractVector, κ::Kernel, X, Y; obsdim::Int=2) kernelmatrix_diag!(
+    K, κ, X, Y; obsdim=obsdim
+)
```
devmotion, line 23: As mentioned earlier, this is not correct. Instead you have to use:
```suggestion
@deprecate kerneldiagmatrix(k::Kernel, x::AbstractVector) kernelmatrix_diag(k, x)
@deprecate kerneldiagmatrix(k::Kernel, x::AbstractVector, Y::AbstractVector) kernelmatrix_diag(k, x, y)
@deprecate kerneldiagmatrix!(out::AbstractVector, k::Kernel, x::AbstractVector) kernelmatrix_diag!(out, k, x)
@deprecate kerneldiagmatrix!(out::AbstractVector, k::Kernel, x::AbstractVector, y::AbstractVector) kernelmatrix_diag!(out, k, x, y)

@deprecate kerneldiagmatrix(k::Kernel, X::AbstractMatrix; obsdim::Int=defaultobs) kernelmatrix_diag(
    k, X; obsdim=obsdim
)
@deprecate kerneldiagmatrix(k::Kernel, X::AbstractMatrix, Y::AbstractMatrix; obsdim::Int=defaultobs) kernelmatrix_diag(
    k, X, Y; obsdim=obsdim
)
@deprecate kerneldiagmatrix!(out::AbstractVector, k::Kernel, X::AbstractMatrix; obsdim::Int=defaultobs) kernelmatrix_diag!(
    out, k, X; obsdim=obsdim
)
@deprecate kerneldiagmatrix!(out::AbstractVector, k::Kernel, X::AbstractMatrix, Y::AbstractMatrix; obsdim::Int=defaultobs) kernelmatrix_diag!(
    out, k, X, Y; obsdim=obsdim
)
```

devmotion (changes requested):
The deprecations are not correct, please see my comments.

test/matrix/kernelmatrix.jl
```diff
@@ -40,7 +40,14 @@ KernelFunctions.kappa(::ToySimpleKernel, d) = exp(-d / 2)
...
             tmp_diag = Vector{Float64}(undef, length(x))
-            @test kerneldiagmatrix!(tmp_diag, k, x) ≈ kerneldiagmatrix(k, x)
+            @test kernelmatrix_diag!(tmp_diag, k, x) ≈ kernelmatrix_diag(k, x)
+
+            # Test deprecations
+            @test @test_deprecated kerneldiagmatrix(k, x) ≈ kernelmatrix_diag(k, x)
+
+            tmp_diag_dep = Vector{Float64}(undef, length(x))
+            @test_deprecated kerneldiagmatrix!(tmp_diag_dep, k, x)
+            @test tmp_diag_dep ≈ kernelmatrix_diag(k, x)
```
devmotion, line 50: Can we test all deprecations? I.e., one + two inputs, vectors (i.e. ColVecs or RowVecs), and matrices without and with different `obsdim`?

devmotion (changes requested):

test/matrix/kernelmatrix.jl
st--: We can do that of course! Could you give me a bit more insight into your motivation for that - what's the value of extensive tests of code-to-be-removed? I know the value of tests for code that's changed or new code, just here it seems more effort than worth it to me, so I'm wondering what I might be missing!

devmotion: If we don't test it, the deprecations might be incorrect, which in turn breaks existing code.

st--: Ok... in this case getting the deprecation test right (and making sure they're all covered) seems more effort than checking the deprecations (manually in deprecated.jl) as it's purely a rename. Is there some way of testing for un-tested deprecations? The coverage report seems to entirely ignore deprecated.jl.

st--:
I noticed that the in-place kernelmatrix[_diag]! forms only tested the return value, not whether it correctly got written to the passed-in array, so I've added those as well

devmotion:
Oh, good that you added these other tests as well!

> The tests of utils.jl and pairwise.jl in contrast only test the passed-in array and don't check the return value, should we add those as well?

Probably, but in my opinion it should be done in a separate PR.

devmotion (approved):
Thanks, looks good!

Can you bump the version? Then it can be merged and released, if tests pass.

src/matrix/kernelmatrix.jl
```diff
@@ -58,20 +58,20 @@ function kernelmatrix(κ::Kernel, x::AbstractVector, y::AbstractVector)
...
+function kernelmatrix_diag!(K::AbstractVector, κ::Kernel, x::AbstractVector)
     validate_inplace_dims(K, x)
     return map!(x -> κ(x, x), K, x)
 end
 
-function kerneldiagmatrix!(
+function kernelmatrix_diag!(
     K::AbstractVector, κ::Kernel, x::AbstractVector, y::AbstractVector
 )
     return map!(κ, x, y)
```
devmotion, line 69:
```suggestion
    return map!(κ, K, x, y)
```

test/matrix/kernelmatrix.jl
```diff
@@ -31,16 +31,29 @@ KernelFunctions.kappa(::ToySimpleKernel, d) = exp(-d / 2)
...
+            @test tmp_diag ≈ kernelmatrix_diag(k, x)
+
+            # Test deprecations
+            @test @test_deprecated kerneldiagmatrix(k, x) == kernelmatrix_diag(k, x)
+            @test @test_deprecated kerneldiagmatrix(k, x, x) == kernelmatrix_diag(k, x, x)
+
+            tmp_diag_dep = Vector{Float64}(undef, length(x))
+            @test_deprecated kerneldiagmatrix!(tmp_diag_dep, k, x)
+            @test tmp_diag_dep == kernelmatrix_diag(k, x)
+            @test_deprecated kerneldiagmatrix!(tmp_diag_dep, k, x, x)
```
devmotion, line 55: It would be cleaner/safer to reset `tmp_diag_dep`. Otherwise the test of the return value might pass even if the function is implemented incorrectly:

```suggestion

            tmp_diag_dep = Vector{Float64}(undef, length(x))            
            @test_deprecated kerneldiagmatrix!(tmp_diag_dep, k, x, x)
```

BTW it seems we only test `kernelmatrix_diag!(out, k, x)` but not `kernelmatrix_diag!(out, k, x, y)` above in the non-deprecated section?

test/matrix/kernelmatrix.jl
```diff
@@ -111,16 +127,32 @@ KernelFunctions.kappa(::ToySimpleKernel, d) = exp(-d / 2)
...
+            @test @test_deprecated kerneldiagmatrix(k, X; obsdim=obsdim) ==
+                                   kernelmatrix_diag(k, X; obsdim=obsdim)
+            @test @test_deprecated kerneldiagmatrix(k, X, X; obsdim=obsdim) ==
+                                   kernelmatrix_diag(k, X, X; obsdim=obsdim)
+
+            tmp_diag_dep = Vector{Float64}(undef, length(x))
+            @test_deprecated kerneldiagmatrix!(tmp_diag_dep, k, X; obsdim=obsdim)
+            @test tmp_diag_dep == kernelmatrix_diag(k, X; obsdim=obsdim)
+            @test_deprecated kerneldiagmatrix!(tmp_diag_dep, k, X, X; obsdim=obsdim)
+            @test tmp_diag_dep == kernelmatrix_diag(k, X, X; obsdim=obsdim)
```
devmotion, line 155: Same here, it would be better to reset `tmp_diag_dep` and `kernelmatrix_diag!(out, k, X, Y; obsdim)` seems to be missing in the non-deprecated tests.

src/matrix/kernelmatrix.jl
st--: Ah, thanks for spotting that, I was just trying to look through why the tests were failing before. Turns out my added tests for in-place array actually uncovered a bug in the preexisting code :grin:

test/matrix/kernelmatrix.jl
st--: I'll add that as well - though I was already wondering about what the use-case for kernelmatrix_diag(k, x, y) is! it seems like code that needs maintaining without any benefit I can see from it - what would you use it for?

test/matrix/kernelmatrix.jl
```diff
@@ -31,16 +31,29 @@ KernelFunctions.kappa(::ToySimpleKernel, d) = exp(-d / 2)
...
+            @test tmp ≈ kernelmatrix(k, x, y)
 
             tmp_square = Matrix{Float64}(undef, length(x), length(x))
             @test kernelmatrix!(tmp_square, k, x) ≈ kernelmatrix(k, x)
+            @test tmp_square ≈ kernelmatrix(k, x)
 
             tmp_diag = Vector{Float64}(undef, length(x))
-            @test kerneldiagmatrix!(tmp_diag, k, x) ≈ kerneldiagmatrix(k, x)
+            @test kernelmatrix_diag!(tmp_diag, k, x) ≈ kernelmatrix_diag(k, x)
+            @test tmp_diag ≈ kernelmatrix_diag(k, x)
```
st--, line 46:
```suggestion
            @test tmp_diag ≈ kernelmatrix_diag(k, x)
            tmp_diag = Vector{Float64}(undef, length(x))
            @test kernelmatrix_diag!(tmp_diag, k, x, x) ≈ kernelmatrix_diag(k, x, x)
            @test tmp_diag ≈ kernelmatrix_diag(k, x, x)
```

test/matrix/kernelmatrix.jl
```diff
@@ -111,16 +127,32 @@ KernelFunctions.kappa(::ToySimpleKernel, d) = exp(-d / 2)
...
 
             tmp_square = Matrix{Float64}(undef, length(x), length(x))
             @test kernelmatrix(k, x) ≈ kernelmatrix!(tmp_square, k, X; obsdim=obsdim)
+            @test kernelmatrix(k, x) ≈ tmp_square
 
             tmp_diag = Vector{Float64}(undef, length(x))
-            @test kerneldiagmatrix(k, x) ≈ kerneldiagmatrix!(tmp_diag, k, X; obsdim=obsdim)
+            @test kernelmatrix_diag(k, x) ≈
+                  kernelmatrix_diag!(tmp_diag, k, X; obsdim=obsdim)
+            @test kernelmatrix_diag(k, x) ≈ tmp_diag
```
st--, line 143:
```suggestion
            @test kernelmatrix_diag(k, x) ≈ tmp_diag
            tmp_diag = Vector{Float64}(undef, length(x))
            @test kernelmatrix_diag!(tmp_diag, k, X, X; obsdim=obsdim) ≈
                  kernelmatrix_diag(k, x, x)
            @test tmp_diag ≈ kernelmatrix_diag(k, x, x)
```

test/matrix/kernelmatrix.jl
```diff
@@ -111,16 +127,32 @@ KernelFunctions.kappa(::ToySimpleKernel, d) = exp(-d / 2)
...
+            # Test deprecations
+            @test @test_deprecated kerneldiagmatrix(k, X; obsdim=obsdim) ==
+                                   kernelmatrix_diag(k, X; obsdim=obsdim)
+            @test @test_deprecated kerneldiagmatrix(k, X, X; obsdim=obsdim) ==
+                                   kernelmatrix_diag(k, X, X; obsdim=obsdim)
+
+            tmp_diag_dep = Vector{Float64}(undef, length(x))
+            @test_deprecated kerneldiagmatrix!(tmp_diag_dep, k, X; obsdim=obsdim)
+            @test tmp_diag_dep == kernelmatrix_diag(k, X; obsdim=obsdim)
+            @test_deprecated kerneldiagmatrix!(tmp_diag_dep, k, X, X; obsdim=obsdim)
```
st--, line 154:
```suggestion
            tmp_diag_dep = Vector{Float64}(undef, length(x))
            @test_deprecated kerneldiagmatrix!(tmp_diag_dep, k, X, X; obsdim=obsdim)
```

st--:
While we're looking at this: what's the use-case for kernelmatrix_diag(k, x, y) (with x != y but same shape)? I can't think of any place I'd actually use that code, so is it worth incurring the maintenance cost (like we've had now)? What for or where would you use it?

devmotion:
I haven't used `kerneldiagmat` in any downstream code but I would have assumed that the two argument version is more useful. For e.g. translation-invariant kernels `kerneldiagmat(k, x)` is just trivial and returns a constant vector (usually of ones for unscaled kernels) whereas `kerneldiagmat(k, x, y) = [k(xi, yi) for (xi, yi) in zip(x, y)]` is a more interesting non-constant vector.

st--:
> I haven't used `kerneldiagmat` in any downstream code but I would have assumed that the two argument version is more useful. For e.g. translation-invariant kernels `kerneldiagmat(k, x)` is just trivial and returns a constant vector (usually of ones for unscaled kernels) whereas `kerneldiagmat(k, x, y) = [k(xi, yi) for (xi, yi) in zip(x, y)]` is a more interesting non-constant vector.

In Gaussian process models it's exactly the `kernelmatrix_diag(k, x)` that's needed - it may be trivial for stationary kernels (but also has the advantage it's super fast to compute), but it's also common to use non-stationary kernels (e.g. linear, arc-cosine, change-points). I've not come across a use-case for `kernelmatrix_diag(k, x, y)` with y!=x which is why I asked. If we can't think of any use-cases, would it not be better to save the maintenance cost? What's the value of keeping around functions just because we can write them if we have no use for it?

st--:
The one failing CI run seems to be some blip (if anyone can explain why that would happen intermittently, I'd be curious to find out). So I'm going to go ahead and merge it as is. I've moved the open question to a separate issue (#250).

theogf:
> While we're looking at this: what's the use-case for kernelmatrix_diag(k, x, y) (with x != y but same shape)? I can't think of any place I'd actually use that code, so is it worth incurring the maintenance cost (like we've had now)? What for or where would you use it?

@willtebbutt motivated this as he needs it for Stheno

st--:
I've bumped the version. Is there anything else that needs doing to release?

@willtebbutt can you point me to the concrete use? would be curious to understand what you do with it.

theogf:
> I've bumped the version. Is there anything else that needs doing to release?

Yes, I just did. You can check it out all you need is to write a comment to the merging commit with `@JuliaRegistrator register`

st--:
Ah, neat. The docs suggest that given TagBot is installed in this repo's workflow it should automatically add a tag for the new release. How long does it usually take for that to happen? I can't see a new tag yet..

theogf:
So first it needs to be registered in the Registry and that takes around 20 min (there is a cycle of automerge every 15 min or so) and then TagBot will automatically create a new tag with closes issues / pull request merged, that we can then edit

Issue #218, opened 2021-01-09
Consolidate mathematical description of kernels between docs and docstrings


st--:
@willtebbutt wrote on #213:
> I'm concerned is that we're duplicating documentation a bit at the minute by having both docstrings and having the entire `Kernel`s section written out manually. This creates an opportunity for the docs to decouple from the docstrings and increases the maintenance burden. I wonder whether it would make sense, therefore, to consolidate this documentation -- if we were to move the concise mathematical description of the kernels from the docs to the docstrings, and just move /duplicate bits of the `API` section in the `Kernel Functions` section we could get the best of both worlds?

_Originally posted by @willtebbutt in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/213#issuecomment-757064536_

devmotion:
This is basically what I had in mind with the suggestion in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/216 as well.

devmotion:
Closed by https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/236.

Issue #216, opened 2021-01-08
Split API docs into separate pages, remove redundancy with kernels/transforms


st--:
Maybe one could also just move the list of transforms here (the full API section is a bit overwhelming to be honest). But if we want to do this, it might best to change it in a separate PR and maybe for some other sections as well.

_Originally posted by @devmotion in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/213#discussion_r553648569_

devmotion:
Fixed by https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/236.

PR #236, opened 2021-01-19
Use only docstrings in documentation + fix them

devmotion:
Addresses https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/218 and https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/216 and uses the docstrings in the documentation of kernel functions and transformation.

The docstrings are updated and synced with the documentation, a bunch of errors are fixed (e.g., incorrect Matérn kernel functions and rational quadratic kernels), formatting of most docstrings of kernels and transformations is fixed (basically everything apart from the Wiener kernel and spectral mixture kernel, I was too tired to fix the mess there), doctests are added to most transformations, and incorrect explanations and outdated code of the transformations are fixed.

### DIFF
--- a/docs/src/kernels.md
+++ b/docs/src/kernels.md
@@ -2,11 +2,18 @@
   CurrentModule = KernelFunctions
 ```
 
-# Base Kernels
+# Kernel Functions
+
+```@contents
+Pages = ["kernels.md"]
+Depth = 3
+```
+
+## Base Kernels
 
 These are the basic kernels without any transformation of the data. They are the building blocks of KernelFunctions.
 
-## Constant Kernels
+### Constant Kernels
 
 ```@docs
 ZeroKernel
@@ -15,13 +22,13 @@ WhiteKernel
 EyeKernel
 ```
 
-## Cosine Kernel
+### Cosine Kernel
 
 ```@docs
 CosineKernel
 ```
 
-## Exponential Kernels
+### Exponential Kernels
 
 ```@docs
 ExponentialKernel
@@ -33,25 +40,25 @@ RBFKernel
 GammaExponentialKernel
 ```
 
-## Exponentiated Kernel
+### Exponentiated Kernel
 
 ```@docs
 ExponentiatedKernel
 ```
 
-## Fractional Brownian Motion Kernel
+### Fractional Brownian Motion Kernel
 
 ```@docs
 FBMKernel
 ```
 
-## Gabor Kernel
+### Gabor Kernel
 
 ```@docs
 GaborKernel
 ```
 
-## Matérn Kernels
+### Matérn Kernels
 
 ```@docs
 MaternKernel
@@ -60,53 +67,53 @@ Matern32Kernel
 Matern52Kernel
 ```
 
-## Neural Network Kernel
+### Neural Network Kernel
 
 ```@docs
 NeuralNetworkKernel
 ```
 
-## Periodic Kernel
+### Periodic Kernel
 
 ```@docs
 PeriodicKernel
 PeriodicKernel(::DataType, ::Int)
 ```
 
-## Piecewise Polynomial Kernel
+### Piecewise Polynomial Kernel
 
 ```@docs
 PiecewisePolynomialKernel
 ```
 
-## Polynomial Kernels
+### Polynomial Kernels
 
 ```@docs
 LinearKernel
 PolynomialKernel
 ```
 
-## Rational Quadratic Kernels
+### Rational Quadratic Kernels
 
 ```@docs
 RationalQuadraticKernel
 GammaRationalQuadraticKernel
 ```
 
-## Spectral Mixture Kernels
+### Spectral Mixture Kernels
 
 ```@docs
 spectral_mixture_kernel
 spectral_mixture_product_kernel
 ```
 
-## Wiener Kernel
+### Wiener Kernel
 
 ```@docs
 WienerKernel
 ```
 
-# Composite Kernels
+## Composite Kernels
 
 The modular design of KernelFunctions.jl uses [base kernels](@ref) as building blocks for
 more complex kernels. These composite kernels can be created by adding
@@ -124,7 +131,7 @@ KernelProduct
 TensorProduct
 ```
 
-# Multi-output Kernels
+## Multi-output Kernels
 
 ```@docs
 MOKernel
--- a/docs/src/kernels.md
+++ b/docs/src/kernels.md
@@ -86,7 +86,7 @@ LinearKernel
 PolynomialKernel
 ```
 
-## Rational Quadratic
+## Rational Quadratic Kernels
 
 ```@docs
 RationalQuadraticKernel
--- a/docs/src/kernels.md
+++ b/docs/src/kernels.md
@@ -108,6 +108,11 @@ WienerKernel
 
 # Composite Kernels
 
+The modular design of KernelFunctions.jl uses [base kernels](@ref) as building blocks for
+more complex kernels. These composite kernels can be created by adding
+[input transforms](@ref) such as scaling the inputs with an inverse lengthscale, by scaling the
+variance of a kernel, or by summing or multiplying different kernels.
+
 ```@docs
 TransformedKernel
 transform(::Kernel, ::Transform)
--- a/docs/src/kernels.md
+++ b/docs/src/kernels.md
@@ -6,270 +6,123 @@
 
 These are the basic kernels without any transformation of the data. They are the building blocks of KernelFunctions.
 
-
 ## Constant Kernels
 
-### Constant Kernel
-
-The [`ConstantKernel`](@ref) is defined as
-
-```math
-  k(x,x';c) = c,
-```
-
-where $c \geq 0$.
-
-### White Kernel
-
-The [`WhiteKernel`](@ref) is defined as
-
-```math
-  k(x,x') = \delta(x-x').
-```
-
-### Zero Kernel
-
-The [`ZeroKernel`](@ref) is defined as
-
-```math
-  k(x,x') = 0.
+```@docs
+ZeroKernel
+ConstantKernel
+WhiteKernel
+EyeKernel
 ```
 
 ## Cosine Kernel
 
-The [`CosineKernel`](@ref) is defined as
-
-```math
-  k(x, x') = \cos(\pi \|x-x'\|).
+```@docs
+CosineKernel
 ```
 
 ## Exponential Kernels
 
-### Exponential Kernel
-
-The [`ExponentialKernel`](@ref) is defined as
-```math
-  k(x,x') = \exp\left(-|x-x'|\right).
+```@docs
+ExponentialKernel
+LaplacianKernel
+SqExponentialKernel
+SEKernel
+GaussianKernel
+RBFKernel
+GammaExponentialKernel
 ```
-
-### Square Exponential Kernel
-
-The [`SqExponentialKernel`](@ref) is defined as
-```math
-  k(x,x') = \exp\left(-\|x-x'\|^2\right).
-```
-
-### Gamma Exponential Kernel
-
-The [`GammaExponentialKernel`](@ref) is defined as
-
-```math
-  k(x,x';\gamma) = \exp\left(-\|x-x'\|^{2\gamma}\right),
-```
-where $\gamma > 0$.
 
 ## Exponentiated Kernel
 
-The [`ExponentiatedKernel`](@ref) is defined as
-
-```math
-  k(x,x') = \exp\left(\langle x,x'\rangle\right).
+```@docs
+ExponentiatedKernel
 ```
 
-## Fractional Brownian Motion
+## Fractional Brownian Motion Kernel
 
-The [`FBMKernel`](@ref) is defined as
-
-```math
-  k(x,x';h) =  \frac{|x|^{2h} + |x'|^{2h} - |x-x'|^{2h}}{2},
+```@docs
+FBMKernel
 ```
-
-where $h$ is the [Hurst index](https://en.wikipedia.org/wiki/Hurst_exponent#Generalized_exponent) and $0 < h < 1$.
 
 ## Gabor Kernel
 
-The [`GaborKernel`](@ref) is defined as
-
-```math
-  k(x,x'; l,p) = \exp\left(-\cos\left(\pi \sum_i \frac{x_i - x'_i}{p_i}\right)\sum_i \frac{(x_i - x'_i)^2}{l_i^2}\right),
+```@docs
+GaborKernel
 ```
-where $l_i > 0$ is the lengthscale and $p_i > 0$ is the period.
 
 ## Matérn Kernels
 
-### General Matérn Kernel
-
-The [`MaternKernel`](@ref) is defined as
-
-```math
-  k(x,x';\nu) = \frac{2^{1-\nu}}{\Gamma(\nu)}\left(\sqrt{2\nu}|x-x'|\right)K_\nu\left(\sqrt{2\nu}|x-x'|\right),
-```
-
-where $\nu > 0$.
-
-### Matérn 1/2 Kernel
-
-The Matérn 1/2 kernel is defined as
-```math
-  k(x,x') = \exp\left(-|x-x'|\right),
-```
-equivalent to the Exponential kernel. `Matern12Kernel` is an alias for [`ExponentialKernel`](@ref).
-
-### Matérn 3/2 Kernel
-
-The [`Matern32Kernel`](@ref) is defined as
-
-```math
-  k(x,x') = \left(1+\sqrt{3}|x-x'|\right)\exp\left(\sqrt{3}|x-x'|\right).
-```
-
-### Matérn 5/2 Kernel
-
-The [`Matern52Kernel`](@ref) is defined as
-
-```math
-  k(x,x') = \left(1+\sqrt{5}|x-x'|+\frac{5}{2}\|x-x'\|^2\right)\exp\left(\sqrt{5}|x-x'|\right).
+```@docs
+MaternKernel
+Matern12Kernel
+Matern32Kernel
+Matern52Kernel
 ```
 
 ## Neural Network Kernel
 
-The [`NeuralNetworkKernel`](@ref) (as in the kernel for an infinitely wide neural network interpreted as a Gaussian process) is defined as
-
-```math
-  k(x, x') = \arcsin\left(\frac{\langle x, x'\rangle}{\sqrt{(1+\langle x, x\rangle)(1+\langle x',x'\rangle)}}\right).
+```@docs
+NeuralNetworkKernel
 ```
 
 ## Periodic Kernel
 
-The [`PeriodicKernel`](@ref) is defined as
-
-```math
-  k(x,x';r) = \exp\left(-0.5 \sum_i (\sin (\pi(x_i - x'_i))/r_i)^2\right),
+```@docs
+PeriodicKernel
+PeriodicKernel(::DataType, ::Int)
 ```
-
-where $r$ has the same dimension as $x$ and $r_i > 0$.
 
 ## Piecewise Polynomial Kernel
 
-The [`PiecewisePolynomialKernel`](@ref) of degree $v \in \{0,1,2,3\}$ is defined for
-inputs $x, x' \in \mathbb{R}^d$ of dimension $d$ as
-```math
-k(x, x'; v) = \max(1 - \|x - x'\|, 0)^{\alpha} f_{v,d}(\|x - x'\|),
+```@docs
+PiecewisePolynomialKernel
 ```
-where $\alpha = \lfloor \frac{d}{2}\rfloor + 2v + 1$, and $f_{v,d}$ are polynomials of
-degree $v$ given by
-```math
-\begin{aligned}
-f_{0,d}(r) &= 1, \\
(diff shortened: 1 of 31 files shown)


docs/src/kernels.md
```diff
@@ -6,270 +6,123 @@
...
-
-The [`PolynomialKernel`](@ref) is defined as
-
-```math
-  k(x,x';c,d) = \left(\langle x,x'\rangle + c\right)^d,
-```
-
-where $c \geq 0$ and $d \in \mathbb{N}$.
-
 ## Rational Quadratic
```
devmotion, line 89:
```suggestion
## Rational Quadratic Kernels
```

src/basekernels/exponential.jl
```diff
@@ -66,35 +74,39 @@ Base.show(io::IO, ::ExponentialKernel) = print(io, "Exponential Kernel")
...
+
+[^RW]: C. E. Rasmussen & C. K. I. Williams (2006). Gaussian Processes for Machine Learning.
 """
 struct GammaExponentialKernel{Tγ<:Real} <: SimpleKernel
     γ::Vector{Tγ}
-    function GammaExponentialKernel(; gamma::T=2.0, γ::T=gamma) where {T<:Real}
-        @check_args(GammaExponentialKernel, γ, γ >= zero(T), "γ > 0")
-        return new{T}([γ])
+    function GammaExponentialKernel(; gamma::Real=2.0, γ::Real=gamma)
+        @check_args(GammaExponentialKernel, γ, zero(γ) < γ ≤ 2, "γ ∈ (0, 2]")
```
devmotion, line 108: Otherwise the kernel is not positive definite (the case 0 is not interesting and was excluded in the docstring but not in the check itself).

src/basekernels/fbm.jl
```diff
@@ -1,19 +1,22 @@
...
+k(x, x'; h) =  \\frac{\\|x\\|_2^{2h} + \\|x'\\|_2^{2h} - \\|x - x'\\|^{2h}}{2}.
+```
 """
 struct FBMKernel{T<:Real} <: Kernel
     h::Vector{T}
-    function FBMKernel(; h::T=0.5) where {T<:Real}
-        @assert 0.0 <= h <= 1.0 "FBMKernel: Given Hurst index h is invalid."
-        return new{T}([h])
+    function FBMKernel(; h::Real=0.5)
+        @check_args(FBMKernel, h, zero(h) ≤ h ≤ one(h), "h ∈ [0, 1]")
```
devmotion, line 18: Also here `h=0` is uninteresting.

src/basekernels/matern.jl
```diff
@@ -38,14 +47,20 @@ Base.show(io::IO, κ::MaternKernel) = print(io, "Matern Kernel (ν = ", first(κ
...
-The matern 3/2 kernel is a Mercer kernel given by the formula:
-```
-    κ(x,y) = (1+√(3)‖x-y‖)exp(-√(3)‖x-y‖)
+Matérn kernel of order ``3/2``.
+
+# Definition
+
+For inputs ``x, x' \\in \\mathbb{R}^d``, the Matérn kernel of order ``3/2`` is given by
+```math
+k(x, x') = \\big(1 + \\sqrt{3} \\|x - x'\\|_2 \\big) \\exp\\big(- \\sqrt{3}\\|x - x'\\|_2\\big).
```
devmotion, line 56: The current documentation is incorrect, it is missing the minus sign inside the exponential.

src/basekernels/matern.jl
```diff
@@ -54,14 +69,21 @@ Base.show(io::IO, ::Matern32Kernel) = print(io, "Matern 3/2 Kernel")
...
-```
-    κ(x,y) = (1+√(5)‖x-y‖ + 5/3‖x-y‖^2)exp(-√(5)‖x-y‖)
+Matérn kernel of order ``5/2``.
+
+# Definition
+
+For inputs ``x, x' \\in \\mathbb{R}^d``, the Matérn kernel of order ``5/2`` is given by
+```math
+k(x, x') = \\bigg(1 + \\sqrt{5} \\|x - x'\\|_2 + \\frac{5}{3}\\|x - x'\\|_2^2\\bigg)
+           \\exp\\big(- \\sqrt{5}\\|x - x'\\|_2\\big).
```
devmotion, line 79: Same here, the current docs are missing the minus sign.

src/basekernels/rationalquad.jl
```diff
@@ -1,12 +1,17 @@
...
-```
-    κ(x, y) = (1 + ||x − y||² / (2α))^(-α)
+Rational-quadratic kernel with shape parameter `α`.
+
+# Definition
+
+For inputs ``x, x' \\in \\mathbb{R}^d``, the rational-quadratic kernel with shape parameter
+``\\alpha > 0`` is defined as
+```math
+k(x, x'; \\alpha) = \\bigg(1 + \\frac{\\|x - x'\\|_2^2}{2\\alpha}\\bigg)^{-\\alpha}.
```
devmotion, line 11: Currently, docstring and documentation are not consistent - the documentation is missing a factor of 2 in the denominator. Since the docstring and code were consistent, I only fixed the format of the docstring. IIRC you changed this a while back @willtebbutt?

src/basekernels/rationalquad.jl
```diff
@@ -29,33 +34,39 @@ function Base.show(io::IO, κ::RationalQuadraticKernel)
...
-```
-    κ(x, y) = (1 + ||x−y||^γ / α)^(-α)
+γ-rational-quadratic kernel with shape parameters `α` and `γ`.
+
+# Definition
+
+For inputs ``x, x' \\in \\mathbb{R}^d``, the γ-rational-quadratic kernel with shape
+parameters ``\\alpha > 0`` and ``\\gamma \\in (0, 2]`` is defined as
+```math
+k(x, x'; \\alpha, \\gamma) = \\bigg(1 + \\frac{\\|x - x'\\|_2^{\\gamma}}{2\\alpha}\\bigg)^{-\\alpha}.
```
devmotion, line 46: Also here's an inconsistency between docstring and documentation currently - the documentation includes an additional factor of 2 in the exponent. Since `RationalQuadraticKernel` and `GammaRationalQuadraticKernel` are grouped together in the documentation and refer to each other in the docstring, I noticed an inconsistency in the implementation: `RationalQuadraticKernel` includes an additional factor of 2 in the denominator whereas `GammaRationalQuadraticKernel` doesn't. Is there a reason for this behaviour or is this a bug, @willtebbutt?

With the additional factor of 2 in the denominator, the more general kernel would yield the same kernel with its default values whereas currently the inputs have to be scaled. On the other hand, one would lose the property that the `GammaExponentialKernel` is the limit of the `GammaRationalQuadraticKernel` as alpha goes to infinity. But the name "GammaRationalQuadraticKernel" seems to indicate that it is more closely related to the `RationalQuadraticKernel`...

Maybe
- one should rename `GammaRationalQuadraticKernel` to `GammaRationalKernel` now that the factor 2 in the exponent is removed
- add an explicit `RationalKernel` for the case `gamma=1` of the `GammaRationalQuadraticKernel`, which yields an `ExponentialKernel` as alpha goes to infinity
- change the default value of gamma of the `GammaRationalKernel` to 1, such that it is a generalization of `RationalKernel` instead of `RationalQuadraticKernel`, similar to the relation of `ExponentialKernel` and `GammaExponentialKernel`

Then everything would be consistent with the setup in the limit for alpha=\infty.

What do you think, @willtebbutt?

src/basekernels/rationalquad.jl
```diff
@@ -29,33 +34,39 @@ function Base.show(io::IO, κ::RationalQuadraticKernel)
...
+        return new{typeof(α),typeof(γ)}([α], [γ])
     end
 end
 
 @functor GammaRationalQuadraticKernel
 
-function kappa(κ::GammaRationalQuadraticKernel, d²::Real)
-    return (one(d²) + d²^(first(κ.γ) / 2) / first(κ.α))^(-first(κ.α))
+function kappa(κ::GammaRationalQuadraticKernel, d::Real)
+    return (one(d) + d^first(κ.γ) / (2 * first(κ.α)))^(-first(κ.α))
```
devmotion, line 66: Is there a reason for using the `SqEuclidean` distance and dividing gamma by 2 instead of this approach? (For a discussion of the 2 here, see above)

test/basekernels/exponential.jl
```diff
@@ -44,7 +44,7 @@
         @test repr(k) == "Gamma Exponential Kernel (γ = $(γ))"
         @test KernelFunctions.iskroncompatible(k) == true
 
-        test_ADs(γ -> GammaExponentialKernel(; gamma=first(γ)), [γ])
+        test_ADs(γ -> GammaExponentialKernel(; gamma=first(γ)), [1 + 0.5 * rand()])
```
devmotion, line 47: Ideally, one should use a reparameterization `R -> (0,2]` here, e.g. `f(x) = 2 * logistic`. With the current boundary point `gamma = 2` finite differencing errors.

src/basekernels/rationalquad.jl
devmotion: I kept the current definition for now since these changes should not be part of the PR.

docs/src/transform.md
```diff
@@ -1,9 +1,31 @@
 # Input Transforms
 
-`Transform` is the object that takes care of transforming the input data before distances are being computed. It can be as standard as `IdentityTransform` returning the same input, or multiplying the data by a scalar with `ScaleTransform` or by a vector with `ARDTransform`.
-There is a more general `Transform`: `FunctionTransform` that uses a function and applies it on each vector via `mapslices`.
-You can also create a pipeline of `Transform` via `TransformChain`. For example, `LowRankTransform(rand(10,5))∘ScaleTransform(2.0)`.
+[`Transform`](@ref) is the object that takes care of transforming the input data before distances are being computed.
```
willtebbutt, line 3:
```suggestion
[`Transform`](@ref)s are designed to change input data before passing it on to a kernel object.
```

docs/src/transform.md
```diff
@@ -1,9 +1,31 @@
...
+each input.
 
-Check the full list of provided transforms on the [API page](@ref Transforms).
+You can also create a pipeline of [`Transform`](@ref)s via [`ChainTransform`](@ref), e.g.,
+```julia
+LowRankTransform(rand(10, 5)) ∘ ScaleTransform(2.0)
+```
+
+A transformation `t` can be applied to a vector `v` with `map`.
+
```
willtebbutt, line 17: I think we should at least mention the `transform` function and `TransformedKernel` type in this function, so as to explain to users how to actually _use_ transforms.

docs/src/kernels.md
```diff
@@ -6,270 +6,123 @@
...
-  \delta(x, x') & i = -1\\
-  \min(x,x') & i = 0\\
-  \frac{\min(x,x')^{2i+1}}{a_i} + b_i \min(x,x')^{i+1}|x-x'|r_i(x,x') & i\geq 1
-\end{array}\right.,
+```@docs
+WienerKernel
 ```
-where $i\in\{-1,0,1,2,3\}$ and coefficients $a_i$, $b_i$ are fixed and residuals $r_i$ are defined in the code.
 
 # Composite Kernels
```
willtebbutt, line 110: Could you add a sentence or two here explaining what we mean by `Composite Kernels` and why we have them etc?

src/kernels/scaledkernel.jl
```diff
@@ -1,7 +1,15 @@
 """
-    ScaledKernel(k::Kernel, σ²::Real)
+    ScaledKernel(k::Kernel, σ²::Real=1.0)
```
willtebbutt, line 2: There's an inconsistency here between the name of the argument `k`, and what its called in the definition below: ``\\widetilde{k}``. Could you maybe swap some of the names around? Perhaps ``\\widetilde{k}`` should be the derived kernel?

src/kernels/transformedkernel.jl
```diff
@@ -1,8 +1,15 @@
 """
-    TransformedKernel(k::Kernel,t::Transform)
+    TransformedKernel(k::Kernel, t::Transform)
 
-Return a kernel where inputs are pretransformed by `t` : `k(t(x),t(x'))`
-Can also be called via [`transform`](@ref) : `transform(k, t)`
+Kernel derived from `k` for which inputs are transformed via a [`Transform`](@ref) `t`.
+
+# Definition
```
willtebbutt, line 6: Could you add a note suggesting that users should prefer to use `transform`, rather than `TransformedKernel` directly?

src/kernels/transformedkernel.jl
```diff
@@ -1,8 +1,15 @@
 """
-    TransformedKernel(k::Kernel,t::Transform)
+    TransformedKernel(k::Kernel, t::Transform)
```
willtebbutt, line 2: Same here regarding the inconsistency between the name in the signature and the name in the maths.

willtebbutt (changes requested):
This is really fantastic work. Just a few things to fix.

src/kernels/transformedkernel.jl
```diff
@@ -1,8 +1,15 @@
...
 
-Return a kernel where inputs are pretransformed by `t` : `k(t(x),t(x'))`
-Can also be called via [`transform`](@ref) : `transform(k, t)`
+Kernel derived from `k` for which inputs are transformed via a [`Transform`](@ref) `t`.
+
+# Definition
+
+For inputs ``x, x'``, the transformed kernel derived from kernel ``\\widetilde{k}`` by
```
devmotion, line 8:
```suggestion
For inputs ``x, x'``, the transformed kernel ``\\widetilde{k}`` derived from kernel ``k`` by
```

src/kernels/transformedkernel.jl
```diff
@@ -1,8 +1,15 @@
...
+input transformation ``t`` is defined as
+```math
+k(x, x'; t, \\widetilde{k}) = \\widetilde{k}(t(x), t(x')).
```
devmotion, line 11:
```suggestion
\\widetilde{k}(x, x'; k, t) = k\\big(t(x), t(x')\\big).
```

src/kernels/transformedkernel.jl
```diff
@@ -1,8 +1,15 @@
 """
-    TransformedKernel(k::Kernel,t::Transform)
+    TransformedKernel(k::Kernel, t::Transform)
 
-Return a kernel where inputs are pretransformed by `t` : `k(t(x),t(x'))`
-Can also be called via [`transform`](@ref) : `transform(k, t)`
+Kernel derived from `k` for which inputs are transformed via a [`Transform`](@ref) `t`.
+
```
devmotion, line 5:
```suggestion

It is preferred to use [`transform`](@ref) instead of constructing `TransformedKernel`
directly.

```

src/kernels/scaledkernel.jl
```diff
@@ -1,7 +1,15 @@
...
 
-Return a kernel premultiplied by the variance `σ²` : `σ² k(x,x')`
+Scaled kernel derived from `k` by multiplication with variance `σ²`.
+
+# Definition
+
+For inputs ``x, x'``, the scaled kernel derived from kernel ``\\widetilde{k}`` by
+multiplication with variance ``\\sigma^2 > 0`` is defined as
+```math
+k(x, x'; \\sigma^2, \\widetilde{k}) = \\sigma^2 \\widetilde{k}(x, x').
```
devmotion, line 11:
```suggestion
For inputs ``x, x'``, the scaled kernel ``\\widetilde{k}`` derived from kernel ``k`` by
multiplication with variance ``\\sigma^2 > 0`` is defined as
```math
\\widetilde{k}(x, x'; k, \\sigma^2) = \\sigma^2 k(x, x').
```

docs/src/kernels.md
```diff
@@ -6,270 +6,123 @@
...
-```
-where $\widetilde{k}$ is another kernel and $f$ is an arbitrary mapping.
-
-### Scaled Kernel
-
-The [`ScaledKernel`](@ref) is defined as
-
-```math
-  k(x,x';\sigma^2,\widetilde{k}) = \sigma^2\widetilde{k}(x,x') ,
+```@docs
```
devmotion, line 111:
```suggestion
The modular design of KernelFunctions.jl uses [base kernels](@ref) as building blocks for
more complex kernels. These composite kernels can be created by adding
[input transforms](@ref) such as scaling the inputs with an inverse lengthscale, by scaling the
variance of a kernel, or by summing or multiplying different kernels.

```@docs
```

docs/src/transform.md
```diff
@@ -1,9 +1,31 @@
...
+There is a more general [`FunctionTransform`](@ref) that uses a function and applies it to
+each input.
 
-Check the full list of provided transforms on the [API page](@ref Transforms).
+You can also create a pipeline of [`Transform`](@ref)s via [`ChainTransform`](@ref), e.g.,
+```julia
+LowRankTransform(rand(10, 5)) ∘ ScaleTransform(2.0)
+```
+
+A transformation `t` can be applied to a vector `v` with `map`.
```
devmotion, line 16:
```suggestion

A transformation `t` can be applied to a single input `x` with `t(x)` and to multiple inputs
`xs` with `map(t, xs)`.

Kernels can be coupled with input transformations with
[`transform`](@ref). It falls back to creating a [`TransformedKernel`](@ref) but allows more
optimized implementations for specific kernels and transformations.
```

docs/src/transform.md
devmotion: I added some additional sentences about `transform` and `TransformedKernel`.

docs/src/kernels.md
devmotion: I added a short paragraph for now.

src/kernels/scaledkernel.jl
devmotion: I swapped it in the mathematical notation.

src/kernels/transformedkernel.jl
devmotion: I added a short statement.

src/kernels/transformedkernel.jl
devmotion: I changed it.

docs/src/kernels.md
willtebbutt: This is good, but as currently phrased it sounds like this list is somewhat final, and unlikely to be extended. I think it would be better to phrase it in a manner that suggests that we've got a variety of things implemented, but you can always add more. Perhaps something like:

```
There are a variety of composite kernels implemented, including those which
[transform the inputs](@ref input_transforms) to a wrapped kernel to implement length scales, scale the variance of a kernel, and sum or multiply collections of kernels together.
```

Something like that perhaps?

devmotion: Sure, sounds good :+1: I was not very creative and did not think too much about the wording. I'll update it according to your suggestions!

willtebbutt (approved):
Thanks for addressing my points. Now LGTM.

st--:
The downside of this is that while the maths may look nice on the rendered docs online, the REPL help is now rather less helpful compared to the previous use of unicode maths in the docstrings, e.g.:
```
  MaternKernel(; ν::Real=1.5)

  Matérn kernel of order ν.

  Definition
  ≡≡≡≡≡≡≡≡≡≡≡≡

  For inputs x, x' \in \mathbb{R}^d, the Matérn kernel of order \nu > 0 is defined as

  k(x,x';\nu) = \frac{2^{1-\nu}}{\Gamma(\nu)}\big(\sqrt{2\nu}\|x-x'\|_2\big) K_\nu\big(\sqrt{2\nu}\|x-x'\|_2\big),

  where \Gamma is the Gamma function and K_{\nu} is the modified Bessel function of the second kind of order \nu.

  A Gaussian process with a Matérn kernel is \lceil \nu \rceil - 1-times differentiable in the mean-square sense.
```
How about keeping the unification (:+1:) but using unicode for maths as in the docstrings before?

devmotion:
I tried this recently in some other package but unfortunately unicode math does not work reliably with Documenter (or the javascript libraries). Greek letters are displayed in a different non-math font, brackets can not be adjusted properly, unicode supports only a limited number of sub- and superscripts etc. Other things such as `\frac` are just not possible to write with unicode characters only. So in any case one would end up with a mix of both LaTeX and unicode characters, which makes e.g. the differences of the fonts even more prominent and does not solve the problem of LaTeX syntax in docstrings.

devmotion:
Possibly also due to these reasons, LaTeX syntax in docstrings is common practice in other packages such as Distributions (see e.g. https://github.com/JuliaStats/Distributions.jl/blob/master/src/univariate/discrete/betabinomial.jl) and standard libraries (see e.g. https://github.com/JuliaLang/julia/blob/d6f99481b1d5cf399165be2eb3893c5bdbe90dcf/stdlib/LinearAlgebra/src/qr.jl#L19)

theogf:
I chime in definitely too late, but that was the reason for two versions of the docs (docstring vs docs.md), unicode readable for docstring and latex for documenter.

devmotion:
I definitely agree that it would be nice to have some more readable output in the REPL and I tried to use unicode but it does not work. Only the simplest formulas can be written with unicode only, so one basically always has to write them at least partly with LaTeX syntax. And even the simplest unicode formulas look terrible on the webpage (e.g. even subscripts and operators such as `exp`).

devmotion:
On a different note, it might also be problematic if the docstring contains characters that are only supported by very few fonts.

PR #239, opened 2021-01-19
Build documentation more strictly

devmotion:
This PR ensures that failures when building the documentation and in doctests actually cause a failing build. It contains some of the fixes in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/232 and takes inspiration from AbstractGPs.

### DIFF
--- a/test/Project.toml
+++ b/test/Project.toml
@@ -1,6 +1,7 @@
 [deps]
 AxisArrays = "39de3d68-74b9-583c-8d2d-e117c070f3a9"
 Distances = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
+Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 FiniteDifferences = "26cc04aa-876d-5657-8c51-4c34ba976000"
 Flux = "587475ba-b771-5e3f-ad9e-33799f191a9c"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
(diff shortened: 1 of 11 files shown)


willtebbutt:
This looks good. I wonder whether we shouldn't also run the doctests as part of the usual testing cycle -- i.e. when `test/runtests.jl` is run? This would give you more feedback during the development cycle.

devmotion:
I thought about it (e.g., they are included as a separate test set in AbstractGPs) but assumed that it would be sufficient if the build errors if they fail (which is not the case in AbstractGPs currently). I don't have a strong opinion though, I'm fine with adding it to the test pipeline as well.

willtebbutt:
I agree that it is sufficient for the sake of getting CI to complain when the doctests fail, I'm just thinking about the overall development cycle. If you're happy to add it to the test pipeline, that would be great.

devmotion:
OK, done.

test/Project.toml
```diff
@@ -17,6 +18,7 @@ Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
 [compat]
 AxisArrays = "0.4.3"
 Distances = "0.9, 0.10"
+Documenter = "0.25"
```
devmotion, line 21: JuliaFormatter is not compatible with Documenter 0.26

willtebbutt: Booooo

willtebbutt (approved):
I'm happy with this now. I'll leave it up to you to figure out what to do about the compat issues.

devmotion:
I opened a PR (https://github.com/domluna/JuliaFormatter.jl/pull/354) but for the tests it should not matter. The webpage is still built with the most recent version since it uses a different project environment.

Issue #238, opened 2021-01-19
Inconsistency in rational kernels


devmotion:
Since `RationalQuadraticKernel` and `GammaRationalQuadraticKernel` are grouped together in the documentation and refer to each other in the docstring, I noticed an inconsistency in the implementation: `RationalQuadraticKernel` includes an additional factor of 2 in the denominator whereas `GammaRationalQuadraticKernel` doesn't. Is there a reason for this behaviour or is this a bug, @willtebbutt?

With the additional factor of 2 in the denominator, the more general kernel would yield the same kernel with its default values whereas currently the inputs have to be scaled. On the other hand, one would lose the property that the `GammaExponentialKernel` is the limit of the `GammaRationalQuadraticKernel` as alpha goes to infinity. But the name "GammaRationalQuadraticKernel" seems to indicate that it is more closely related to the `RationalQuadraticKernel`...

Maybe
- one should rename `GammaRationalQuadraticKernel` to `GammaRationalKernel` now that the factor 2 in the exponent is removed
- add an explicit `RationalKernel` for the case `gamma=1` of the `GammaRationalQuadraticKernel`, which yields an `ExponentialKernel` as alpha goes to infinity
- change the default value of gamma of the `GammaRationalKernel` to 1, such that it is a generalization of `RationalKernel` instead of `RationalQuadraticKernel`, similar to the relation of `ExponentialKernel` and `GammaExponentialKernel`

Then everything would be consistent with the setup in the limit for alpha=\infty.

What do you think, @willtebbutt?

_Originally posted by @devmotion in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/236#discussion_r560083376_

PR #242, opened 2021-01-20
Add RationalKernel and rename GammaRationalQuadraticKernel

devmotion:
Implements the suggestion in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/238 to increase consistency. Not sure if it is the best solution to the problem though :smile: 

Fixes https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/238.


src/deprecated.jl
```diff
@@ -7,3 +7,6 @@
 @deprecate PiecewisePolynomialKernel{V}(A::AbstractMatrix{<:Real}) where {V} transform(
     PiecewisePolynomialKernel{V}(size(A, 1)), LinearTransform(cholesky(A).U)
 )
+
+# TODO: remove in next breaking release
+const GammaRationalQuadraticKernel = GammaRationalKernel
```
devmotion, line 12: I don't know how to deprecate a type (+ constructors) correctly.

devmotion: `@deprecate GammaRationalQuadraticKernel(; kwargs...) GammaRationalKernel(; kwargs...)` only deprecates the constructor.

theogf: If the constructors are depecrated it should be fine unless someone explicitly created functions with the `GammaRationalQuadraticKernel` (which I highly doubt)
I would just remove the type altogether, it's better to have a breaking release now than later

devmotion: Everything that dispatches on the type will be broken.

theogf: Yes that would be a breaking change, but again I don't believe anyone dispatch on `GammaRationalQuadraticKernel` :D
After a search on github I only found it mentionned here :
https://github.com/julbinb/julia-sub/blob/d21178ac0a8eb805e6abeea80aa7a276b68f0e67/reports/initial/1000.txt#L156
If we exclude MLKernels and KernelFunctions

st--:
The kernel is commonly called `RationalQuadratic` (or some variant thereof such as `RQ` or `RatQuad`), not `Rational` (e.g. [GPML book](http://www.gaussianprocess.org/gpml/chapters/RW4.pdf), [Scikit-Learn](https://scikit-learn.org/stable/modules/generated/sklearn.gaussian_process.kernels.RationalQuadratic.html#sklearn.gaussian_process.kernels.RationalQuadratic), [GPflow](https://gpflow.readthedocs.io/en/master/gpflow/kernels/index.html#gpflow-kernels-rationalquadratic), [GPyTorch](https://docs.gpytorch.ai/en/v1.1.1/kernels.html#rqkernel), [GPy](https://gpy.readthedocs.io/en/deploy/GPy.kern.src.html#GPy.kern.src.stationary.RatQuad)...). I think it'd be valuable to keep the name and convention consistent with the literature & other packages so as to make it easier for users. (If that helps to convince you, using the SqEuclidean distance saves an unnecessary `sqrt()` and so leads to marginally faster code.)

devmotion:
This PR leaves the `RationalQuadraticKernel` unchanged. There's no deviation from the literature or other packages.

The PR only addresses the inconsistency explained in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/238. Currently we have a `GammaRationalQuadraticKernel` which has not the form of a quadratic kernel and is not a simple generalization of the `RationalQuadraticKernel` due to an additional factor in the denominator.

As mentioned in the issue, an alternative would be to make the `GammaRationalQuadraticKernel` an actual generalization of the `RationalQuadraticKernel` by adding the additional factor in the denominator. In this case, however, one would lose the property that it converges to the `GammaExponentialKernel` as `alpha` goes to infinity. Thus then the group of `ExponentialKernel`, `SqExponentialKernel` (which is the limit of the `RationalQuadraticKernel`), and `GammaExponentialKernel` is not consistent with the `RationalQuadraticKernel` and the `GammaRationalQuadraticKernel` anymore.

Therefore the approach in this PR is to just rename the `GammaRationalQuadraticKernel` to `GammaRationalKernel` (to indicate that it is not derived as a simple generalization of the `RationalQuadraticKernel`) and to add a kernel that is naturally generalized by the `GammaRationalKernel` which I therefore called `RationalKernel`. This also consistent with the exponential kernels - in the limit, the `RationalKernel` goes to the `ExponentialKernel`, the `RationalQuadraticKernel` to the `SqExponentialKernel`, and the `GammaRationalKernel` to the `GammaExponentialKernel`.

st--:
Ah, I misunderstood what was going on then. Thanks for the thorough explanation - that was helpful :)

st--:
@devmotion do you still want to merge this in ?

devmotion:
Yes, I'd still like to resolve https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/238.

st--:
Great! What's missing / still needs doing before we can merge it & resolve the issue?:)

devmotion:
Someone has to approve the proposal or suggest an alternative for how the inconsistency could be resolved.

src/basekernels/rationalquad.jl
```diff
@@ -1,3 +1,40 @@
...
+    RationalKernel(; α::Real=2.0)
+
+Rational kernel with shape parameter `α`.
+
+# Definition
+
+For inputs ``x, x' \\in \\mathbb{R}^d``, the rational kernel with shape parameter
+``\\alpha > 0`` is defined as
+```math
+k(x, x'; \\alpha) = \\bigg(1 + \\frac{\\|x - x'\\|_2}{\\alpha}\\bigg)^{-\\alpha}.
```
st--, line 11: how about turning it into a `raw`string so we can do away with all the escaping of LaTeX backslashes?

st--:
I think this proposal is fine, I'm just wondering about the choice of default values for the arguments... should the default gamma of GammaRational be such that it's just a Rational kernel?

devmotion:
No, this would break the consistency again if we do not change the default values of `GammaExponentialKernel` in the same way :smile: 

> This also consistent with the exponential kernels - in the limit, the RationalKernel goes to the ExponentialKernel, the RationalQuadraticKernel to the SqExponentialKernel, and the GammaRationalKernel to the GammaExponentialKernel.

The `GammaExponentialKernel` with default arguments is not an `ExponentialKernel` but a `SqExponentialKernel`, so it seems fine that the `GammaRationalKernel` with default arguments is not a `RationalKernel` but a `RationalQuadraticKernel`. If we want to change this, the default values of both `GammaRationalKernel` and `GammaExponentialKernel` should be changed.

st--:
> The GammaExponentialKernel with default arguments is not an ExponentialKernel but a SqExponentialKernel

Not quite though; they differ by a factor of 2 in the exponent, no?

st--:
I'm wondering if we _need_ to provide default arguments? It might be clearer to signal to the user "Hey, here's this extra argument, think about it - if you don't want to think about it you've got the wrong kernel"

devmotion:
> Not quite though; they differ by a factor of 2 in the exponent, no?

True, I missed this. In any case, you don't get the `ExponentialKernel` with the default value :wink: 

> I'm wondering if we need to provide default arguments? It might be clearer to signal to the user "Hey, here's this extra argument, think about it - if you don't want to think about it you've got the wrong kernel"

Sure, this could be done, but I think it would be better to drop default arguments in a separate PR since it is breaking and could be done for other kernels as well (e.g., `MaternKernel` which just gives a less efficient `Matern32Kernel` with its default value).

st--:
> True, I missed this. In any case, you don't get the `ExponentialKernel` with the default value 

Not with the default value currently in this PR! What I meant was, I think _if_ having a default value, setting it to gamma=1 makes more intuitive sense (to me, anyways) as in that case you do recover the Exponential kernel.

> Sure, this could be done, but I think it would be better to drop default arguments in a separate PR since it is breaking and could be done for other kernels as well (e.g., `MaternKernel` which just gives a less efficient `Matern32Kernel` with its default value).

So how about changing the default values to gamma=1, then GammaExponentialKernel is just an ExponentialKernel with more flexibility and same for GammaRationalKernel and RationalKernel ? And opening an issue to keep track of separately removing all not-particularly-helpful/intuitive default values?

Did you have a particular reason not to use `raw""` for the docstrings?

devmotion:
I would strongly prefer to not change the default values in this PR since it would be a breaking change whereas the PR otherwise is non-breaking. Therefore I suggest to discuss this in a separate issue and fix it in a separate breaking PR.

To summarize more clearly the current state:
- currently, on master and in this PR, the default value for `GammaExponentialKernel` does not give you an `ExponentialKernel` (https://juliagaussianprocesses.github.io/KernelFunctions.jl/dev/kernels/#KernelFunctions.GammaExponentialKernel)
- this PR fixes the inconsistencies of `GammaRationalQuadraticKernel` and `GammaExponentialKernel`
- to keep the relations between the kernels consistent, one has to change the default values for `GammaExponentialKernel` and `GammaRationalKernel` in the same way

> Did you have a particular reason not to use raw"" for the docstrings?

The docstring for `RationalKernel` is copied from the one for `GammaRationalQuadraticKernel` which does not use `raw"""`. Personally I don't mind the backslashes, in particular since I expect that the docstring will not be edited frequently, and therefore it seemed unnecessary work to remove them.

st--:
Yes that all sounds good. Do you want to create the issue? This PR has a merge conflict due to removing the deprecated.jl with the latest release. Happy to approve once that's fixed.

The backslashes are a bit of a pet peeve of mine, I'd be happy to go do that work myself in a separate PR (applied everywhere), would you be up for reviewing/approving that then?

devmotion:
I noticed that I had also suggested in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/238 to change the default value of gamma to 1 :smile: So I thought I could already add a deprecation warning in this PR (I also added a warning to the docstring that is shown prominently on the webpage as well).

> The backslashes are a bit of a pet peeve of mine, I'd be happy to go do that work myself in a separate PR (applied everywhere), would you be up for reviewing/approving that then?

Sure, I can review and approve such a PR.

theogf (approved):

devmotion:
I merge this now since there were no additional comments or suggestions.

PR #243, opened 2021-01-20
Deprecate `TensorProduct` correctly

devmotion:
This PR corrects the deprecation of `TensorProduct` and makes sure that dispatches on `TensorProduct` still work correctly. Unfortunately, I learnt about `Base.@deprecate_binding` too late (see also https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/242).

### DIFF
--- a/test/kernels/kerneltensorproduct.jl
+++ b/test/kernels/kerneltensorproduct.jl
@@ -28,9 +28,10 @@
     end
 
     # Deprecations
-    @test (@test_deprecated TensorProduct(k1, k2)) == k1 ⊗ k2
-    @test (@test_deprecated TensorProduct((k1, k2))) == k1 ⊗ k2
-    @test (@test_deprecated TensorProduct([k1, k2])) == k1 ⊗ k2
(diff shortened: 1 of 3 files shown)


test/kernels/kerneltensorproduct.jl
```diff
@@ -28,9 +28,10 @@
     end
 
     # Deprecations
-    @test (@test_deprecated TensorProduct(k1, k2)) == k1 ⊗ k2
-    @test (@test_deprecated TensorProduct((k1, k2))) == k1 ⊗ k2
-    @test (@test_deprecated TensorProduct([k1, k2])) == k1 ⊗ k2
+    @test TensorProduct === KernelTensorProduct
+    @test TensorProduct(k1, k2) == k1 ⊗ k2
```
devmotion, line 32: Deprecated bindings can't be tested with neither `@test_deprecated`, `@test_warn`, nor `@test_logs` it seems. But I think these tests should be sufficient since `TensorProduct` is only defined in `src/deprecated.jl`.

theogf (approved):

PR #257, opened 2021-03-18
Show git diff for formatting instead of git diff --name-only

theogf:
Right now, the format check only returns the faulty files, this is quite impractical to figure out what's wrong. The git diff would show what needs to be changed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,7 +61,7 @@ jobs:
           julia --project=test/ -e 'using JuliaFormatter; format(".", verbose=true)'
       - name: Format check
         run: |
-          CHANGED="$(git diff --name-only)"
+          CHANGED="$(git diff)"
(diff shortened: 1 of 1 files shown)


devmotion:
I'm happy to change it if it improves your workflow. We should make sure that it works though, e.g. by pushing a commit with incorrectly formatted code (which will be reverted again). In general, at least my workflow is not affected by this PR - I am too lazy to check the logs and copy from the output. For me an action that formats the code and can be triggered by a comment or label in the PR would be more helpful, in particular when changing multiple files. I think this would also be better than the suggestion in the issue (formatting the repository every day at midnight) since we don't want to merge incorrectly formatted PRs. I can make a PR with such an action if desired.

theogf:
Uuuuh a comment triggered action to format a branch sounds great!
I can also have a try at it.
Although I guess this would not work on a forked repo right?

theogf:
Closed in favor of #260

PR #258, opened 2021-03-18
Use a comment : [format] to trigger formatting.

theogf:
Everytime someone comment with "[format]" this should trigger an action where formatting is run and changes are committed to the branch as proposed in #257

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -1,0 +1,33 @@
+on:
+  pull_request:
+    types: [opened]
+  pull_request_review_comment:
+    types: [created]
+
+jobs:
+  check-comment:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: khan/pull-request-comment-trigger@master
+        id: check
+        with:
+          trigger: '[format]'
+          reaction: rocket
+        env:
+          GITHUB_TOKEN: '${{ secrets.GITHUB_TOKEN }}'
+      - run: 'echo Found it!'
+        if: steps.check.outputs.triggered == 'true'
+        steps:
+          - uses: julia-actions/setup-julia@latest
+            with:
+              version: 1
+          - uses: actions/checkout@v1
+          - name: Install JuliaFormatter and format
+            run: |
+              julia --project=test/ -e 'using Pkg; Pkg.instantiate()'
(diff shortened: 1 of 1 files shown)


theogf:
[format]

theogf:
[format]

theogf:
[format]

theogf:
test

theogf:
Hmm I believe this needs to be merged into master first to work properly...

devmotion:
Strange. AFAIK updated and new Github actions in PR are active immediately. So IMO this should be merged only if the action runs successfully.

.github/workflows/format.yml
```diff
@@ -0,0 +1,42 @@
+name: Comment-Triggered Formatting
+
+on:
+  issue_comment:
+    types: created
```
devmotion, line 5: This triggers the action also when a comment is created in a regular issue.

.github/workflows/format.yml
```diff
@@ -0,0 +1,42 @@
...
+  pull_request:
+    types: opened
```
devmotion, line 7: Is this needed?

.github/workflows/format.yml
```diff
@@ -0,0 +1,42 @@
...
+      with:
+        repo_token: ${{ secrets.GITHUB_TOKEN }}
+    - uses: khan/pull-request-comment-trigger@master # Check for a specific trigger in the comment
+      id: check
+      if: success()
+      with:
+        trigger: '[format]'
+        reaction: rocket
+      env:
+        GITHUB_TOKEN: '${{ secrets.GITHUB_TOKEN }}'
```
devmotion, line 26: IMO it is a bit unnecessary to depend on two external actions here - this can be checked with a simple `if:` statement below `runs-on:`. This alternative approach should also be more efficient since the action is skipped completely in this case.

.github/workflows/format.yml
theogf: issue-comment is for both PR and issues (see the docs of the actions below)

.github/workflows/format.yml
theogf: One could want to format right away when making the PR 😂

theogf:
Closed in favor of #260

PR #260, opened 2021-03-18
Formatting suggestions 

devmotion:
This is an additional idea: post automatic suggestions. Since I do not fully understand reviewdog yet, probably the first iterations don't work as expected :smile:


theogf:
Oh man you strike again with a much better and efficient solution again 😭😭😭

theogf:
What does the fail_on_error option does?

devmotion:
It seems to work. I am only wondering if there exists a more lightweight solution without reviewdog.

> What does the fail_on_error option does?

It ensures that the "format" action fails, i.e., it is marked prominently as failing on Github. Otherwise the action is always successful and one has to check manually if all suggestions were applied. So with this option the new action runs the current format check and posts suggestions.

theogf (approved):

PR #261, opened 2021-03-20
Fix kernel matrix functions for ScaledKernel

simonschoelly:
I noticed that some of the functions for `ScaledKernel` where incorrectly implemented, leading to a stack overflow when calling them. 

I thought I should add some tests for these cases, but then I saw that you have some general interface test function, so I was wondering if it might be better to add the mutating kernel matrix functions also to this interface test.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "KernelFunctions"
 uuid = "ec8451be-7e33-11e9-00cf-bbf324bd1392"
-version = "0.8.24"
+version = "0.8.25"
 
 [deps]
 Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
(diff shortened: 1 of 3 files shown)


willtebbutt:
Good catch! Yes, the standard test interface is definitely the right place for these.

That being said, I would have thought these cases would already be caught by e.g. [this test](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/600df21de4465c50a0bb73be344a9bf95e6212f6/src/test_utils.jl#L83). I wonder why it's not catching it... :grimacing:

willtebbutt:
We're running the standard tests on the wrong kernel :( `k` should be `ks` on [this](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/600df21de4465c50a0bb73be344a9bf95e6212f6/test/kernels/scaledkernel.jl#L13)

simonschoelly:
Yes, you are right. I fixed the test and verified that it would indeed fail without the changes made in this PR.

willtebbutt (approved):
Fantastic. Thanks for sorting this out. If you could bump the patch version to 0.8.25, I'll merge and register the new version.

PR #262, opened 2021-03-21
Avoid failures when cleaning docs

devmotion:
This PR should make the cleanup action a bit safer. Currently it fails if the preview does not exist, e.g., if a PR is opened from a fork (e.g., https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/261).

### DIFF
--- a/.github/workflows/doc_preview_cleanup.yml
+++ b/.github/workflows/doc_preview_cleanup.yml
@@ -22,3 +22,5 @@ jobs:
               git branch gh-pages-new $(echo "delete history" | git commit-tree HEAD^{tree})
               git push --force origin gh-pages-new:gh-pages
             fi
+        env:
+            PRNUM: ${{ github.event.number }}
--- a/.github/workflows/doc_preview_cleanup.yml
+++ b/.github/workflows/doc_preview_cleanup.yml
@@ -12,17 +12,13 @@ jobs:
         uses: actions/checkout@v2
         with:
           ref: gh-pages
-
-      - name: Delete preview and history
+      - name: Delete preview and history + push changes
         run: |
-            git config user.name "Documenter.jl"
-            git config user.email "documenter@juliadocs.github.io"
-            git rm -rf "previews/PR$PRNUM"
-            git commit -m "delete preview"
-            git branch gh-pages-new $(echo "delete history" | git commit-tree HEAD^{tree})
-        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/doc_preview_cleanup.yml
```diff
@@ -12,17 +12,13 @@ jobs:
...
           ref: gh-pages
-
-      - name: Delete preview and history
-        run: |
-            git config user.name "Documenter.jl"
-            git config user.email "documenter@juliadocs.github.io"
-            git rm -rf "previews/PR$PRNUM"
-            git commit -m "delete preview"
-            git branch gh-pages-new $(echo "delete history" | git commit-tree HEAD^{tree})
-        env:
```
theogf, line 23: You probably don't want to delete that :)

.github/workflows/doc_preview_cleanup.yml
```diff
@@ -12,17 +12,13 @@ jobs:
...
-            PRNUM: ${{ github.event.number }}
```
theogf, line 24: I don't think you want to delete that :smile:

.github/workflows/doc_preview_cleanup.yml
```diff
@@ -12,17 +12,13 @@ jobs:
...
-
-      - name: Push changes
+      - name: Delete preview and history + push changes
         run: |
-            git push --force origin gh-pages-new:gh-pages
+            if [ -d "previews/PR$PRNUM" ]; then
```
theogf, line 17: Also where does the `[ -d ]` syntax comes from? I could not find anything in the docs...

.github/workflows/doc_preview_cleanup.yml
devmotion: Haha no :smile:

.github/workflows/doc_preview_cleanup.yml
devmotion: This is just bash syntax. It checks if the folder actually exists.

theogf: TIL

devmotion:
Let's see if it works as expected the next time someone opens a PR from a forked repository.

devmotion:
At least it worked for this PR, so we didn't break the existing setup :slightly_smiling_face:

Issue #128, opened 2020-06-26
Adding Coregion Kernel


tirthasheshpatel:
Coregion kernel is used in coregionalization GP models and AFAIK it is only present in gpy, pymc3, and gpflow. Would it be helpful? If so, I can take a shot at it.

I am not sure how AD would work. [Here's the link to GPy's implementation](http://gpy.readthedocs.io/en/deploy/GPy.kern.src.html#GPy.kern.src.coregionalize.Coregionalize)

willtebbutt:
Hiya! Yeah, this is something that's on the TODO list for multi-output GPs. Please see [this](https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/issues/9) discussion regarding multi-output GPs. Would be great if you could have a stab at it, but I think our approach to it is going to be slightly non-standard.

tirthasheshpatel:
I am new to Julia and facing some problems setting up the environment. Is there a standard guide to build and test this package?

tirthasheshpatel:
<details>

<summary>I have something like this on my mind:</summary>

```julia
using LinearAlgebra

struct CoregionKernel{T<:Real} <: BaseKernel
    B::AbstractMatrix{T}
    CoregionKernel(; B::AbstractMatrix{T}) where {T<:Real} = new{T}(B)
    CoregionKernel(; W::AbstractMatrix{T}, κ::AbstractVector{T}) where {T<:Real} = new{T}(W*W' + diagm(κ))
end

function (κ::CoregionKernel)(x::AbstractVector{<:Real}, y::AbstractVector{<:Real})
    @assert length(x) ≠ 1 "Only one observed dimension allowed"
    idx = convert.(AbstractVector{Int}, floor.(x))
    idy = convert.(AbstractVector{Int}, floor.(y))
    return κ.B[idx, idy]
end

function (κ::CoregionKernel)(x::Real, y::Real)
    idx = convert(Int, floor(x))
    idy = convert(Int, floor(y))
    return κ.B[idx, idy]
end

Base.show(io::IO, κ::CoregionKernel) = print(io, "Coregion Kernel (B = ", κ.B, ")")

function kernelmatrix(κ::CoregionKernel, x::AbstractVector{T}) where {T<:Real}
    @assert size(x) ≠ 1 "Only one observed dimension allowed"
    idx = convert.(AbstractVector{Int}, floor.(x))
    return κ.B[idx, idx']
end

[...]

PR #263, opened 2021-03-22
Adds kernel related to the Intrinsic Coregionalization Model

4aHxKzD:
As requested and discussed in #128 

I don't know how to work with the AD packages, so I didn't add tests to cover that. Can you guide me on this please?


theogf:
So to test AD we have a `test_ADs` function which takes a function creating the kernel given some arguments : https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/0867f29270593d87f89424a89770292c60afaa21/test/test_utils.jl#L54

One could also add `B` as a training parameter using `@functor CoregionMOKernel` but I don't think it's a good idea since `B` needs to be PSD.

src/mokernels/coregion.jl
```diff
@@ -0,0 +1,36 @@
+@doc raw"""
+    CoregionMOKernel(kernel::Kernel, B::AbstractMatrix)
+
+Kernel associated with the intrinsic coregionalization model.
+
+# Definition
+
+For inputs ``x, x'`` and output dimensions ``p_x, p_{x'}'``, the kernel is defined as[^ARL]
+```math
+k\big((x, p_x), (x, p_{x'})\big) = [B]_{p_x, p_{x'}} k^{x}\big(x, x'\big),
```
devmotion, line 10: IIRC we usually mention the parameters on the LHS:
```suggestion
k\big((x, p_x), (x', p_{x'}); B, \tilde{k}\big) = [B]_{p_x, p_{x'}} \tilde{k}\big(x, x'\big),
```

src/mokernels/coregion.jl
```diff
@@ -0,0 +1,36 @@
...
+```
+where ``B`` is a positive semidefinite matrix of size ``m \times m``, with ``m`` being the number of outputs, and k^{x} is a scalar-valued kernel shared by the latent processes.
```
devmotion, line 12:
```suggestion
where ``B`` is a positive semidefinite matrix of size ``m \times m``, with ``m`` being the number of outputs, and ``\tilde{k}`` is a scalar-valued kernel shared by the latent processes.
```

4aHxKzD:
> So to test AD we have a `test_ADs` function which takes a function creating the kernel given some arguments :
> 
> https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/0867f29270593d87f89424a89770292c60afaa21/test/test_utils.jl#L54
> 
> In your case this would be something like:
> 
> ```julia
> test_ADs(L -> CoRegionKernel(SqExponentialKernel(), L * L'), rand(2, 2))
> ```

I think that function cannot be applied here because the random inputs it generates aren't designed for Multi output kernels. 

https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/0867f29270593d87f89424a89770292c60afaa21/test/test_utils.jl#L82

4aHxKzD:
I also think that the `Int` part of the input tuple is a problem for the implemented tests. We would have to change the code so that only the location is variable and the output dimension is fixed, when testing the gradient of the covariance functions.

``` julia
using KernelFunctions
using FiniteDifferences

A = randn(2,2)
B = A * A'

k = CoregionMOKernel(ConstantKernel(), B)

FDM = FiniteDifferences.central_fdm(5, 1)

x = rand(2)
y = (rand(2), 1)
```

``` julia
julia> FiniteDifferences.grad(FDM, x -> k((x,1), y), x)
([1.7763042552775138e-14, 1.7763042552775138e-14],)
```

Otherwise, if we would get an error

```julia
ERROR: MethodError: no method matching (::CoregionMOKernel{ConstantKernel{Float64},Array{Float64,2}})(::Tuple{Array{Float64,1},Float64}, ::Tuple{Array{Float64,1},Int64})
```

theogf:
> I think that function cannot be applied here because the random inputs it generates aren't designed for Multi output kernels.

Oh you're right, we probably need a different solution then...
It's probably worth it to have a different suite of tests for MO kernels. @willtebbutt @devmotion ?

willtebbutt:
We shouldn't need a separate suite of tests -- we should just need to generate inputs that are appropriate for multi-output kernels. Should just need an additional method around [here](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/ae78b7337eced196f8f90e656252cf12b287391b/src/test_utils.jl#L100).

4aHxKzD:
Now I'm confused on how does AD work for MO kernels in a general setting, i.e, if we have to adapt the inputs for it to work in our  tests how will it work when the inputs are not adapted?

4aHxKzD:
Should we create a new type for the inputs of Multioutput kernels, and then create a new method for `FiniteDifferences.to_vec` to dispatch on, as well as new `ChainRules` rules for this type of kernels?

willtebbutt:
> Such type could be called MOInput, and the current MOInput could be renamed to something like AllObservedMOInput or FullyObservedMOInput.

Is there a problem with just using a `Vector{Tuple{T, Int}}` in cases where not every output is observed all of the time?

4aHxKzD:
> Is there a problem with just using a `Vector{Tuple{T, Int}}` in cases where not every output is observed all of the time?

Yeah I was reading the documentation now and realized that too. I guess I'll try to implement that. Thanks.

4aHxKzD:
I added methods to some functions that are owned by `FiniteDifferences.jl`. I didn't know exactly where to put them so ended up placing them in the `test` directory. Would you be kind to review this?

4aHxKzD:
If this gets accepted we can improve this [test](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/1dd0378083a6bc68d1ab9aa8843929fb4424a9bf/test/mokernels/slfm.jl#L33) afterwards.

src/mokernels/coregion.jl
```diff
@@ -0,0 +1,36 @@
+@doc raw"""
+    CoregionMOKernel(kernel::Kernel, B::AbstractMatrix)
```
willtebbutt, line 2: I wonder whether this choice of name is optimal? IIUC, this is the kernel for the intrinsic coregionalization model, as opposed to the more general linear coregionalization model. So perhaps `IntrinsicCoregionMOKernel` would be a more appropriate name?

test/mokernels/coregion.jl
```diff
@@ -0,0 +1,26 @@
...
+    coregionkernel = CoregionMOKernel(kernel, B)
+
+    @test coregionkernel isa CoregionMOKernel
+    @test coregionkernel isa MOKernel
+    @test coregionkernel isa Kernel
+    @test coregionkernel(X[1], X[1]) isa Real
+    @test coregionkernel(X[1], X[1]) ≈ B[X[1][2], X[1][2]] * kernel(X[1][1], X[1][1])
+
+    @test kernelmatrix(coregionkernel, X, X) ≈ kernelmatrix(coregionkernel, X)
+
```
willtebbutt, line 24: Would it be possible additionally to run the standard suite of tests on this kernel? I.e. at least the `test_interface` function [here](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/1dd0378083a6bc68d1ab9aa8843929fb4424a9bf/test/kernels/kernelproduct.jl#L13), and preferably the `test_AD` function as well?

test/mokernels/coregion.jl
```diff
@@ -0,0 +1,26 @@
...
+    kernel = ExponentialKernel()
+    coregionkernel = CoregionMOKernel(kernel, B)
+
+    @test coregionkernel isa CoregionMOKernel
+    @test coregionkernel isa MOKernel
+    @test coregionkernel isa Kernel
+    @test coregionkernel(X[1], X[1]) isa Real
+    @test coregionkernel(X[1], X[1]) ≈ B[X[1][2], X[1][2]] * kernel(X[1][1], X[1][1])
+
+    @test kernelmatrix(coregionkernel, X, X) ≈ kernelmatrix(coregionkernel, X)
```
willtebbutt, line 23: This test ought to be redundant -- see comment below about the standard test suite.

willtebbutt (commented):
This looks really close to being ready to go. Just a couple of things.

willtebbutt:
@4aHxKzD sorry, I missed your last comment before reviewing.

I think the main point here is that we really want to ignore the `Int` output dimension-specifiers in the input dimensions. My suggestions for a `to_vec` method should sort this out.

src/mokernels/coregion.jl
4aHxKzD: Yeah I also think `IntrinsicCoregionMOKernel` is better.

4aHxKzD (commented):
> @4aHxKzD sorry, I missed your last comment before reviewing.

which comment are you referring to?

test/finitedifferences.jl
```diff
@@ -0,0 +1,33 @@
+function FiniteDifferences.to_vec(x::Tuple{T,Int}) where {T}
```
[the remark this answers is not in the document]
4aHxKzD: Yeah it seems to work.

4aHxKzD:
Just one question, should de gradients return Tuples?

test/finitedifferences.jl
```diff
@@ -0,0 +1,33 @@
...
+    function MOinput_from_vec(x_vec)
+        return first(x_vec)
+    end
+    return [x], MOinput_from_vec
+end
+
+FiniteDifferences.to_vec(x::Vector{Tuple{T,Int}}) where {T} = (x, identity)
+
+function FiniteDifferences._j′vp(
+    fdm, f, ȳ::Vector{<:Real}, x::Vector{Tuple{T,Int}}
```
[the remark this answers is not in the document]
willtebbutt: Ah fair enough. Either way, I guess you don't need this method now?

willtebbutt:
> Just one question, should de gradients return Tuples?

Good question. Certainly `Zygote` should be returning `(value, nothing)` as you were seeing before. This is part of the thing where `to_vec`'s semantics are a bit tricky.

If it's causing you problems, possibly it would make sense to run the AD checks on a function that builds the inputs from a `Vector{<:Real}`, so that the Int element of each of the `Tuple`s doesn't come in to play when comparing the gradients.

4aHxKzD:
> Good question. Certainly `Zygote` should be returning `(value, nothing)` as you were seeing before.

Even from a mathematical point of view?

willtebbutt:
> Even from a mathematical point of view?

That's my view -- in this instance the `Int`s involved are better thought of as symbols that refer to an output, rather than `Real` numbers. For example, `Int`s are used all of the time in `getindex(::Array, ::Int)`, and it's probably from this that you're getting the `nothing`s from -- you can't call `getindex` on `Real`s, so it doesn't make sense to consider perturbing the `Int` argument.

willtebbutt:
> From a programming and automatic differentiation perspective that makes sense to me. But from a mathematical one, I was assuming that the gradient would return two reals, one for each input tuple.

That's fair enough, it just turns out that AD doesn't line perfectly with one's intuition about what should happen in lots of cases -- in any case it's really a question of definition. For example, there are some pretty compelling arguments that, for the sake of AD, `Int`s should _never_ be treated as `Real` numbers, and that `Real` numbers should never be treated as `Complex` numbers, for the sake of AD. More generally, it's not clear that treating types that in some sense represent a subset of another type (Int being a subset of Real, Real of Complex, etc) is typically a good idea. There's a whole separate debate about the best way to treat subtypes of AbstractArray (are they array-like things or struct-like things?).

I guess flipping the question on its head a bit, in this instance, if it were the case that the second element of the tuples were to be a `Real`, what value would you expect them to take?

willtebbutt:
Haha fair enough -- AD is a bit unintuitive at times. Definitely it's better to think of a `Tuple` as a collection of variables, and treat each element separately.

4aHxKzD:
Imagine I'm using Julia REPL and I'm `using` `KernelFunctions`, how do I have access to the functions defined in the `test` folder?

willtebbutt:
Probably the easiest way is to `]activate test` and `include` them directly - I'm assuming you're referring to things like [this](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/master/test/test_utils.jl) file?

4aHxKzD:
> I'm assuming you're referring to things like [this](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/master/test/test_utils.jl) file?

Yes, I want to try `test_AD`.

4aHxKzD:
I get this
```julia
(test) pkg> dev .
[ Info: Resolving package identifier `.` as a directory at `~/.julia/dev/KernelFunctions/test`.
ERROR: expected a `name` entry in project file at `~/.julia/dev/KernelFunctions/test/Project.toml`
```

src/test_utils.jl
```diff
@@ -97,6 +97,18 @@ function test_interface(
     )
 end
 
+function test_interface(
+    rng::AbstractRNG, k::MOKernel, ::Type{Vector{Tuple{T,Int}}}; kwargs...
+) where {T<:Real}
+    return test_interface(
+        k,
+        [(randn(rng, T), 1) for i in 1:1001],
```
4aHxKzD, line 105: I would be better to generate random `Int`s depending on the output dimension of the kernel, instead of testing just the first dimension. Something like `rand(1:out_dim)`.

willtebbutt: Agreed. Also, would you consider making these vectors quite a lot shorter in length? Perhaps only a handful of entries. CI already takes quite a while to run.

4aHxKzD: Sure, I was just following what I saw in the other methods. How would you do the `Int` issue? For this MOKernel I could use `size(k.B,1)`, but other MOkernels don't have the `B` attribute.

willtebbutt: Oh, good point -- I hadn't considered that.

Perhaps just go with `1` for now, and open an issue about this? It feels like something that needs fixing more broadly -- perhaps we need all of the `MOKernel`s to implement a method that returns the number of output dimensions that have or something.

test/mokernels/intrinsiccoregion.jl
```diff
@@ -0,0 +1,26 @@
...
+    kernel = ExponentialKernel()
+    icoregionkernel = IntrinsicCoregionMOKernel(kernel, B)
+
+    @test icoregionkernel isa IntrinsicCoregionMOKernel
+    @test icoregionkernel isa MOKernel
+    @test icoregionkernel isa Kernel
+    @test icoregionkernel(X[1], X[1]) isa Real
+    @test icoregionkernel(X[1], X[1]) ≈ B[X[1][2], X[1][2]] * kernel(X[1][1], X[1][1])
+
+    KernelFunctions.TestUtils.test_interface(icoregionkernel, Vector{Tuple{Float64,Int}}, dim_out=out_dim)
```
willtebbutt, line 23: Thanks for adding this. Could you also add AD tests? (See one of the other kernel test sets for an example). We've not rolled them into the same function yet.

willtebbutt: Hmmm good question. @theogf wrote these tests. Is there a way to provide a custom function to test the gradient of @theogf ? I think we want to ignore the integer dimension of the input tuples.

4aHxKzD:
I think that currently we are able to use the standard AD test suit for this Kernel, but only for Zygote. ForwardDiff and ReverseDiff still need some work.

willtebbutt:
Interesting. I'm keen to get this in, so perhaps it makes sense to open an issue about the limitations with the test suite you've encountered, and we push this through. @devmotion @theogf what are your thoughts on this?

test/mokernels/intrinsiccoregion.jl
theogf: I wrote them some time ago :sweat_smile: . But yeah `x` and `y` are the example for trying AD on one sample and `A` and `B` are for collection of samples.
I don't know if the best approach would be to adapt the current testing function to allow "multi-output inputs" or to write a whole new function for it.

theogf:
> Interesting. I'm keen to get this in, so perhaps it makes sense to open an issue about the limitations with the test suite you've encountered, and we push this through. @devmotion @theogf what are your thoughts on this?

If I understood correctly the problem is to have a AD testing suite for multi-output kernels?
Maybe we could just adapt the current tests, especially the `compare_gradient` which when it takes an additional argument like for example the number of outputs uses a different method.

4aHxKzD:
> If I understood correctly the problem is to have a AD testing suite for multi-output kernels?
> Maybe we could just adapt the current tests, especially the `compare_gradient` which when it takes an additional argument like for example the number of outputs uses a different method.

I went with that approach and created new methods that dispatch depending on the inputs. If the inputs are for MOKernels it uses the methods I wrote.

4aHxKzD:
I think that the only remaining issue has to do with ForwardDiff and ReverseDiff, for example
```julia
FDM = FiniteDifferences.central_fdm(5, 1)
dims = (in=3, out=2, obs=3)
rank = 1

A = randn(dims.out, rank)
B = A * transpose(A) + Diagonal(rand(dims.out))

x = (rand(rng, dims.in), rand(rng, 1:(dims.out)))
y = (rand(rng, dims.in), rand(rng, 1:(dims.out)))

k = IntrinsicCoregionMOKernel(ExponentialKernel(), B)
```

```julia
julia> FiniteDifferences.grad(FDM, a->k(a,y), x)
(([-0.07237237059988992, 0.0448664468371161, 0.05577453225011874], 1),)
```

```julia
julia> Zygote.gradient(a->k(a,y), x)
(([-0.07237237059987332, 0.04486644683712366, 0.0557745322501128], nothing),)
```

```julia
julia> ForwardDiff.gradient(a->k(a,y), x)
ERROR: MethodError: no method matching gradient(::var"#162#163", ::Tuple{Vector{Float64}, Int64})
```

```julia
julia> ReverseDiff.gradient(a->k(a,y), x)
ERROR: MethodError: no method matching similar(::Int64)
```

theogf:
Since ForwardDiff/ReverseDiff expect vectors, what you could do is something like:
```julia
gradient(x[1]) do x
   X = (x, output_dims)
   return k(X, y)
end
```
Or something similar.

4aHxKzD:
What about the case when we have collections of observations?

theogf:
Something similar like:
```julia
A = [(randn(rng, dims.in), rand(rng, 1:(dims.out))) for i in 1:(dims.obs)]
gradient(first.(A)) do a
      A = tuple.(a, last.(A))
      my_test_function(A, B)
end
```
(don't know if it would work though...)

4aHxKzD:
> Something similar like:
> 
> ```julia
> A = [(randn(rng, dims.in), rand(rng, 1:(dims.out))) for i in 1:(dims.obs)]
> gradient(first.(A)) do a
>       A = tuple.(a, last.(A))
>       my_test_function(A, B)
> end
> ```
> 
> (don't know if it would work though...)
> The whole idea is to reconstruct the desired structure in the gradient function

For collections it doesn't seem to work.

ForwarDiff
```julia
ERROR: MethodError: no method matching one(::Type{Vector{Float64}})
```

ReverseDiff
```julia
ERROR: MethodError: no method matching zero(::Type{Vector{Float64}})

```

theogf:
OK then you would have to first make it a matrix and then transform it as a vector of vector 😓

4aHxKzD:
> OK then you would have to first make it a matrix and then transform it as a vector of vector sweat
> So something like this :
> 
> ```julia
> A = [(randn(rng, dims.in), rand(rng, 1:(dims.out))) for i in 1:(dims.obs)]
> gradient(reduce(hcat, first.(A))) do a
>       A = tuple.(eachcol(a), last.(A))
>       my_test_function(A, B)
> end
> ```

Oh, I didn't see that you also provided the code. So I wrote my own ahah. But yours is better looking, maybe I should change to it.

4aHxKzD:
I think that this is now complete. Can you guys review it please?

src/mokernels/intrinsiccoregion.jl
```diff
@@ -0,0 +1,36 @@
...
+where ``B`` is a positive semidefinite matrix of size ``m \times m``, with ``m`` being the number of outputs, and ``\tilde{k}`` is a scalar-valued kernel shared by the latent processes.
+
+[^ARL]: M. Álvarez, L. Rosasco, & N. Lawrence (2012). [Kernels for Vector-Valued Functions: a Review](https://arxiv.org/pdf/1106.6251.pdf).
+"""
+struct IntrinsicCoregionMOKernel{K<:Kernel,T<:AbstractMatrix} <: MOKernel
+    kernel::K
+    B::T
+
+    function IntrinsicCoregionMOKernel{K,T}(kernel::K, B::T) where {K,T}
+        @check_args(IntrinsicCoregionMOKernel, B, (eigmin(B) >= 0), "B is Positive semi-definite")
```
devmotion, line 21:
```suggestion
        @check_args(IntrinsicCoregionMOKernel, B, eigmin(B) >= 0, "B has to be positive semi-definite")
```

src/mokernels/intrinsiccoregion.jl
```diff
@@ -0,0 +1,36 @@
...
+        return new{K,T}(kernel, B)
+    end
+end
+
+function IntrinsicCoregionMOKernel(kernel::Kernel, B::AbstractMatrix)
```
devmotion, line 26: Maybe add a keyword argument constructor instead?

src/mokernels/intrinsiccoregion.jl
```diff
@@ -0,0 +1,36 @@
...
+    return IntrinsicCoregionMOKernel{typeof(kernel),typeof(B)}(kernel, B)
+end
+
+function (k::IntrinsicCoregionMOKernel)((x, px)::Tuple{Any,Int}, (y, py)::Tuple{Any,Int})
+    return k.B[px, py] * k.kernel(x, y)
+end
+
+function Base.show(io::IO, ::IntrinsicCoregionMOKernel)
+    return print(io, "Intrinsic Coregion Multi-Output Kernel")
```
devmotion, line 35: Add some additional information about e.g. output dimension and the kernel?

test/test_utils.jl
```diff
@@ -194,3 +249,72 @@ function test_AD(AD::Symbol, kernelfunction, args=nothing, dims=[3, 3])
...
+        end
+
+        # Testing kernel matrices
+        A = [(randn(rng, dims.in), rand(rng, 1:(dims.out))) for i in 1:(dims.obs)]
+        B = [(randn(rng, dims.in), rand(rng, 1:(dims.out))) for i in 1:(dims.obs)]
+
+        A_out_idxs = last.(A)
+        B_out_idxs = last.(B)
+
+        compare_gradient(AD, hcat(first.(A)...)) do a
```
4aHxKzD, line 283: Don't know what I was thinking, this doesn't make sense.

test/test_utils.jl
```diff
@@ -194,3 +249,72 @@ function test_AD(AD::Symbol, kernelfunction, args=nothing, dims=[3, 3])
...
+            matrix = hcat(first.(A)...)
```
4aHxKzD, line 284: Don't know what I was thinking, this doesn't make sense.

test/runtests.jl
```diff
@@ -52,6 +52,7 @@ using KernelFunctions.TestUtils: test_interface
 @info "Packages Loaded"
 
 include("test_utils.jl")
+# include("finitedifferences.jl")
```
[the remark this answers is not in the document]
4aHxKzD: I guess not, because now we are just passing the vector of reals to the gradient functions.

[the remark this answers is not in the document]
devmotion: Then it would be good to remove the file and the comment :slightly_smiling_face:

test/finitedifferences.jl
```diff
@@ -0,0 +1,7 @@
+function FiniteDifferences.to_vec(x::Tuple{T,Int}) where {T}
```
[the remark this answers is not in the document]
willtebbutt: Agreed -- an issue in FiniteDifferences probably suffices. There's a fairly general issue here about the treatment of `Int` in FiniteDifferences, so could to point out that we've found this to be an issue.

test/runtests.jl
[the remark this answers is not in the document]
4aHxKzD: You are right. I didn't yet because I wanted to be sure that the new approach worked for all cases.

theogf:
> Oh, I didn't see that you also provided the code. So I wrote my own ahah

You did not dream, I added it later. I did not try it though. What's the error with it?

4aHxKzD:
> > Oh, I didn't see that you also provided the code. So I wrote my own ahah
> 
> You did not dream, I added it later. I did not try it though. What's the error with it?

It returns a matrix of NaNs.

test/test_utils.jl
4aHxKzD: Does it? @theogf

theogf:
> > > Oh, I didn't see that you also provided the code. So I wrote my own ahah
> > 
> > 
> > You did not dream, I added it later. I did not try it though. What's the error with it?
> 
> It returns a matrix of NaNs, when using ForwardDiff and ReverseDiff.

Maybe try to see where the error comes from, if it is from the test function or from all the transformations we first did.
Try to do something like:
```julia
gradient(reduce(hcat, first.(A)) do a
    A = tuple.(eachcol(a), outdims)
    return sum(sum, first.(A))
end
```

4aHxKzD:
> Maybe try to see where the error comes from, if it is from the test function or from all the transformations we first did.
> Try to do something like:
> 
> ```julia
> gradient(reduce(hcat, first.(A)) do a
>     A = tuple.(eachcol(a), outdims)
>     return sum(sum, first.(A))
> end
> ```

This seems to work.

theogf:
> > Maybe try to see where the error comes from, if it is from the test function or from all the transformations we first did.
> > Try to do something like:
> > ```julia
> > gradient(reduce(hcat, first.(A)) do a
> >     A = tuple.(eachcol(a), outdims)
> >     return sum(sum, first.(A))
> > end
> > ```
> 
> This seems to work.

Then I would say the problem is coming from the test_function itself...

4aHxKzD:
> > > Maybe try to see where the error comes from, if it is from the test function or from all the transformations we first did.
> > > Try to do something like:
> > > ```julia
> > > gradient(reduce(hcat, first.(A)) do a
> > >     A = tuple.(eachcol(a), outdims)
> > >     return sum(sum, first.(A))
> > > end
> > > ```
> > 
> > 
> > This seems to work.
> 
> Then I would say the problem is coming from the test_function itself...

Zygote and FiniteDifferences work though.

4aHxKzD:
The issue seems to be `kernelmatrix` with just one collection of observations. `sum(kernelmatrix(k, A))` gives an error but not `sum(kernelmatrix(k,A,B))`.

devmotion:
Does it fail only with `ExponentialKernel` or also for other kernels such as `SqExponentialKernel` and `WhiteKernel`? I ran into some AD issues with `Euclidean()` recently since the case `euclidean(x, x)` is handled e.g. in Zygote only for collections of inputs but not scalars.

4aHxKzD:
I think that was the issue @devmotion. Thank you!

test/mokernels/intrinsiccoregion.jl
```diff
@@ -0,0 +1,28 @@
...
+    B = A * transpose(A) + Diagonal(rand(dims.out))
+
+    X = [(rand(dims.in), rand(1:dims.out)) for i in 1:(dims.obs)]
+
+    kernel = ExponentialKernel()
+    icoregionkernel = IntrinsicCoregionMOKernel(kernel, B)
+
+    @test icoregionkernel isa IntrinsicCoregionMOKernel
+    @test icoregionkernel isa MOKernel
+    @test icoregionkernel isa Kernel
```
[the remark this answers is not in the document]
4aHxKzD: What do you mean by correct?

src/mokernels/intrinsiccoregion.jl
4aHxKzD:
```julia
print(io, "Intrinsic Coregion Multi-Output Kernel, with $(size(k.B, 1)) outputs and a $(k.kernel) latent process")
```
What do you think?

test/mokernels/intrinsiccoregion.jl
```diff
@@ -0,0 +1,28 @@
...
+    @test icoregionkernel(X[1], X[1]) isa Real
+    @test icoregionkernel(X[1], X[1]) ≈ B[X[1][2], X[1][2]] * kernel(X[1][1], X[1][1])
```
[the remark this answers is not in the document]
4aHxKzD: That depends if the randomly generated inputs have inputs for different output dimensions in them. Should I  create X manually instead and then test what you are suggesting?

src/mokernels/intrinsiccoregion.jl
```diff
@@ -0,0 +1,36 @@
+@doc raw"""
+    IntrinsicIntrinsicCoregionMOKernel(kernel::Kernel, B::AbstractMatrix)
+
+Kernel associated with the intrinsic coregionalization model.
+
+# Definition
+
+For inputs ``x, x'`` and output dimensions ``p, p'``, the kernel is defined as[^ARL]
+```math
+k\big((x, p_x), (x', p_{x'}); B, \tilde{k}\big) = [B]_{p_x, p_{x'}} \tilde{k}\big(x, x'\big),
```
devmotion, line 10:
```suggestion
k\big((x, p), (x', p'); B, \tilde{k}\big) = B_{p, p'} \tilde{k}\big(x, x'\big),
```

src/mokernels/intrinsiccoregion.jl
```diff
@@ -0,0 +1,36 @@
...
+```
+where ``B`` is a positive semidefinite matrix of size ``m \times m``, with ``m`` being the number of outputs, and ``\tilde{k}`` is a scalar-valued kernel shared by the latent processes.
```
devmotion, line 12: 92 characters limit:
```suggestion
where ``B`` is a positive semidefinite matrix of size ``m \times m``, with ``m`` being the
number of outputs, and ``\tilde{k}`` is a scalar-valued kernel shared by the latent
processes.
```

src/mokernels/intrinsiccoregion.jl
devmotion: Maybe a bit shorter and without interpolations:
```suggestion
    return print(io, "Intrinsic Coregion Kernel: ", k.kernel, " with ", size(k.B, 1), " outputs")
```

test/mokernels/intrinsiccoregion.jl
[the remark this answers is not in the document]
devmotion: That `icoregionkernel.B == B` and `icoregionkernel.kernel == kernel`.

test/mokernels/intrinsiccoregion.jl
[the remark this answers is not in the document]
devmotion: It should be fine with the `X` that you generated above, the kernel does not require `X[i][2] == X[j][2]`. You could just replace this line with
```suggestion
    @test icoregionkernel(X[1], X[1]) ≈ B[X[1][2], X[1][2]] * kernel(X[1][1], X[1][1])
    @test icoregionkernel(X[1], X[end]) ≈ B[X[1][2], X[end][2]] * kernel(X[1][1], X[end][1])
```

4aHxKzD:
I think that this is done guys.

willtebbutt:
I'm happy -- I'll leave @devmotion to review.

Looks like there are some formatting things to sort though.

src/mokernels/intrinsiccoregion.jl
```diff
@@ -0,0 +1,42 @@
...
+    function IntrinsicCoregionMOKernel{K,T}(kernel::K, B::T) where {K,T}
+        @check_args(IntrinsicCoregionMOKernel, B, eigmin(B) >= 0, "B has to be positive semi-definite")
+        return new{K,T}(kernel, B)
+    end
+end
+
+function IntrinsicCoregionMOKernel(kernel::Kernel, B::AbstractMatrix)
+    return IntrinsicCoregionMOKernel{typeof(kernel),typeof(B)}(kernel, B)
+end
+
```
devmotion, line 31: I don't think we should have two outer constructors, one with and one without keyword arguments. Can we remove the one without keyword arguments?
```suggestion
```

Of course, that means you also have to adjust the tests and the docstring.

devmotion (commented):
Looks good, apart from the constructor and the format issues. Can you also update the version number in Project.toml?

devmotion (approved):

willtebbutt:
Thanks for the contribution @4aHxKzD and working through the pain points. Our tools for implementing / testing multi-output kernels aren't very refined yet, so it's been really helpful to have someone work through the implementation and clarify what needs work etc.

4aHxKzD:
Thank you 😄 I learned a lot of Julia with you guys.

PR #265, opened 2021-03-23
[WIP] Run format check for PRs from forks as well

devmotion:
In principle, it should be possible to run format checks (and build the documentation) also when a PR is opened from a fork.

The security settings on Github do not allow actions triggered by forks to post suggestions or commit to the repo. However, one can upload artifacts (e.g., the diff of the formatter output or the generated documentation) and reuse them in a `workflow_run` that has write permissions and is always run with the version from the main repo (if I understand correctly). Unfortunately, it is a bit tricky to download the artifacts in the `workflow_run` action and one has to somehow save the number of the PR manually if needed (e.g., by saving it in a text file that is part of the artifacts or, as in this PR, as part of the name of the zipped artifacts).

I am not sure if it works, and probably one should test the setup in a separate dummy repo if ne wants to make sure that it actually works for PRs from forks.

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -1,4 +1,4 @@
-name: Format suggestions
+name: Format check
 
 on:
   pull_request:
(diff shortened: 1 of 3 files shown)


devmotion:
Ah so due to the restriction that the workflow definition of the main repo is used for the action with guaranteed write permission it can't be tested in this PR, it seems. The format check still passes on the master branch and does not upload an artifact in this case (first commit) but fails correctly and uploads an artifact of name "diff-256" (this is PR 256) when the format is wrong (second commit). I'll test it in a dummy repo.

devmotion:
Hahaha I just noticed that https://github.com/devmotion/test_github_action already exists :rofl:

willtebbutt:
I believe that we do this now, so that this is redundant. Please correct me if not.

PR #266, opened 2021-03-25
Removed all deprecated methods and tests

theogf:
It's time for a break :p and 0.9.0. I hope this is going to be the last minor version before 1.0 :)

I removed all the deprecation methods and their respective tests. I also removed the maha.jl tests as indicated in the TODO


theogf:
Ready for review :)

devmotion (approved):
Looks good, maybe one could remove `src/deprecated.jl` completely? It seems it is an empty file now.

theogf:
I thought I could leave it there if new deprecations are needed, instead of deleting/creating it every time

Issue #217, opened 2021-01-09
Clarify behaviour/documentation of transform()


st--:
There is probably some matter of "getting used to it", but for a user new to this package, having to specify the _inverse_ lengthscale seems deeply unintuitive compared to other GP packages.

At least `transform(k, ScaleTransform(2.0))` is still explicitly saying what it is (and there could be a `LengthscaleTransform()` that divides instead of multiplying...), but I think it is not very helpful to new users that `transform(k, 2.0)` will give you a kernel with lengthscale 0.5.

Moreover, the syntax `TransformedKernel(kernel, ...)` suggests to me that the transform is changing something about the kernel, rather than about the inputs. (And ScaleTransform on inputs vs ScaledKernel on outputs is somewhat ambiguous/confusing as well.) If not just renaming it to the more verbose `input_transform` or similar, what other ways could we clarify what is actually going on here?

Comment from @willtebbutt:
> I think I agree that we could do more here to signpost what's going on here / offer a set-the-lengthscale option. I also think we could probably do with providing a bit more of an explanation for the rationale behind the transform approach to setting length scales, but that's a job for a different PR.

(Also, would be good to document/demonstrate how to be able to set bijectors on parameters within this framework. Should that be a separate issue?)

_Originally posted by @st-- in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/213#discussion_r554259595_

willtebbutt:
> Moreover, the syntax TransformedKernel(kernel, ...) suggests to me that the transform is changing something about the kernel, rather than about the inputs.

This is a really good point. I agree that going for more verbosity by renaming to `input_transform` would make sense. The convention I've adopted before (in [Stheno.jl](https://github.com/willtebbutt/Stheno.jl)) is to call linear transformations of the inputs `stretch` rather than scale -- I don't think this is a great option though.

> At least transform(k, ScaleTransform(2.0)) is still explicitly saying what it is (and there could be a LengthscaleTransform() that divides instead of multiplying...), but I think it is not very helpful to new users that transform(k, 2.0) will give you a kernel with lengthscale 0.5.

Hmmm this is an interesting point. Would you prefer to disallow this default in favour of insisting that the more verbose version is used? (I could live with that, and I agree that it might be clearer). Moreover, I agree that a `LengthscaleTransform` object would be a good idea.

devmotion:
[...]

PR #269, opened 2021-03-27
Added \circ syntax for kernel input transformations

kaandocal:
I implemented my suggestions re: #217 by extending the \circ syntax to kernel/transformation compositions. For reasons explained in my comment there this behaves differently from `transform` for TransformedKernels in that the order of composition is different so as to preserve associativity.

### DIFF
--- a/src/kernels/transformedkernel.jl
+++ b/src/kernels/transformedkernel.jl
@@ -51,6 +51,9 @@ function transform(k::TransformedKernel, t::Transform)
     return TransformedKernel(k.kernel, t ∘ k.transform)
 end
 
+Base.:∘(k::Kernel, t::Transform) = TransformedKernel(k, t)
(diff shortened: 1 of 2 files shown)


theogf:
Nice! I would be supportive for these changes but we cannot have a discrepancy between using TransformedKernel and \circle.
I would change the legacy as well.

devmotion:
As mentioned in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/217#issuecomment-817303098, probably we should reexport `compose` from https://github.com/JuliaFunctional/CompositionsBase.jl.

devmotion:
Thanks a lot @kaandocal, I extended your PR in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/276 and merged it!

Issue #259, opened 2021-03-18
Kernel matrix normalization


simonschoelly:
Would it make sense to have a kernel in this package that wraps another kernel and ensures that the matrix is normalized by calculating

```math
K'(x, y) = K(x, y) / sqrt(K(x, x), K(y, y))
```
? I think this might be helpful in cases of non-numeric input such as strings or graphs, where one cannot simply normalize the input features. For example, kernels in GraKeL have this option: <https://github.com/ysig/GraKeL/blob/3dc692a86a49542de7314c6d904d9d0afee9d196/grakel/kernels/kernel.py#L60>.
Although in GraKeL this is done by passing an rgument to the kernel constructor.

Also, here is some paper on that topic: https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.83.2658&rep=rep1&type=pdf

willtebbutt:
Hi @simonschoelly . I think this could make sense, although I've not come across these kernels before. Maybe @devmotion or @theogf have?

willtebbutt:
Seem reasonable to me. @simonschoelly if you fancied opening a PR that implements this, I would be happy to review.

Probably best to take inspiration from how we implement [ScaledKernel](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/master/src/kernels/scaledkernel.jl), or one of the other kernels in that directory.

theogf:
Actually wouldn't that be more efficient for `kernelmatrix(X, Y)` to do something like
```julia
function kernelmatrix(k::NormalizedKernel, X, Y)
	dX = inv.(sqrt.(kernelmatrix_diag(k.kernel, X)))
	dY = inv.(sqrt.(kernelmatrix_diag(k.kernel, Y)))
	K = kernelmatrix(k.kernel, X, Y)
	return dX .* K .* dY'
end
```
I am not sure about the return (needs to be checked) but just wanted to throw the idea.

devmotion:
Fixed by https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/274.

PR #274, opened 2021-04-07
Add NormalizedKernel

rossviljoen:
This PR adds a composite kernel to normalize a kernel as follows:

`k'(x, y) = k(x, y) / sqrt(k(x, x), k(y, y))`

As discussed in #259.


src/kernels/normalizedkernel.jl
```diff
@@ -0,0 +1,88 @@
...
+A normalized kernel derived from `k`.
+
+# Definition
+
+For inputs ``x, x'``, the normalized kernel ``\\widetilde{k}`` derived from
+kernel ``k`` is defined as
+```math
+\\widetilde{k}(x, x'; k) = \\frac{k(x, x')}{\\sqrt{k(x, x) k(x', x')}}.
+```
+
```
devmotion, line 13:
```suggestion
```

src/kernels/normalizedkernel.jl
```diff
@@ -0,0 +1,88 @@
...
+struct NormalizedKernel{Tk<:Kernel} <: Kernel
+    kernel::Tk
+end
+
+@functor NormalizedKernel
+
+(κ::NormalizedKernel)(x, y) = κ.kernel(x, y) / sqrt(κ.kernel(x, x) * κ.kernel(y, y))
+
+function kernelmatrix(κ::NormalizedKernel, x::AbstractVector, y::AbstractVector)
+    return kernelmatrix(κ.kernel, x, y) ./ _normalizationmatrix(κ, x, y)
```
devmotion, line 24: The separate functions `_normalizationmatrix` and `_normalizationmatrix_diag` cause additional allocations since the broadcasting operations are not fused anymore.

src/kernels/normalizedkernel.jl
```diff
@@ -0,0 +1,88 @@
...
+    )
+end
+
+function _normalizationmatrix_diag(
+    κ::NormalizedKernel, x::AbstractVector, y::AbstractVector
+)
+    return sqrt.(kernelmatrix_diag(κ.kernel, x) .* kernelmatrix_diag(κ.kernel, y))
+end
+
+Base.show(io::IO, κ::NormalizedKernel) = printshifted(io, κ, 0)
```
devmotion, line 80: Isn't this the default fallback?

src/kernels/normalizedkernel.jl
```diff
@@ -0,0 +1,88 @@
...
+
+function printshifted(io::IO, κ::NormalizedKernel, shift::Int)
+    print(io, "Normalized Kernel:" * "\n")
```
devmotion, line 83:
```suggestion
    println(io, "Normalized Kernel:")
```

devmotion (changes requested):

devmotion:
Thanks for your PR! It looks very good :+1: I left some comments, mostly I am worried about the additional allocations. Could we try to make sure that broadcasting operations are fused as much as possible?

src/kernels/normalizedkernel.jl
rossviljoen: It looks like the default is just plain print, so I think this should stay?

devmotion: Yes, you're right, I remembered it incorrectly :+1:

rossviljoen:
Thanks for the comments! I think I fused the broadcasts as much as possible?

I also optimised the `kernelmatrix_diag` functions to just return the one vector, but is there a better way to get the right return shape/type of the inner kernel than something like `map(one, kernelmatrix_diag(κ.kernel, x))`?

devmotion:
Hm, in general it is not guaranteed that `map(one, ...)` returns a vector with the same element type as `kernelmatrix_diag(κ, x, x)`, e.g., if the kernel does not return floating point numbers. So it is a bit unsatisfying even if probably such a situation is quite rare. Maybe it would be better (and more efficient, I think) to define
```julia
function kernelmatrix_diag(κ::NormalizedKernel, x::AbstractVector)
    first_x = first(x)
    return Fill(κ.kernel(first_x, first_x), length(x))
end
```
Then the number type of `kernelmatrix_diag(κ, x)` and `kernelmatrix_diag(κ, x, x)` would be the same, it seems, and it would not even allocate a full vector. However, of course, this would imply that the exact types of `kernelmatrix_diag(κ, x)` and `kernelmatrix_diag(κ, x, x)` would still be different.

src/kernels/normalizedkernel.jl
```diff
@@ -0,0 +1,81 @@
...
+end
+
+function kernelmatrix(κ::NormalizedKernel, x::AbstractVector)
+    x_diag = kernelmatrix_diag(κ.kernel, x)
+    return kernelmatrix(κ.kernel, x) ./ sqrt.(x_diag .* permutedims(x_diag))
+end
+
+function kernelmatrix_diag(κ::NormalizedKernel, x::AbstractVector)
+    first_x = first(x)
+    return fill(κ(first_x, first_x), length(x))
```
devmotion, line 36: Have you tried to use
```suggestion
    return Fill(κ(first_x, first_x), length(x))
```
with `Fill` from FillArrays? It should behave like `fill` but does not actually allocate an array.

devmotion (approved):
Looks good to me if tests pass. Thank you for this great PR! Can you update the version number in Project.toml to 0.9.2 so we can make a new release once the PR is merged?

rossviljoen:
Thank you!

There is a failing test for the Zygote gradient of `kernelmatrix_diag`, seemingly because:

```
julia> A = [1 2; 3 4]
julia> Zygote.gradient(x -> sum(kernelmatrix_diag(NormalizedKernel(SqExponentialKernel()), x)), A)

ERROR: MethodError: no method matching (::KernelFunctions.var"#ColVecs_pullback#175")(::Vector{Union{Nothing, Vector{Float64}}})
```

I'm not that familiar with Zygote, so I'm not sure if the problem is with `ColVecs_pullback` or elsewhere?
Incidentally, the gradient for a vector is fine:
```
julia> B = [1, 2]
julia> Zygote.gradient(x -> sum(kernelmatrix_diag(NormalizedKernel(SqExponentialKernel()), x)), B)
([0.0, 0.0],)
```

theogf:
Hey! kernelmatrix_diag expert here! I think the issue might from using broadcast over each sample. Typically using broadcast over ColVecs or RowVecs lead to problems with Zygote...

theogf:
Also we did not have the FillArrays dependency before, there is no way to do without it?

devmotion:
Does it work with `fill` instead of `Fill`?

rossviljoen:
The gradient doesn't work with `fill` either, no.

devmotion:
> Also we did not have the FillArrays dependency before, there is no way to do without it? 

It seemed fine to me since it only depends on standard libraries and is used by many packages in the ecosystem (such as AbstractGPs 😉). In principle, it should be more efficient to use `Fill` instead of `fill`, I was just worried if it breaks AD but it doesn't seem to be the case.

src/kernels/normalizedkernel.jl
```diff
@@ -0,0 +1,81 @@
...
+end
+
+function kernelmatrix(κ::NormalizedKernel, x::AbstractVector)
+    x_diag = kernelmatrix_diag(κ.kernel, x)
+    return kernelmatrix(κ.kernel, x) ./ sqrt.(x_diag .* permutedims(x_diag))
+end
+
+function kernelmatrix_diag(κ::NormalizedKernel, x::AbstractVector)
+    first_x = first(x)
+    return Fill(κ(first_x, first_x), length(x))
```
theogf, line 36: Can't we replace `Fill` by `Ones` here?
```suggestion
    return Ones{typeof(κ(first_x, first_x))},(length(x))
```

theogf: Or can some kernel return a negative value...

src/kernels/normalizedkernel.jl
```diff
@@ -0,0 +1,81 @@
...
+    kernel::Tk
+end
+
+@functor NormalizedKernel
+
+(κ::NormalizedKernel)(x, y) = κ.kernel(x, y) / sqrt(κ.kernel(x, x) * κ.kernel(y, y))
+
+function kernelmatrix(κ::NormalizedKernel, x::AbstractVector, y::AbstractVector)
+    return kernelmatrix(κ.kernel, x, y) ./
+           sqrt.(
```
theogf, line 24: Isn't it generally more efficient to ocmpute the `sqrt` first? This is probably a performance detail though (and I don't know about machine accuracy)

src/kernels/normalizedkernel.jl
rossviljoen: Look like this fixes the AD, cheers :)

src/kernels/normalizedkernel.jl
rossviljoen: Doesn't this all get fused anyway though?

src/kernels/normalizedkernel.jl
rossviljoen: I suppose k(x, x) could be zero as well?

theogf: But then we're in trouble cause we would be dividing by 0 and you would probably have bigger problems generally.

devmotion:
Are you satisfied with the PR @theogf? To me it seems it could be merged after updating the version number again (version 0.9.2 already exists) and adding a compatibility bound for FillArrays.

theogf:
Sure LGTM!

theogf (approved):

devmotion:
Thank you @rossviljoen!

PR #276, opened 2021-04-11
Extension of #269: Use `\circ` and `compose` and deprecate `transform`

devmotion:
This PR builds on https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/269 and takes it even further. `transform` is deprecated and instead `\circ` and its alias `CompositionsBase.compose` are used to construct transformed kernels.

Fixes https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/217.


src/deprecations.jl
```diff
@@ -0,0 +1,4 @@
+@deprecate transform(k::Kernel, t::Transform) ∘(k, t)
+@deprecate transform(k::TransformedKernel, t::Transform) ∘(k.kernel, t ∘ k.transform)
```
devmotion, line 2: As noted in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/269, currently the transformations of TransformedKernels are chained in a weird way (IMO it's a bug). Therefore the deprecation looks a bit strange as well.

src/kernels/transformedkernel.jl
```diff
@@ -41,31 +41,8 @@ function _scale(t::ScaleTransform, metric::Union{SqEuclidean,DotProduct}, x, y)
...
-transform(k::Kernel, ρ::Real) = transform(k, ScaleTransform(ρ))
-
-"""
-    transform(k::Kernel, ρ::AbstractVector)
-
-Create a [`TransformedKernel`](@ref) for kernel `k` and inverse lengthscales `ρ`.
-"""
-transform(k::Kernel, ρ::AbstractVector) = transform(k, ARDTransform(ρ))
-
-kernel(κ) = κ.kernel
```
devmotion, line 68: I used the opportunity to get rid of this unexported (and quite surprising) function as well.

src/deprecations.jl
```diff
@@ -0,0 +1,4 @@
...
+@deprecate transform(k::Kernel, ρ::Real) ∘(k, ScaleTransform(ρ))
+@deprecate transform(k::Kernel, ρ::AbstractVector) ∘(k, ARDTransform(ρ))
```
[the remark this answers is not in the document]
devmotion:
```suggestion
@deprecate transform(k::Kernel, t::Transform) k ∘ t
@deprecate transform(k::TransformedKernel, t::Transform) k.kernel ∘ t ∘ k.transform
@deprecate transform(k::Kernel, ρ::Real) k ∘ ScaleTransform(ρ)
@deprecate transform(k::Kernel, ρ::AbstractVector) k ∘ ARDTransform(ρ)
```

src/basekernels/gabor.jl
```diff
@@ -27,7 +27,7 @@ end
 
 (κ::GaborKernel)(x, y) = κ.kernel(x, y)
 
-_lengthscale_transform(::Nothing) = IdentitytTransform()
+_lengthscale_transform(::Nothing) = IdentityTransform()
```
theogf, line 30: Nice catch

docs/src/kernels.md
```diff
@@ -118,9 +118,6 @@ of kernels together.
 
 ```@docs
 TransformedKernel
-transform(::Kernel, ::Transform)
-transform(::Kernel, ::Real)
-transform(::Kernel, ::AbstractVector)
 ScaledKernel
```
theogf, line 121: Maybe add `compose` instead?

docs/src/transform.md
```diff
@@ -18,8 +18,8 @@ LowRankTransform(rand(10, 5)) ∘ ScaleTransform(2.0)
 A transformation `t` can be applied to a single input `x` with `t(x)` and to multiple inputs
 `xs` with `map(t, xs)`.
 
-Kernels can be coupled with input transformations with
-[`transform`](@ref). It falls back to creating a [`TransformedKernel`](@ref) but allows more
+Kernels can be coupled with input transformations with `∘` or its alias `compose`. It falls
```
theogf, line 21:
```suggestion
Kernels can be coupled with input transformations with the binary operation `∘` or its alias `compose`. It falls
```

devmotion: Hmm I don't think this is correct? `∘` can not only be used as an infix operator but also as a prefix `∘(f, g, ....)` that applies to multiple arguments, in the same way as `compose`.

src/kernels/transformedkernel.jl
theogf: That was probably my mistake :sweat_smile:

src/kernels/transformedkernel.jl
```diff
@@ -42,30 +42,18 @@ end
...
-Create a [`TransformedKernel`](@ref) for kernel `k` and transform `t`.
-"""
-transform(k::Kernel, t::Transform) = TransformedKernel(k, t)
-function transform(k::TransformedKernel, t::Transform)
-    return TransformedKernel(k.kernel, t ∘ k.transform)
-end
-
-"""
-    transform(k::Kernel, ρ::Real)
+Compose a `kernel` with a transformation `transform` of its inputs.
```
theogf, line 49: Should we repeat the mathematical definition from `TransformedKernel`? Since we want users to avoid using it directly this will remove the need to navigate the docs.

src/kernels/transformedkernel.jl
```diff
@@ -42,30 +42,18 @@ end
...
 
-Create a [`TransformedKernel`](@ref) for kernel `k` and inverse lengthscale `ρ`.
-"""
-transform(k::Kernel, ρ::Real) = transform(k, ScaleTransform(ρ))
-
-"""
-    transform(k::Kernel, ρ::AbstractVector)
+If no optimized implementation exists, a [`TransformedKernel`](@ref) is created.
```
theogf, line 51: What is meant here? Do we have any optimized implementation?

theogf:
This would also be great to add some examples for the docs of \circ.

src/kernels/transformedkernel.jl
devmotion: That's basically copied from the docs of `TransformedKernel` but I don't think there exists an optimized implementation for any transformation yet. I assumed the intention is the same as for `Distributions.truncated` vs `Distributions.Truncated`: the former may return arbitrary types whereas the latter is restricted to `Truncated` which might not be the most efficient implementation for specific truncated distributions.

src/kernels/transformedkernel.jl
devmotion: I don't like redundancy :stuck_out_tongue: The docstring of `TransformedKernel` is displayed right above this docstring, so users don't have to navigate to another page. If `TransformedKernel` would not be exported anymore, one could remove its docstring and move the mathematical definition here.

docs/src/transform.md
theogf: Yeah you're right I think I just wanted to make it clear that people could use `kernel \circ transform`

devmotion: I'll add a link to its docstring to make it clear:
```suggestion
Kernels can be coupled with input transformations with [`∘`](@ref) or its alias `compose`. It falls
```

src/kernels/transformedkernel.jl
theogf: I meant to access the docs from the REPL :) I personally almost never check API pages from the docs but overuse `?`!

willtebbutt:
I can find nothing objectionable here. Happy for this to be merged once @theogf is happy :)

src/kernels/transformedkernel.jl
theogf: Ah ok got it, so we leave it there as we might find some optimisation later?

theogf:
For me the only problem are the docs of `compose`/`circ` which should be self contained and maybe have an example or two if we decide that it should be the main API.

devmotion:
I updated the docstrings and added an example.

theogf (approved):

PR #279, opened 2021-04-16
Correction of the spectral_mixture_kernel docstring

theogf:
I was checking the docstrings and the one for the `spectral_mixture_kernel` is particularly wrong I think. I double checked with the first reference paper as well as the implementation and changed the formula/description of the arguments

### DIFF
--- a/src/basekernels/sm.jl
+++ b/src/basekernels/sm.jl
@@ -6,17 +6,17 @@
         ωs::AbstractMatrix{<:Real},
     )
 
-where αs are the weights of dimension (A, ), γs is the covariance matrix of
-dimension (D, A) and ωs are the mean vectors and is of dimension (D, A).
-Here, D is input dimension and A is the number of spectral components.
+where `αs` are the weight vector of dimension `A`, `γs` is the sqrt of the covariance matrix of
+dimension `(D, A)` and `ωs` are the concatenated mean vectors of dimension (D, A).
+Here, `D` is input dimension and `A` is the number of spectral components.
 
-`h` is the kernel, which defaults to [`SqExponentialKernel`](@ref) if not specified.
+`h` is the stationary kernel, which defaults to [`SqExponentialKernel`](@ref) if not specified.
 
-Generalised Spectral Mixture kernel function. This family of functions is  dense
+Generalised Spectral Mixture kernel function. This family of functions is dense
 in the family of stationary real-valued kernels with respect to the pointwise convergence.[1]
 
 ```math
-   κ(x, y) = αs' (h(-(γs' * t)^2) .* cos(π * ωs' * t), t = x - y
+   κ(x, y) = \sum_k \alpha_k^\top (h(γ_k x, γ_k y) \cos(π \cdot ω_k \odot (x-y)),
 ```
 
 # References:


willtebbutt:
Just checking: the main technical change here is gammas being described as the covariance vs sqrt of covariance matrix?

theogf:
Wait actually also this is wrong 😅, the gammas is the collection of the standard deviations (according to the implementation)

devmotion:
Ah yeah, I didn't clean up this docstring when updating the documentation since I am not familiar with it. Thought I'd leave it for someone else 😛

theogf:
I will take care of it more in depth (more is needed I think). I am just a bit puzzled by the Matrix representation of the variables arguments, we should allow vector of vectors and Co as well.
Also I would like to remove the s suffix of all the Greek letters

theogf:
@willtebbutt @devmotion @st-- I would like to move the `sm.jl` file into the `kernels` folder as it is a composite kernel and change its name to `spectral_mixture.jl` should I do this in this PR or do it in a future PR?

st--:
Probably easier to do it separately (then the only change would be the
include() and git mv, instant approval) and then we don't risk complicating
the review of this pr if GitHub doesn't display the diff properly anymore.

On Tuesday, August 31, 2021, Théo Galy-Fajou ***@***.***>
wrote:
> @willtebbutt @devmotion @st-- I would like to move the sm.jl file into
the kernels folder as it is a composite kernel and change its name to
spectral_mixture.jl should I do this in this PR or do it in a future PR?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or unsubscribe.
> Triage notifications on the go with GitHub Mobile for iOS or Android. <
https://ci3.googleusercontent.com/proxy/AcQ25A1ylDC0zqDJKu61QI_to28DRB56NJtELryfLp92xCwhrI7aBv2AI631Y7U40UfDxp8ADzt_jc2z2cN0S1SeqJk0CqjAq2CiGejInEXLppx8YPxfkcE-BSFuAuXpSOmfSaieKta21mymYBin6nHwcWHorc6CXUZjMwIGgPHilf01fmfNv0l8VXPY45h89etcRuVNboCWDOV482KuQeOdPgyeaess3dyUev_P1A=s0-d-e1-ft#https://github.com/notifications/beacon/ABL7FD4KOGQIQOEGOBO2NVTT7STBDA5CNFSM43BPENS2YY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGOGYX3BGY.gif>

theogf:
I had to make some substantial changes  but now things should be all correct!

test/basekernels/sm.jl
```diff
@@ -3,32 +3,33 @@
...
-    k1 = spectral_mixture_kernel(αs₁, γs, ωs)
+    k1 = SpectralMixtureKernel(αs₁, γs, ωs)
     k2 = spectral_mixture_product_kernel(αs₂, γs, ωs)
 
     t = v1 - v2
 
-    @test k1(v1, v2) ≈ sum(αs₁ .* exp.(-(t' * γs)' .^ 2 ./ 2) .* cospi.((t' * ωs)')) atol =
-        1e-5
+    @test k1(v1, v2) ≈ sum(
+        αs₁[k] * exp(-norm(t .* γs[:, k])^2 / 2) * cospi(2 * dot(ωs[:, k], t)) for k in 1:K
```
st--, line 18: Can you comment on the change of the factor 2? (Also in the [docstring](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/279/files#diff-ec083e8fd2a1f2ad871d1ae801aa9325fd2d81b178e9d4c861d9c59624b2558aR19)) - this also affects whether this is "just" a deprecation or actually a breaking change?

theogf:
@devmotion @st--bump :)

src/basekernels/sm.jl
```diff
@@ -29,77 +44,129 @@ in the family of stationary real-valued kernels with respect to the pointwise co
...
-dimension (D, A) and ωs are the mean vectors and is of dimension (D, A).
-Here, D is input dimension and A is the number of spectral components.
-
-Spectral Mixture Product Kernel. With enough components A, the SMP kernel
+The spectral mixture product is tensor product of spectral mixture kernel applied
+on each dimension as described in [1] in equations 13 and 14.
+With enough components, the SMP kernel
 can model any product kernel to arbitrary precision, and is flexible even
-with a small number of components [1]
+with a small number of components
```
st--, line 119:
```suggestion
with a small number of components.
```

src/transform/lineartransform.jl
```diff
@@ -2,6 +2,7 @@
     LinearTransform(A::AbstractMatrix)
 
 Linear transformation of the input realised by the matrix `A`.
+If `a` is an `AbstractVector`, `Diagonal(a)` is passed
```
st--, line 5:
```suggestion

    LinearTransform(a::AbstractVector)

Linear transformation of the input with `A = Diagonal(a)`.    
```

st--:
I did not go through the code/maths in detail so I can't comment on that, but I trust you in having fixed it.

Still needs a version bump.

Do you think it'd be worthwhile to add tests for the methods not yet covered ? (Not gonna complain if you say no not worth it)

src/basekernels/sm.jl
```diff
@@ -29,77 +44,129 @@ in the family of stationary real-valued kernels with respect to the pointwise co
...
-    ωs::AbstractMatrix{<:Real},
+    α::AbstractMatrix{<:Real},
+    γ::AbstractMatrix{<:Real},
+    ω::AbstractMatrix{<:Real},
 )
-    if !(size(αs) == size(γs) == size(ωs))
-        throw(DimensionMismatch("The dimensions of αs, γs, ans ωs do not match"))
+    (size(α) == size(γ) == size(ω)) ||
+        throw(DimensionMismatch("α, γ and ω have different dimensions"))
+    return spectral_mixture_product_kernel(h, RowVecs(α), RowVecs(γ), RowVecs(ω))
```
devmotion, line 154: Maybe we should remove `spectral_mixture_product_kernel`? Or not allow matrices as inputs? Or add a `dim` argument? It seems very weird to only use `RowVecs` here and this makes me wonder if this method and the method below should be removed completely.

st--: Seems to me to be its own issue, beyond the scope of this PR. "Reference in new issue" this?

theogf: There will be another PR to move this stuff to a different folder, this could be done together.

devmotion: > Seems to me to be its own issue, beyond the scope of this PR

Hmm, the purpose seems to be to correct the spectral mixture kernel and there seem to be many changes of `spectral_mixture_product_kernel`. So I think this PR is the right place to e.g. deprecate this function or add a `obsdim` argument.

theogf: Adding an `obsdim` argument sounds reasonable, I will add it

PR #283, opened 2021-05-03
Transfer NeuralKernelNetwork over from Stheno

willtebbutt:
To be honest, this implementation probably needs some love. I've transferred it over as-is (more or less) so that it's here, but we should open an issue about it upon merging.


src/kernels/neuralkernelnetwork.jl
```diff
@@ -0,0 +1,116 @@
+import KernelFunctions: kernelmatrix, kernelmatrix_diag
+export LinearLayer, product, Primitive, NeuralKernelNetwork
+
+using .Flux
+using .Flux: softplus, @functor
```
devmotion, line 5: Are only `softplus` and `@functor` required from Flux? `@functor` is defined in the lightweight Functors package, `softplus` in NNlib (which we also don't want to depend on, I assume, but maybe `NNlib` could just reexport the implementation in LogExpFunctions?).

willtebbutt: This are both good points -- I hadn't thought too much about this.

The functor stuff seems like a no-brainer.

`softplus` I agree that LogExpFunctions.jl would make more sense. Indeed, overall there seems to be minimal need to use Flux.

willtebbutt: Think I've resolved this now, please let me know if this is what you're after.

test/Project.toml
```diff
@@ -7,6 +7,7 @@ Flux = "587475ba-b771-5e3f-ad9e-33799f191a9c"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 Kronecker = "2c470bb0-bcc8-11e8-3dad-c9649493f05e"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+LogExpFunctions = "2ab3a3ac-af41-5b50-aa03-7779005ae688"
```
devmotion, line 10: We should add a compat entry as well.

theogf:
If I understand correctly, you map your samples with a series of kernel (the primitives) and then pass this to a NN with the guarantee of a positive output? I think that the docs for the kernel could be made more explicit in this sense, maybe in a series of equations?
And I see that the original name is indeed NeuralKernelNetwork so let's just keep this one :)

devmotion:
Isn't `LinearLayer` just a `LinearTransform` with a special parameterization? And couldn't `Primitives` be replaced by just a tuple or vector of kernels, similar to what we use in `KernelSum`, `KernelProduct`, and `KernelTensorProduct`? But maybe these are things you thought should be addressed in subsequent PRs?

willtebbutt:
> But maybe these are things you thought should be addressed in subsequent PRs?

Exactly -- in this PR I'm just copying over the implementation (which was contributed to Stheno a while ago, prior to moving over to KernelFunctions etc) so that we have _an_ implementation of this thing. I agree that this should be addressed in a follow up at some point.

> I think that the docs for the kernel could be made more explicit in this sense, maybe in a series of equations?

This seems reasonable. Again, I'm keen to address this in follow up -- if you're both comfortable with it, I would just like to get this in so that future work can look at it.

I would actually be really happy _not_ to export this for the time being so that we don't have to make a breaking change in KernelFunctions if we make a breaking change to this kernel in follow up work.

devmotion:
Seems reasonable to me, I agree that probably it should not be exported currently.

willtebbutt:
@devmotion @theogf if there's nothing else, is this ready to go in?

src/kernels/neuralkernelnetwork.jl
devmotion: Sorry, I missed your reply. Looks good now, thanks for fixing it!

willtebbutt:
@theogf have you seen failures like this before on 1.3? The error look really quite large, but it's running quite an old version of Zygote now... also weird that it's only happening on ubuntu.

willtebbutt:
😬 in that case, can I merge this?

theogf (approved):

PR #291, opened 2021-05-30
Added vcat for ColVecs/RowVecs

kaandocal:
Tentative fix for [this issue](https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/issues/155) in AbstractGPs.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -76,6 +76,9 @@ Base.getindex(D::ColVecs, i::CartesianIndex{1}) = view(D.X, :, i)
 Base.getindex(D::ColVecs, i) = ColVecs(view(D.X, :, i))
 Base.setindex!(D::ColVecs, v::AbstractVector, i) = setindex!(D.X, v, :, i)
 
+Base.vcat(a::ColVecs, b::ColVecs) = ColVecs(vcat(a.X, b.X))
+Base.vcat(a::ColVecs, b::AbstractMatrix) = ColVecs(vcat(a.X, b))
+
 dim(x::ColVecs) = size(x.X, 1)
 
 pairwise(d::PreMetric, x::ColVecs) = Distances.pairwise(d, x.X; dims=2)
@@ -143,6 +146,8 @@ Base.getindex(D::RowVecs, i::Int) = view(D.X, i, :)
 Base.getindex(D::RowVecs, i::CartesianIndex{1}) = view(D.X, i, :)
 Base.getindex(D::RowVecs, i) = RowVecs(view(D.X, i, :))
 Base.setindex!(D::RowVecs, v::AbstractVector, i) = setindex!(D.X, v, i, :)
+
+Base.vcat(a::RowVecs, b::RowVecs) = RowVecs(vcat(a.X, b.X))
 
 dim(x::RowVecs) = size(x.X, 2)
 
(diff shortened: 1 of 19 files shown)


src/utils.jl
```diff
@@ -76,6 +76,9 @@ Base.getindex(D::ColVecs, i::CartesianIndex{1}) = view(D.X, :, i)
 Base.getindex(D::ColVecs, i) = ColVecs(view(D.X, :, i))
 Base.setindex!(D::ColVecs, v::AbstractVector, i) = setindex!(D.X, v, :, i)
 
+Base.vcat(a::ColVecs, b::ColVecs) = ColVecs(vcat(a.X, b.X))
```
devmotion, line 79: I guess this should be
```suggestion
Base.vcat(a::ColVecs, b::ColVecs) = ColVecs(hcat(a.X, b.X))
```

src/utils.jl
```diff
@@ -76,6 +76,9 @@ Base.getindex(D::ColVecs, i::CartesianIndex{1}) = view(D.X, :, i)
...
+Base.vcat(a::ColVecs, b::AbstractMatrix) = ColVecs(vcat(a.X, b))
```
devmotion, line 80: This is not well defined, I'd suggest to remove it:
```suggestion
```

devmotion (changes requested):

devmotion:
Can you also add some tests?

src/utils.jl
kaandocal: I thought it could be useful but you're right.

willtebbutt (approved):
LGTM. Just requires a patch version bump then we're good to go provided that @devmotion is happy.

devmotion (commented):
I think the tests could be improved a bit and actually check that we do the correct thing, e.g., by concatenating two different objects and checking that the concatenated values are correct (I also don't think that the check `typeof(DX) == typeof(vcat(DX, DX))` is important - it's not the main purpose of the implementation and only a by-product if one actually concatenates two `RowVecs` or `ColVecs` whose underlying matrices have the same type). Otherwise, only the version bump seems to be missing :slightly_smiling_face:

kaandocal:
@devmotion I've modified the tests to what you suggested - nothing extensive, but given the simplicity of the PR I wasn't at my most creative. Let me know if you have other suggestions!

test/utils.jl
```diff
@@ -33,6 +33,7 @@
               pairwise(SqEuclidean(), X; dims=2)
         @test KernelFunctions.pairwise(SqEuclidean(), DX, DY) ≈
               pairwise(SqEuclidean(), X, Y; dims=2)
+        @test vcat(DX, DY) == ColVecs(hcat(X, Y))
```
devmotion, line 36: A small suggestion since `ColVecs(X) == collect(eachcol(X))` is `true` as well:
```suggestion
        @test vcat(DX, DY) isa ColVecs
        @test vcat(DX, DY).X == hcat(X, Y)
```

test/utils.jl
```diff
@@ -72,6 +73,7 @@
               pairwise(SqEuclidean(), X; dims=1)
         @test KernelFunctions.pairwise(SqEuclidean(), DX, DY) ≈
               pairwise(SqEuclidean(), X, Y; dims=1)
+        @test vcat(DX, DY) == RowVecs(vcat(X, Y))
```
devmotion, line 76:
```suggestion
        @test vcat(DX, DY) isa RowVecs
        @test vcat(DX, DY).X == vcat(X, Y)
```

devmotion (approved):

willtebbutt (approved):

devmotion:
Thanks a lot for the PR :+1:

Issue #281, opened 2021-04-20
CI not cancelled when new commit pushed


willtebbutt:
I've been commiting lots of stuff in #280 , thereby generating a very large number of jobs for CI to churn through.

Is there a way to ensure that existing jobs on a particular branch are cancelled when someone pushes a new commit?

edit: for example, [our actions page](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/actions) is now flooded with CI for tiny docs udpates.

devmotion:
AFAICT (e.g., https://github.community/t/is-it-possible-to-cancel-previous-jobs-automatically-when-the-pull-request-is-updated/16779/5 and https://github.community/t/github-actions-cancel-redundant-builds-not-solved/16025) such a feature is not supported by Github actions. However, it seems some people created custom actions for this purpose which one could maybe use. Since I haven't looked into this more carefully, I don't know which one is recommended though.

Of course, an alternative would be to not push after each commit if there are many more to come or to cancel jobs manually if they are not needed anymore.

devmotion:
Seems to be addressed by https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/292.

PR #292, opened 2021-06-01
Cancel previous running workflows on push

devmotion:
Just learnt about https://github.com/styfle/cancel-workflow-action from https://github.com/JuliaDiff/ChainRules.jl/issues/426. Seems this could solve https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/281.

I followed the instructions in https://github.com/styfle/cancel-workflow-action#advanced-pull-requests-from-forks which should also work for PRs from forks.

### DIFF
--- a/.github/workflows/Cancel.yml
+++ b/.github/workflows/Cancel.yml
@@ -1,0 +1,18 @@
+name: Cancel
+
+on:
+  workflow_run:
+    workflows:
+      - "CI"
+      - "Documentation"
+      - "Format suggestions"
+    types:
+      - requested
+
+jobs:
+  cancel:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: styfle/cancel-workflow-action@0.9.0
+      with:
+        workflow_id: ${{ github.event.workflow.id }}


mzgubic:
It doesn't seem like it ran, did it?

devmotion:
IIRC this type of actions always uses the version on master for security reasons. It doesn't exist currently so it's not run.

devmotion:
It is mentioned in the [documentation](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_run):

> Note: This event will only trigger a workflow run if the workflow file is on the default branch.

devmotion:
Of course, this makes it difficult to assess if the action works properly here :smile: 

@willtebbutt @theogf What do you think, should we just try it and see if it works and possibly fix it in follow-up PRs?

willtebbutt (approved):
Yeah, let's do it, 🤞

theogf:
Go for it!

devmotion:
It seems to be at least somewhat correct and working: https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/actions/runs/897590843

PR #297, opened 2021-06-02
[Do not merge] testing cancel action with `all_but_latest: true`

mzgubic:
https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/295

Edit: the above seems to be the only difference between this repo (which seemed to work) and ChainRules.jl (which didn't seem to) so I wonder whether adding this line breaks it for some reason.


willtebbutt:
@mzgubic can we close this?

PR #251, opened 2021-01-27
Contributing guidelines

st--:
So far only contains initial notes on how to make a release (thanks @theogf for the explanation). Just to get the ball rolling. Please feel free to push to this branch and add to it!

PR #303, opened 2021-06-16
Examples in the docs: base PR for cleaning up notebook workflow

st--:
Following #314, this PR adds READMEs for how to build & add to the docs and examples, as well as tidying the workflow up a bit more. It makes all current examples run (with a warning to say they're under construction). This PR is NOT intended to clean up the notebooks themselves - that'll come in later PRs.


docs/make.jl
```diff
@@ -7,6 +8,25 @@ end
 
 using KernelFunctions
 
+const EXAMPLES_SRC = joinpath(@__DIR__, "..", "examples")
+const EXAMPLES_OUT = joinpath(@__DIR__, "src", "examples")
+const BLACKLIST = [
+    "deepkernellearning",
+    "kernelridgeregression",
+    "svm",
+]
```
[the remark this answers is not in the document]
st--: This is really nice, but showing how I could batch-format the code before committing/pushing would be even nicer. :) @devmotion
(one of the things I would like to cover in #251)

docs/README.md
```diff
@@ -0,0 +1,35 @@
+# How to build the docs locally
```
st--, line 1: Happy to get improvements to the workflow. :)

docs/README.md
```diff
@@ -0,0 +1,35 @@
...
+The built docs will be underneath build/
+
+# How to contribute to the docs
+
+## To add additional docs dependencies
+
+```bash
+julia --project=. -e 'using Pkg; Pkg.add("NewDependency")'
+```
+and commit the changes to Project.toml
```
st--, line 33: should we commit the Manifest.toml as well?

docs/Project.toml
```diff
@@ -1,6 +1,12 @@
 [deps]
+Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
```
devmotion, line 2: I think it would be cleaner and probably easier to maintain (less conflicting dependencies etc) if `docs/Project.toml` only contains the dependencies for the actual documentation and each example uses a separate project environment with its own pinned dependencies.

devmotion: I have played around with this in https://github.com/devmotion/CalibrationErrors.jl.

docs/README.md
devmotion: It's common at least. But I think it is more important to commit the Manifest.toml for the examples.

docs/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+Install docs dependencies:
+```bash
+julia --project=. -e 'using Pkg; Pkg.instantiate()'
+```
+This will use the last released KernelFunctions.jl for building the docs. 
+
+If instead you want to reflect the local state of the Git repository in your built docs, make sure to change the dependency to the development version:
+```bash
+julia --project=. -e "using Pkg; Pkg.develop(path=\"$(readlink -f ..)\")"
```
devmotion, line 13: I guess this does not work on every OS?

docs/README.md
st--: The examples/ subdirectory doesn't have its own Project.toml - to me it would make sense to simply move them underneath docs/ (and then commit the Manifest.toml) - what do you think?

docs/Project.toml
st--: Yes, I agree - I was wondering how you would make that work but the Pkg.activate() context manager is really neat, so I'll try to add that :)

docs/README.md
st--: good point ... the key issue is that relative paths mean that you can only execute correctly from the same directory, so it's better to use the absolute path; I'll add a second example with a "manual" path (I do like snippets that I can simply copy&paste instead of having to edit).

docs/Project.toml
st--: Mhm, that would mean having to run the Pkg.develop in every single example subdirectory if you want to work on them though, and would that not make it harder to check the examples run consistently with any future changes to KernelFunctions (in a PR)?

devmotion: > Mhm, that would mean having to run the Pkg.develop in every single example subdirectory if you want to work on them

Yes, the main point is to have independent environments. In this way you can avoid that dependencies of different examples are conflicting and it is clear which packages and which versions were used to run the individual examples.

They are re-run every time the documentation is built so one would just check that the build passes and the webpage looks fine.

docs/README.md
devmotion: The suggestion would be to have a separate Project.toml + Manifest.toml for every example.

docs/Project.toml
st--: Hm, for the examples to actually pull in the in-development version of the package I think your [make.jl](https://github.com/devmotion/CalibrationErrors.jl/blob/main/docs/make.jl) would need to also include a `Pkg.develop(path=joinpath(@__DIR__, ".."))`, but with that I think it should work:)

I think it'd still be good to provide a Makefile or some other way of making it easy for contributors to locally run tests, formatter, etc..

docs/make.jl
```diff
@@ -7,6 +8,30 @@ end
 
 using KernelFunctions
 
+const PACKAGE_DIR = joinpath(@__DIR__, "..")
+const EXAMPLES_SRC = joinpath(PACKAGE_DIR, "examples")
+const EXAMPLES_OUT = joinpath(@__DIR__, "src", "examples")
+const BLACKLIST = [
```
devmotion, line 14: Maybe it would be easier to just keep a list of the examples and iterate through them instead of using a blacklist?

docs/README.md
```diff
@@ -0,0 +1,37 @@
...
+```
+To undo the pinning to the local path by the previous command, run
+```bash
+julia --project=. -e 'using Pkg; Pkg.free("KernelFunctions")'
+```
+
+To actually build the docs, run
+```bash
+julia --project=. make.jl
+```
```
st--, line 27: Can we instead just wrap the make.jl core with
```julia
Pkg.activate(@__DIR__) do
    Pkg.develop(path=joinpath(@__DIR__, ".."))
    Pkg.instantiate()
    # ...all the doc-generating code relying on KernelFunctions
end
```
so that developers only have to run `julia docs/make.jl`?
@devmotion

docs/README.md
st--: Should we commit all the Manifest.toml files? In that case, should we remove it from the .gitignore?

docs/make.jl
st--: I thought it was your suggestion to move away from manually maintaining a list and simply pick up everything underneath examples :wink: 

I intended the blacklist to be only temporary so we can get the build to pass focussing on the make.jl setup itself, without having to fix the notebooks. I would be strongly against keeping examples around that don't build, so as soon as we've fixed up the examples, I would delete the blacklist code.

docs/README.md
devmotion: In my experience usually you don't want to run docs/make.jl if you edit or add an example since this will also run doctests, build the documentation, and rerun all other possibly long-running examples.

st--: Yes, if I am just adding an example, I can simply run the example script.jl itself :)
This was, instead, to address the use-case of developers changing the KernelFunctions.jl/src who may want to check (locally) that all examples still build correctly.

st--: (And also just to simplify the workflow for anyone who wants to build docs locally: can we remove steps for the user by automating it inside the script? This would also remove some magic from the CI workflow)

devmotion: I am not sure, I really think usually there's no reason to build the docs locally. Doctests are included in the tests anyways (if you want to run them, they also take forever so most users probably also want to run only a subset of the tests) and other issues with e.g. the examples we'll notice when the Github actions run. I think one should also keep in mind that we do not deviate completely from the usual workflow and setup in Julia packages since this could be quite confusing for contributors. There are community standards for documentation, CI, etc. so I think it's good to follow them if there is no clear reason to do otherwise.

BTW I also think it is more convenient to not rerun the whole example always but use one Julia session and rerun only the modified blocks when editing these files (this can be done e.g. with VSCode, Emacs, or Juno with some shortcuts).

devmotion: Actually, I think that it's probably problematic if the examples are rerun with the master branch of KernelFunctions. It seems it is possible that other packages are not compatible yet with a breaking new version of KernelFunctions, and then we might run into package conflicts in the examples. So probably it is better to use a fixed version of KernelFunctions in the examples and use CompatHelper as a reminder that it is updated to new breaking versions.

Due to the Manifest.toml it will be clear which version was used to run the example and it is reproducible, and due to CompatHelper we won't forget to update it.

st--: My intention is not to change the workflow necessarily, just to make it easier for new contributors, as described in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/251#issuecomment-863193042 - it'd be great if you could add in there what you think the best workflow for all those points is!

st--: @devmotion re CompatHelper, I'm assuming you're much more familiar with how to set that up than I am, can you add code suggestions for what's needed (or just push to this branch directly, I'd be happy with that too!) ?

st--: Though I think in general it'd be better if the burden for fixing examples (if needed) were on the person who makes a PR to KernelFunctions, not to some unfortunate maintainer who wants to bump the compat version some point in the future...

devmotion: > re CompatHelper, I'm assuming you're much more familiar with how to set that up than I am, can you add code suggestions for what's needed (or just push to this branch directly, I'd be happy with that too!) ?

It should be sufficient to add the relevant subdirectories in this line: https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/efca34f298277cf464a841a1ee8acf4e256871c1/.github/workflows/CompatHelper.yml#L16

> Though I think in general it'd be better if the burden for fixing examples (if needed) were on the person who makes a PR to KernelFunctions, not to some unfortunate maintainer who wants to bump the compat version some point in the future...

This would be the ideal but I am worried that it does not work reliably e.g. due to package conflicts. Additional, I fear that it actually makes the examples less reproducible if we update  the Manifest.toml when building them since this will lead to a discrepancy between the version in the repo and the version that was used to run the example.

theogf: Yeah to go in the direction of @st-- , if it's someone who never contributed to this repo we would need to run the actions ourselves. It's a good idea to have instructions on how to do it locally!

devmotion: Everything apart from the preview also works for PRs from forks (and I think even this should be possible but requires some workarounds and custom settings in Documenter - I started to play around with it in a PR but then forgot it...), so we would still know f there are any errors or not.

theogf: @devmotion I meant this problem : 
image

devmotion: Ah I see. But actions still work, we just have to approve them if someone contributes for the first time (IIRC Github introduced it due to the mining problems).

theogf: It's true but it is slow and annoying (for the maintainers) hence the need to run the documentation locally for contributors not having to wait for us approving the workflows and for us to have to approve every new doc PR

devmotion: Yes, I can see that it might be helpful for new contributors to check "quickly" if it builds as expected. But regardless of if they build it locally or not, we have to approve the runs first and wait for the tests to pass before it can be merged.

willtebbutt: Might it make sense to discuss the workflow when you want to re-build the docs multiple times (because you're making lots of small edits)? In particular, I find that it's often better to open up a REPL session and to `include("make.jl")` to avoid having to re-compile everything each time.

docs/README.md
```diff
@@ -0,0 +1,37 @@
...
+The built docs will be underneath build/
```
willtebbutt, line 28:
```suggestion
The built docs will be underneath build/, and are best viewed in a browser.
```
This is probably obvious to most, but perhaps not all.

docs/README.md
devmotion: Yes, I think this should be mentioned (ar least this is how I do it 😄). I also comment out parts that I don't need such as examples. And vice versa, if I just want to run an example I don't build the whole docs with docs/make.jl but just the example in its folder. And if I want to work on an example, I use the REPL and rerun blocks that I modify (with Emacs, Juno, VSCode or whatever other editor). In the REPL you can also just run `] dev KernelFunctions` instead of dealing with paths.

st--: If you already have a decent workflow in mind, please go ahead and leave review suggestions and/or push directly to this branch:) it's supposed to reflect community best practices, not my own stabs at figuring out workflow _without_ any directions :smile:

examples/support-vector-machine/script.jl
```diff
@@ -1,26 +1,23 @@
...
+Xgrid = hcat(collect.(Iterators.product(xgrid, xgrid))...)'  # Combine into a 2D grid
+y = rand([-1, 1], N)  # Select randomly between the two classes
 X = zeros(N, 2)
-X[y .== 1, :] = rand(MvNormal(μ[:, 1], I), count(y .== 1))' #Attribute samples from class 1
-X[y .== -1, :] = rand(MvNormal(μ[:, 2], I), count(y .== -1))' # Attribute samples from class 2
+X[y .== 1, :] = rand(MvNormal(μ[:, 1], I), count(y .== 1))'  # Attribute samples from class 1
+X[y .== -1, :] = rand(MvNormal(μ[:, 2], I), count(y .== -1))'  # Attribute samples from class 2
 
-k = SqExponentialKernel(2.0) # Create kernel function
+k = SqExponentialKernel()  # Create kernel function
```
theogf, line 14:
```suggestion
k = compose(SqExponentialKernel(), ScaleTransform(2.0))  # Create kernel function
```
This would be the correct equivalent

docs/make.jl
```diff
@@ -7,6 +9,25 @@ end
...
+const PACKAGE_DIR = joinpath(@__DIR__, "..")
+const EXAMPLES_SRC = joinpath(PACKAGE_DIR, "examples")
+const EXAMPLES_OUT = joinpath(@__DIR__, "src", "examples")
+const BLACKLIST = ["deep-kernel-learning", "kernel-ridge-regression"]
+
+ispath(EXAMPLES_OUT) && rm(EXAMPLES_OUT; recursive=true)
+mkpath(EXAMPLES_OUT)
+
+for example in readdir(EXAMPLES_SRC)
+    any([occursin(blacklisted, example) for blacklisted in BLACKLIST]) && continue
```
theogf, line 21:
```suggestion
    example ∈ BLACKLIST && continue
```

st--: Ah, now that actually works again:) neater indeed

docs/README.md
st--: I've added @willtebbutt's suggestion to the README. Are you happy to merge this for now, and we can iteratively refine it later? It would be good to have a base setup in place so we can actually work on improving the notebooks.

st--:
This should not stop us from merging the current PR, but the [rendered notebook](https://juliagaussianprocesses.github.io/KernelFunctions.jl/previews/PR303/examples/kernel-ridge-regression/) doesn't correctly include the displayed plot - instead of putting it inline, it looks like GR is trying to open the interactive Qt window (which of course fails on the CI server). I spent some time searching for a solution but did not find anything. Any suggestions for how to fix this?

devmotion:
I guess one should not use `display` but either build a proper animation or just plot! to a fixed plot object that is displayed after the loop. In general, GR errors can be ignored and if desired by setting the environment variable GKSwstype to "nul" (https://discourse.julialang.org/t/deactivate-plot-display-to-avoid-need-for-x-server/19359/2).

st--:
Ok, managed to fix the plotting. :)

willtebbutt (dismissed):
I'm happy with this. @theogf and @devmotion are you also happy?

theogf:
Just one last thing to be picky, maybe add a small admonition at the beginning. So that people don't think this is the definite version?
!!! warning 
    This page is in construction

examples/README.md
```diff
@@ -0,0 +1,16 @@
...
+```julia
+julia> using Literate
+julia> Literate.markdown("script.jl", "output_directory")
+julia> Literate.notebook("script.jl", "output_directory")
+```
+
+## Add a new example
+
+Create a new subdirectory in here, and put your code in a file called `script.jl` so that it will get picked up by the automatic docs build.
+You should also add a `Project.toml` file with all your script's dependencies (including KernelFunctions.jl).
```
devmotion, line 16:
```suggestion

Every example uses a separate project environment. Therefore you should also create a new
project environment in the directory of the example, install all required package there, and
commit the `Project.toml` and a `Manifest.toml` file.
```

examples/kernel-ridge-regression/script.jl
```diff
@@ -10,31 +14,55 @@ Flux.@functor ScaleTransform
...
     Plots.plot!(x_test, f(X_test, k, λ); lab="Prediction", lw=3.0)
-    display(p)
+    push!(plots, p)
 end
+
+#
+
+l = @layout grid(10, 1)
+plot(plots...; layout=l)
+plot!(; size=(300, 1500))
```
devmotion, line 71: You should be able to just use
```suggestion
plot(plots; size=(300, 1500))
```
Plotting arrays of plots should work automatically.

examples/kernel-ridge-regression/script.jl
```diff
@@ -10,31 +14,55 @@ Flux.@functor ScaleTransform
...
+
+plots = []
 for i in 1:10
     grads = Zygote.gradient(() -> loss(k, λ), ps)
     Flux.Optimise.update!(opt, ps, grads)
     p = Plots.scatter(x, y; lab="data", title="Loss = $(loss(k,λ))")
     Plots.plot!(x_test, f(X_test, k, λ); lab="Prediction", lw=3.0)
-    display(p)
+    push!(plots, p)
 end
```
devmotion, line 65: It might be more idiomatic to use
```julia
plots = map(1:10) do _
   grads = ...
   p = scatter(...) # does it have to be qualified?
   plot!(p, ...)
   return p
end
```

examples/kernel-ridge-regression/script.jl
st--: It does! In this case I like the single-column format, but it obviously makes more sense to combine the size argument:)
```suggestion
l = @layout grid(10, 1)
plot(plots...; layout=l, size=(300, 1500))
```

examples/deep-kernel-learning/script.jl
```diff
@@ -10,31 +17,56 @@ Flux.@functor KernelSum
...
     Plots.plot!(x, f(X, k, λ); lab="Prediction", lw=3.0)
-    display(p)
+    push!(plots, p)
 end
+
+#
+
+l = @layout grid(10, 1)
+plot(plots...; layout=l)
+plot!(; size=(300, 1500))
```
st--, line 72:
```suggestion
plot(plots...; layout=l, size=(300, 1500))
```

examples/README.md
```diff
@@ -0,0 +1,33 @@
...
+julia> Literate.markdown("script.jl", "output_directory")
+julia> Literate.notebook("script.jl", "output_directory")
+```
+
+## Add a new example
+
+Create a new subdirectory in here, and put your code in a file called `script.jl` so that it will get picked up by the automatic docs build.
+
+Every example uses a separate project environment. Therefore you should also create a new
+project environment in the directory of the example, install all required package there (including KernelFunctions.jl), and
```
st--, line 32:
```suggestion
project environment in the directory of the example that contains all packages required by your script. Note that you must also include both the `KernelFunctions` and the `Literate` packages. Make sure to both
```

examples/README.md
```diff
@@ -0,0 +1,33 @@
...
+julia> include("script.jl")
+```
+
+You can convert them to markdown and Jupyter notebook formats, respectively, by executing
+```julia
+julia> using Literate
+julia> Literate.markdown("script.jl", "output_directory")
+julia> Literate.notebook("script.jl", "output_directory")
+```
+
```
st--, line 26:
```suggestion
or run
`julia --project=. ../../docs/literate.jl script.jl output_directory`.

```

docs/Project.toml
```diff
@@ -2,6 +2,8 @@
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 KernelFunctions = "ec8451be-7e33-11e9-00cf-bbf324bd1392"
 PDMats = "90014a1f-27ba-587c-ab20-58faa44d9150"
+Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
+Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
```
devmotion: This is not needed, I assume?
```suggestion
```

examples/deep-kernel-learning/script.jl
```diff
@@ -10,31 +17,55 @@ Flux.@functor KernelSum
...
 x = range(xmin, xmax; length=100)
 x_test = rand(Uniform(xmin, xmax), 200)
 x, y = noisy_function(sinc, x; noise=0.1)
 X = reshape(x, :, 1)
 λ = [0.1]
+#md nothing #hide
+
+#
+
 function f(x, k, λ)
```
devmotion, line 38: Ideally this uses the same as in the "SVM" example based on ColVecs etc.

docs/README.md
st--:
```suggestion
and commit the changes to Project.toml and Manifest.toml.
```

docs/literate.jl
```diff
@@ -8,6 +8,7 @@ const OUTDIR = ARGS[2]
 # Activate environment
 using Pkg: Pkg
 Pkg.activate(dirname(SCRIPTJL))
+Pkg.add("Literate"; preserve=Pkg.PRESERVE_ALL)
```
devmotion, line 11: As written in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/314#discussion_r660380835, I am against this change since it impacts reproducibility. In addition to the reasons listed there, the examples are literate scripts that very much depend on the syntax and version of Literate - e.g., some features (not sure if used here) actually require Literate >= 2.8. So I think the appropriate way is to add Literate.jl as a dependency to the Project.toml of the examples. If contributors forget it, it will error and be easy to fix, so I also don't think that one should be worried about new contributors here.

docs/make.jl
```diff
@@ -38,11 +38,10 @@ DocMeta.setdocmeta!(
...
     :DocTestSetup,
     quote
+        using PDMats
+        using Kronecker
         using KernelFunctions
-        using LinearAlgebra
-        using Random
-        using PDMats: PDMats
-    end;
+    end; # we load all optional packages to generate the full API documentation
```
devmotion, line 44: This is just for the doctests, not the API, so I don't think this change is needed?

docs/Project.toml
```diff
@@ -1,7 +1,9 @@
 [deps]
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 KernelFunctions = "ec8451be-7e33-11e9-00cf-bbf324bd1392"
+Kronecker = "2c470bb0-bcc8-11e8-3dad-c9649493f05e"
 PDMats = "90014a1f-27ba-587c-ab20-58faa44d9150"
+Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
```
devmotion, line 6: As far as I can see this is not needed?
```suggestion
```

examples/README.md
```diff
@@ -0,0 +1,35 @@
...
+or run
+`julia --project=. ../../docs/literate.jl script.jl output_directory`.
+
+## Add a new example
+
+Create a new subdirectory in here, and put your code in a file called `script.jl` so that it will get picked up by the automatic docs build.
+
+Every example uses a separate project environment. Therefore you should also create a new
+project environment in the directory of the example that contains all packages required by your script. Note that you must also include both the `KernelFunctions` and the `Literate` packages. Make sure to both
+commit the `Project.toml` and a `Manifest.toml` file.
```
st--, line 35:
``````suggestion
project environment in the directory of the example that contains all packages required by your script.
Note that the dependencies of your example must include the `Literate` package.
From a Julia REPL started in your example script's directory, you can run
```julia
julia> ] activate .
julia> ] add Literate
julia> ] add KernelFunctions
julia> # add any other example-specific dependencies
```
to generate the project files.
Make sure to commit both the `Project.toml` and the `Manifest.toml` file when you want to contribute your example in a pull request.
``````

File: docs/README.md [hunk 23]
```diff
@@ -0,0 +1,52 @@
...
+
+```bash
+julia --project=. -e 'using Pkg; Pkg.add("NewDependency")'
+```
+and commit the changes to Project.toml
+
+
+## To add examples
+
+See [../examples](../examples/README.md)
```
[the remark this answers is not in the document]
st--: thanks!

docs/Project.toml
st--: oh yeah, good catch

docs/make.jl
st--: Without using Kronecker, the `@require` doesn't include the source for `kernelkronmat` so it won't show up in the API docs.

devmotion: I understand the intention but I think the correct approach would be to just load it after `using KernelFunctions` in make.jl. Here the packages are set that are loaded for the doctests.

st--:
If there are examples with errors, the parallel processing unfortunately means you won't know which of the scripts it was that errored. :(

devmotion:
It should print the error of the example that failed. At least it worked this way when I used it locally.

devmotion:
BTW currently the documentation errors since the Manifest.toml files of the examples are incorrect. You added your local path which of course does not exist for other users or in the CI tests. You should add KernelFunctions by navigating to the example and running
```julia
julia> ] activate .

julia> ] instantiate

julia> ] dev ../..
```
The path of KernelFunctions should just be "../.." which is always correct. Alternatively, we could fix it to a specific version of KernelFunctions and just use `] add KernelFunctions@someversion` instead of `] dev ../..`, and probably add a `Pkg.status()` output block to the examples that shows the exact version that was used. I am not completely sure what is the best approach but I think printing `Pkg.status()` would make things clearer regardless of what we use.

st--:
There seems to be an issue with the links that you added @devmotion - could you have a look into this (you're rather more familiar with that bit of the docs code)?
```
┌ Warning: couldn't find file "/u/10/johnt1/unix/JuliaCode/KernelFunctions.jl/docs/src/examples/<unknown>/examples/support-vector-machine/script.jl" when generating URL
└ @ Documenter.Utilities ~/.julia/packages/Documenter/f5jts/src/Utilities/Utilities.jl:385
┌ Warning: invalid local link: unresolved path in examples/support-vector-machine.md
│   link.text =
│    1-element Vector{Any}:
│     Markdown.Image("https://img.shields.io/badge/show-nbviewer-579ACA.svg", "")
│   link.url = "<unknown>/examples/support-vector-machine.ipynb"
└ @ Documenter.Writers.HTMLWriter ~/.julia/packages/Documenter/f5jts/src/Writers/HTMLWriter.jl:1903
┌ Warning: invalid local link: unresolved path in examples/support-vector-machine.md
│   link.text =
│    1-element Vector{Any}:
│     "Julia source file"
│   link.url = "<unknown>/examples/support-vector-machine/script.jl"
└ @ Documenter.Writers.HTMLWriter ~/.julia/packages/Documenter/f5jts/src/Writers/HTMLWriter.jl:1903
┌ Warning: invalid local link: unresolved path in examples/support-vector-machine.md
│   link.text =
│    1-element Vector{Any}:
│     "nbviewer"
│   link.url = "<unknown>/examples/support-vector-machine.ipynb"
└ @ Documenter.Writers.HTMLWriter ~/.julia/packages/Documenter/f5jts/src/Writers/HTMLWriter.jl:1903
```

st--:
> It should print the error of the example that failed. At least it worked this way when I used it locally.

I guess you can grep for the string... but it would be much easier if it simply told you which file it was so you could just open it and jump to the right line:) We could just process them sequentially... currently the rest of the build takes ~40 minutes anyways so that shouldn't make it longer overall.

st--:
> You added your local path which of course does not exist for other users or in the CI tests

Woops, good catch, I just fixed that. I prefer a shell one-liner over multiple manual lines in the REPL though:
```bash
for ex in * ; do julia --project=$ex -e 'using Pkg; Pkg.develop(path="..")' ; done
```

examples/README.md
```diff
@@ -0,0 +1,44 @@
...
+```
+
+You can convert them to markdown and Jupyter notebook formats, respectively, by executing
+```julia
+julia> using Literate
+julia> Literate.markdown("script.jl", "output_directory")
+julia> Literate.notebook("script.jl", "output_directory")
+```
+or run
+`julia --project=. ../../docs/literate.jl script.jl output_directory`.
```
devmotion, line 27: These two approaches are not equivalent though, we use different keyword arguments. Maybe one should just mention `literate.jl` and not the manual conversion (IMO the recommended workflow for writing examples would not involve running Literate at all, so it seems fine to run it in a separate process if you actually want to convert the example).

It is also not necessary to use `--project=.`, the environment is activated by `literate.jl` automatically. However, I think the script does not work if you call it in this way since it extracts the name of the example from the path of `script.jl` which is not given. So it might be better to recommend to call
````suggestion
```shell
julia docs/literate.jl examples/myexample/script.jl output_directory
```
````

**Comment by devmotion** [comment]
> There seems to be an issue with the links that you added @devmotion - could you have a look into this (you're rather more familiar with that bit of the docs code)?

This is expected if you run it locally. This also happens with standard Literate settings without preprocessing etc., so I don't think it can't be avoided if you run it outside of CI. These warnings can just be ignored.

**Comment by devmotion** [comment]
> I guess you can grep for the string... but it would be much easier if it simply told you which file it was so you could just open it and jump to the right line:) We could just process them sequentially... currently the rest of the build takes ~40 minutes anyways so that shouldn't make it longer overall.

When I ran it, it told me what line failed and which file it was. The exception was printed in the same way as standard Julia exceptions, and typically in my editor then I can also just jump to the failing line (did not check in this case though when it occurred).

**Comment by devmotion** [comment]
> Currently the rest of the build takes ~40 minutes anyways so that shouldn't make it longer overall.

IMO this is a motivation for splitting the tests into different groups that are run in parallel (similar what we do in the SciML ecosystem: https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/test/runtests.jl) rather than running the examples in serial.

**Comment by st--** [comment]
This is the error I got when running make.jl (before I fixed bugs in the examples):
```ERROR: ERROR: LoadError: LoadError: DimensionMismatch("second dimension of A, 1, does not match length of x, 100")
in expression starting at string:2
when executing the following code block

```julia
f(x, k, λ) = kernelmatrix(k, x, X) / (kernelmatrix(k, X) + exp(λ) * I) * y
f(X, k, 1.0)
```

Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:33
 [2] execute_block(sb::Module, block::String)
   @ Literate ~/.julia/packages/Literate/XP0eh/src/Literate.jl:720
 [3] execute_markdown!(io::IOBuffer, sb::Module, block::String, outputdir::String)
   @ Literate ~/.julia/packages/Literate/XP0eh/src/Literate.jl:477
 [4] markdown(inputfile::String, outputdir::String; config::Dict{Any, Any}, kwargs::LoadError: LoadError: DimensionMismatch("matrix A has dimensions (3,1), vector B has length 100")
in expression starting at string:2
when executing the following code block

```julia
f(x, k, λ) = kernelmatrix(k, x, X) / (kernelmatrix(k, X) + exp(λ) * I) * y
f(X, k, 1.0)
```

Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:33
 [2] execute_block(sb::Module, block::String)
   @ Literate ~/.julia/packages/Literate/XP0eh/src/Literate.jl:720
 [3] execute_markdown!(io::IOBuffer, sb::Module, block::String, outputdir::String)
   @ Literate ~/.julia/packages/Literate/XP0eh/src/Literate.jl:477
 [4] markdown(inputfile::String, outputdir::String; config::Dict{Any, Any}, kwargs::Base.Iterators.Pairs{Symbol, Any, NTuple{4, Symbol}, NamedTuple{(:name, :documenter, :execute, :preprocess), Tuple{String, Bool, Bool, typeof(preprocess)}}})
   @ Literate ~/.julia/packages/Literate/XP0eh/src/Literate.jl:461
 [5] top-level scope
   @ ~/JuliaCode/KernelFunctions.jl/docs/literate.jl:42
in expression starting at /u/10/johnt1/unix/JuliaCode/KernelFunctions.jl/docs/literate.jl:42
Base.Iterators.Pairs{Symbol, Any, NTuple{4, Symbol}, NamedTuple{(:name, :documenter, :execute, :preprocess), Tuple{String, Bool, Bool, typeof(preprocess)}}})
   @ Literate ~/.julia/packages/Literate/XP0eh/src/Literate.jl:461
 [5] top-level scope
   @ ~/JuliaCode/KernelFunctions.jl/docs/literate.jl:42
in expression starting at /u/10/johnt1/unix/JuliaCode/KernelFunctions.jl/docs/literate.jl:42
```

File: docs/make.jl (hunk 20 above)
Comment (st--) [reply docs/make.jl]: I've now added the comments that make this obvious from looking at the code itself for people without that background knowledge:)

**Comment by devmotion** [comment]
I see, I wonder why it was more descriptive in my case... In any case, I think we could make the errors thrown by Literate more descriptive if we wrap the `Literate.markdown... Literate.notebook` block in `literate.jl` in a try-catch statement and rethrow the error with some additional information that shows the name of the example and possibly some other useful information.

File: docs/literate.jl [hunk 25]
```diff
@@ -37,20 +38,25 @@ function preprocess(content)
...
+        SCRIPTJL,
+        OUTDIR;
+        name=basename(dirname(SCRIPTJL)),
+        documenter=false,
+        execute=true,
+        preprocess=preprocess,
+    )
+catch e
+    # add script path to error message
+    throw(LoadError(SCRIPTJL, 0, e))
```
st--, line 61: Ideally, we would get the actual line out of `e` but Literate throws an `ErrorException` in which the error is just stored as a string.

@devmotion happy to get your suggestions for improving it:)

devmotion: On a second thought, I think probably it would be best to improve the error thrown by Literate instead. Ideally it would already include the file and the line number (not sure how difficult the latter would be).

st--: Opened https://github.com/fredrikekre/Literate.jl/pull/150

docs/literate.jl
st--: I reverted it and clarified the instructions in the README.md

docs/literate.jl
```diff
@@ -23,14 +24,13 @@ function preprocess(content)
...
+# *You are seeing the
 #md # HTML output generated by [Documenter.jl](https://github.com/JuliaDocs/Documenter.jl) and
 #nb # notebook output generated by
 # [Literate.jl](https://github.com/fredrikekre/Literate.jl) from the
 # [Julia source file](@__REPO_ROOT_URL__/examples/@__NAME__/script.jl).
-# The corresponding
-#md # notebook can be viewed in [nbviewer](@__NBVIEWER_ROOT_URL__/examples/@__NAME__.ipynb).
-#nb # HTML output can be viewed [here](https://juliagaussianprocesses.github.io/KernelFunctions.jl/dev/examples/@__NAME__/).
+#md # The corresponding notebook can be viewed in [nbviewer](@__NBVIEWER_ROOT_URL__/examples/@__NAME__.ipynb).*
+#nb # The rendered HTML can be viewed [in the docs](https://juliagaussianprocesses.github.io/KernelFunctions.jl/dev/examples/@__NAME__/).*
```
st--, line 33: How difficult would it be to change the `dev` to whichever build it actually is?

devmotion: Not difficult but a bit annoying it seems. Literate automatically derives the url for the repo and nbviewer when running it with Github actions which are used to replace `@__NBVIEWER_ROOT_URL__` and `@__REPO_ROOT_URL__` after the custom user-defined replacements. However, we can't use these urls or subparts of them in our preprocessing and there are no similar "magic" variables for the github.io webpage (which seems reasonable somehow since one can use e.g. custom domains and hence it is not clear what the correct url for the webpage should be). So I think one has to copy the logic in https://github.com/fredrikekre/Literate.jl/blob/8d8caac6e651eb02ded5699835384d0f0b139e78/src/Literate.jl#L248-L258 to derive the `deploy_folder` and use it instead of `dev`. Maybe one could ask if Literate could add an automatic replacement for `@__DEPLOY_FOLDER__`, similar to `@__REPO_ROOT_URL__`.

st--:
Docs build failed because of an error in installing the CUDA package...

docs/literate.jl
st--: Thanks for looking into this. Would you be willing to write it up as an issue in the Literate.jl repo ?

fredrikekre: https://github.com/fredrikekre/Literate.jl/issues/151

st--: Awesome, thank you @fredrikekre !

examples/kernel-ridge-regression/script.jl
st--: Would you still consider it "more idiomatic" despite the statefulness? `map` makes me think of something that can be parallelised (as in map-reduce), but in this case it's crucial the steps are executed in order... which is why I'm inclined to keep it as it is for now. (Anyways, this PR isn't about fixing the notebooks themselves.)

docs/make.jl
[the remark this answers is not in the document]
theogf: in case you need it:
```julia
using JuliaFormatter # you should add it to your global env
format(".")
```

devmotion:
> Docs build failed because of an error in installing the CUDA package...

I don't understand why the installation error, it seems it tries to access `CUDA/.../.github`?! We don't use CUDA and GPUs in the examples, it's just installed as a dependency of Flux but never used properly. Does CUDA perform any special operations at installation time? I thought it just downloads the GPU/Nvidia stuff when it is actually loaded?

docs/make.jl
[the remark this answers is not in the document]
st--: Yeah it's in the Makefile in #251:)

examples/kernel-ridge-regression/script.jl
```diff
@@ -10,31 +17,50 @@ Flux.@functor ScaleTransform
...
```
devmotion, line 61: Again might want to change this to an animation later on.

docs/Project.toml
```diff
@@ -1,6 +1,7 @@
 [deps]
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 KernelFunctions = "ec8451be-7e33-11e9-00cf-bbf324bd1392"
+Kronecker = "2c470bb0-bcc8-11e8-3dad-c9649493f05e"
```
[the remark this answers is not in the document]
st--: Can you clarify what you mean - what change do you want to make here?

docs/README.md
```diff
@@ -0,0 +1,58 @@
...
+```
+again to rebuild the documentation.
+
+# How to contribute to the docs
+
+## To add additional docs dependencies
+
+```bash
+julia --project=. -e 'using Pkg; Pkg.add("NewDependency")'
+```
```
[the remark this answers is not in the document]
st--: I'm happy to include this, but do you have a strong reason not keep the shell one-liner as well?

docs/make.jl
```diff
@@ -39,10 +41,7 @@ DocMeta.setdocmeta!(
     :DocTestSetup,
     quote
         using KernelFunctions
-        using LinearAlgebra
-        using Random
-        using PDMats: PDMats
```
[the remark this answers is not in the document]
st--: clearly not :)

docs/Project.toml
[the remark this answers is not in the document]
theogf: Under the `[compat]` section, each package should have lower-bounds.
So you would need to add a line
```toml
Kronecker = "0.4"
```
for example

docs/README.md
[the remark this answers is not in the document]
theogf: Maybe give the full commands to activate the documentation as well then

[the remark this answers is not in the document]
st--:
``````suggestion
To add additional dependencies, run (after activating the documentation environment)
```julia
julia> ] add NewDependency
```
or, as a shell one-liner:
```shell
julia --project=. -e 'using Pkg; Pkg.add("NewDependency")'
```
``````

File: examples/README.md [hunk 31]
```diff
@@ -0,0 +1,49 @@
...
+You can convert them to markdown and Jupyter notebook formats, respectively, by executing
+```julia
+julia> using Literate
+julia> Literate.markdown("script.jl", "output_directory")
+julia> Literate.notebook("script.jl", "output_directory")
+```
+or run
+```shell
+julia docs/literate.jl examples/myexample/script.jl output_directory
+```
```
st--, line 30:
``````suggestion
In particular when editing an example, it can be convenient to (re-)run only some parts of
an example.
Many editors with Julia support such as VSCode, Juno, and Emacs support the evaluation of individual lines or code chunks.

You can convert a notebook to markdown and Jupyter notebook formats, respectively, by executing
```julia
julia> using Literate
julia> Literate.markdown("script.jl", "output_directory")
julia> Literate.notebook("script.jl", "output_directory")
```
(see the [Literate.jl docs](https://fredrikekre.github.io/Literate.jl/v2/) for additional options) or run
```shell
julia docs/literate.jl examples/myexample/script.jl output_directory
```
which also executes the code and generates embedded plots etc. in the same way as in building the KernelFunctions documentation.
``````

File: examples/README.md [hunk 32]
```diff
@@ -0,0 +1,48 @@
...
+julia> include("script.jl")
+```
+
+You can convert them to markdown and Jupyter notebook formats, respectively, by executing
+```julia
+julia> using Literate
+julia> Literate.markdown("script.jl", "output_directory")
+julia> Literate.notebook("script.jl", "output_directory")
+```
+or run
```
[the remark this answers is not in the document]
st--: I've rewritten it a bit

docs/Project.toml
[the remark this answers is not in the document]
st--: done

examples/support-vector-machine/Project.toml
```diff
@@ -0,0 +1,14 @@
...
+Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
+KernelFunctions = "ec8451be-7e33-11e9-00cf-bbf324bd1392"
+LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+Literate = "98b081ad-f1c9-55d3-8b20-4c87d4299306"
+Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
+
+[compat]
+Distributions = "0.25"
+KernelFunctions = "0.10"
+LinearAlgebra = "3"
```
devmotion, line 11: AFAIK there should not/don't have to be compat bounds for stdlibs (at least I have never seen them):
```suggestion
```

examples/kernel-ridge-regression/Project.toml
```diff
@@ -0,0 +1,20 @@
...
+Literate = "98b081ad-f1c9-55d3-8b20-4c87d4299306"
+MLDataUtils = "cc2ba9b6-d476-5e6d-8eaf-a92d5412d41d"
+Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
+Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
+
+[compat]
+Distributions = "0.25"
+Flux = "0.12"
+KernelFunctions = "0.10"
+LinearAlgebra = "3"
```
devmotion, line 15:
```suggestion
```

examples/deep-kernel-learning/Project.toml
```diff
@@ -0,0 +1,20 @@
...
+Literate = "98b081ad-f1c9-55d3-8b20-4c87d4299306"
+MLDataUtils = "cc2ba9b6-d476-5e6d-8eaf-a92d5412d41d"
+Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
+Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
+
+[compat]
+Distributions = "0.25"
+Flux = "0.12"
+KernelFunctions = "0.10"
+LinearAlgebra = "3"
```
devmotion, line 15:
```suggestion
```

examples/support-vector-machine/Project.toml
st--: yeah I just realized this as well :woman_facepalming:

devmotion (approved):

PR #304, opened 2021-06-16
Adding naive LMM model

thomasgudjonwright:
# Adding a naive implementation of the Linear Mixing Model

## What is the LMM?

The LMM is a model of GP that assumes `f(x) = H * g(x)` where `H` is a basis of `f` and ``g(x)` are  time varying coefficients  modelled independently with unit-variance GPs.

Resources:
* https://arxiv.org/pdf/1911.06287.pdf
* https://invenia.github.io/blog/2021/02/19/OILMM-pt1/

## Why is this just adding the naive LMM?

I will be adding the more efficient sibling, but as an intro to contributing here felt that the simpler implementation was the place to start!

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "KernelFunctions"
 uuid = "ec8451be-7e33-11e9-00cf-bbf324bd1392"
-version = "0.11.5"
+version = "0.10.6"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "KernelFunctions"
 uuid = "ec8451be-7e33-11e9-00cf-bbf324bd1392"
-version = "0.10.5"
+version = "0.11.5"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/src/mokernels/lmm.jl
+++ b/src/mokernels/lmm.jl
@@ -1,0 +1,86 @@
+@doc raw"""
+    NaiveLMMMOKernel(g, e::MOKernel, A::AbstractMatrix)
+
+Kernel associated with the linear mixing model.
+
+# Definition
+
+For inputs ``x, x'`` and output dimensions ``p_x, p_{x'}'``, the kernel is defined as[^BPTHST]
+```math
+k\big((x, p_x), (x, p_{x'})\big) = H_{p_{x}}k_{simple}(x, x')H_{p_{x'}}
+```
+where ``k_1, \ldots, k_m`` are ``m`` kernels, one for each latent process, ``H`` is a
+mixing matrix of ``m`` basis vectors spanning the output space, and ``σ²`` is noise.
+
+[^BPTHST]: Wessel P. Bruinsma, Eric Perim, Will Tebbutt, J. Scott Hosking, Arno Solin, Richard E. Turner (2020). [Scalable Exact Inference in Multi-Output Gaussian Processes](https://arxiv.org/pdf/1911.06287.pdf).
+"""
+struct NaiveLMMMOKernel{Tk<:Union{Kernel, Vector{<:Kernel}}, Th<:AbstractMatrix, F<:Float64} <: MOKernel
+    K::Tk
+    H::Th
+    σ²::F
+    # if just a simple kernel is provided, we construct the MO kernel
+    function NaiveLMMMOKernel(k::SimpleKernel, H::AbstractMatrix, σ²)
+        σ² >= 0 || error("`σ²` must be positive")
+        m = size(H,2)
+        K = [k for i in 1:m]
+        return new{typeof(K),typeof(H),typeof(σ²)}(K,H,σ²)
+    end
+    function NaiveLMMMOKernel(K::Vector{<:Kernel}, H::AbstractMatrix, σ²)
+        σ² >= 0 || error("`σ²` must be positive")
+        return new{typeof(K),typeof(H),typeof(σ²)}(K,H,σ²)
+    end
+end
+
+function (κ::NaiveLMMMOKernel)((x, px)::Tuple{Any,Int}, (y, py)::Tuple{Any,Int})
+    (px > size(κ.H, 1) || py > size(κ.H, 1) || px < 1 || py < 1) &&
+    error("`px` and `py` must be within the range of the number of outputs")
+    m = size(κ.H,2)
+    K = Matrix(undef, m, m)
+    K .= 0
+    for i in 1:m
+        K[i, i] = κ.K[i](x, y)
+    end
+    return H[px,:]' * K * H[py,:]
+end
+
+# not optimized at all
+function kernelmatrix(k::NaiveLMMMOKernel, x::MOInput, y::MOInput)
+    x.out_dim == y.out_dim || error("`x` and `y` should have the same output dimension")
+    x.out_dim == size(k.H, 1) ||
+        error("Mixing matrix not compatible with the given multi-output inputs")
+    m = size(k.H,2)
+    n_x = size(x.x,1)
+    n_y = size(y.x,1)
+    Σ = Matrix(undef, n_x, n_y)
+    for i in 1:n_x
+        for j in i:n_y
+            K_xy = Matrix(undef, m, m)
+            K_xy .= 0
+            if x.x isa Union{ColVecs,RowVecs}
+                x′ = x.x.X[i]
+            else
+                x′ = x.x[i]
+            end
+            if y.x isa Union{ColVecs,RowVecs}
+                y′ = y.x.X[j]
+            else
+                y′ = y.x[j]
+            end
+                for h in 1:m
+                    K_xy[h, h] = k.K[h](x′, y′)
+                end
+            Σ[i,j] = k.H * K_xy * k.H'
+            Σ[j,i] = Σ[i,j]
+        end
+    end
+    return Σ + k.σ²*I
+end
+
+function Base.show(io::IO, k::NaiveLMMMOKernel)
+    return print(io, "Linear Mixing Model Multi-Output Kernel (naive implementation)")
+end
+
+function Base.show(io::IO, ::MIME"text/plain", k::NaiveLMMMOKernel)
+    print(io, "Linear Mixing Model Multi-Output Kernel (naive implementation)\n\tkernels (K): ")
+    return join(io, k.K, "\n\t\t")
+end
--- a/src/mokernels/lmm.jl
+++ b/src/mokernels/lmm.jl
@@ -7,14 +7,15 @@ Kernel associated with the linear mixing model.
 
 For inputs ``x, x'`` and output dimensions ``p_x, p_{x'}'``, the kernel is defined as[^BPTHST]
 ```math
-k\big((x, p_x), (x, p_{x'})\big) = H_{p_{x}}k_{simple}(x, x')H_{p_{x'}}
+k\big((x, p_x), (x, p_{x'})\big) = H_{p_{x}}K(x, x')H_{p_{x'}}
 ```
-where ``k_1, \ldots, k_m`` are ``m`` kernels, one for each latent process, ``H`` is a
+where ``K(x, x') = Diag(k_1(x, x'), ..., k_m(x, x'))`` with zero off-diagonal entries.
+``k_1, \ldots, k_m`` are ``m`` kernels, one for each latent process, ``H`` is a
 mixing matrix of ``m`` basis vectors spanning the output space, and ``σ²`` is noise.
 
 [^BPTHST]: Wessel P. Bruinsma, Eric Perim, Will Tebbutt, J. Scott Hosking, Arno Solin, Richard E. Turner (2020). [Scalable Exact Inference in Multi-Output Gaussian Processes](https://arxiv.org/pdf/1911.06287.pdf).
 """
-struct NaiveLMMMOKernel{Tk<:Union{Kernel, Vector{<:Kernel}}, Th<:AbstractMatrix, F<:Float64} <: MOKernel
+struct NaiveLMMMOKernel{Tk<:AbstractVector{<:Kernel}, Th<:AbstractMatrix, F<:Float64} <: MOKernel
     K::Tk
     H::Th
     σ²::F
@@ -22,10 +23,10 @@ struct NaiveLMMMOKernel{Tk<:Union{Kernel, Vector{<:Kernel}}, Th<:AbstractMatrix,
     function NaiveLMMMOKernel(k::SimpleKernel, H::AbstractMatrix, σ²)
         σ² >= 0 || error("`σ²` must be positive")
         m = size(H,2)
-        K = [k for i in 1:m]
+        K = fill(k, m)
         return new{typeof(K),typeof(H),typeof(σ²)}(K,H,σ²)
     end
-    function NaiveLMMMOKernel(K::Vector{<:Kernel}, H::AbstractMatrix, σ²)
+    function NaiveLMMMOKernel(K::AbstractVector{<:Kernel}, H::AbstractMatrix, σ²)
         σ² >= 0 || error("`σ²` must be positive")
         return new{typeof(K),typeof(H),typeof(σ²)}(K,H,σ²)
     end
@@ -40,7 +41,7 @@ function (κ::NaiveLMMMOKernel)((x, px)::Tuple{Any,Int}, (y, py)::Tuple{Any,Int}
     for i in 1:m
         K[i, i] = κ.K[i](x, y)
     end
-    return H[px,:]' * K * H[py,:]
+    return H[px,:]' * K * H[py,:] + σ²
 end
 
 # not optimized at all
--- a/src/mokernels/lmm.jl
+++ b/src/mokernels/lmm.jl
@@ -1,5 +1,5 @@
 @doc raw"""
-    NaiveLMMMOKernel(g, e::MOKernel, A::AbstractMatrix)
+    NaiveLinearMixingModelKernel(g, e::MOKernel, A::AbstractMatrix)
 
 Kernel associated with the linear mixing model.
 
@@ -17,24 +17,24 @@ mixing matrix of ``m`` basis vectors spanning the output space.
 
 [^BPTHST]: Wessel P. Bruinsma, Eric Perim, Will Tebbutt, J. Scott Hosking, Arno Solin, Richard E. Turner (2020). [Scalable Exact Inference in Multi-Output Gaussian Processes](https://arxiv.org/pdf/1911.06287.pdf).
 """
-struct NaiveLMMMOKernel{Tk<:AbstractVector{<:Kernel}, Th<:AbstractMatrix} <: MOKernel
+struct NaiveLinearMixingModelKernel{Tk<:AbstractVector{<:Kernel}, Th<:AbstractMatrix} <: MOKernel
     K::Tk
     H::Th
 end
 
-NaiveLMMMOKernel(k::Kernel, H::AbstractMatrix) = NaiveLMMMOKernel(Fill(k, size(H, 1)), H)
+NaiveLinearMixingModelKernel(k::Kernel, H::AbstractMatrix) = NaiveLinearMixingModelKernel(Fill(k, size(H, 1)), H)
 
-function (κ::NaiveLMMMOKernel)((x, px)::Tuple{Any,Int}, (y, py)::Tuple{Any,Int})
+function (κ::NaiveLinearMixingModelKernel)((x, px)::Tuple{Any,Int}, (y, py)::Tuple{Any,Int})
     (px > size(κ.H, 2) || py > size(κ.H, 2) || px < 1 || py < 1) &&
     error("`px` and `py` must be within the range of the number of outputs")
     return sum(κ.H[i, px] * κ.K[i](x, y) * κ.H[i, py] for i in 1:length(κ.K))
 end
 
-function Base.show(io::IO, k::NaiveLMMMOKernel)
+function Base.show(io::IO, k::NaiveLinearMixingModelKernel)
     return print(io, "Linear Mixing Model Multi-Output Kernel (naive implementation)")
 end
 
-function Base.show(io::IO, mime::MIME"text/plain", k::NaiveLMMMOKernel)
+function Base.show(io::IO, mime::MIME"text/plain", k::NaiveLinearMixingModelKernel)
     print(io, "Linear Mixing Model Multi-Output Kernel (naive implementation). Kernels:")
     for k in k.K
         print(io, "\n\t")
--- a/src/mokernels/lmm.jl
+++ b/src/mokernels/lmm.jl
@@ -41,12 +41,3 @@ function Base.show(io::IO, mime::MIME"text/plain", k::LinearMixingModelKernel)
         show(io, mime, k)
     end
 end
-
-## Aliases ##
-
-"""
-    LMMKernel()
-
-Alias of [`LinearMixingModelKernel`](@ref).
-"""
-const LMMKernel = LinearMixingModelKernel
--- a/src/mokernels/lmm.jl
+++ b/src/mokernels/lmm.jl
@@ -34,7 +34,10 @@ function Base.show(io::IO, k::NaiveLMMMOKernel)
     return print(io, "Linear Mixing Model Multi-Output Kernel (naive implementation)")
 end
 
-function Base.show(io::IO, ::MIME"text/plain", k::NaiveLMMMOKernel)
-    print(io, "Linear Mixing Model Multi-Output Kernel (naive implementation)\n\tkernels (K): ")
-    return join(io, k.K, "\n\t\t")
+function Base.show(io::IO, mime::MIME"text/plain", k::NaiveLMMMOKernel)
+    print(io, "Linear Mixing Model Multi-Output Kernel (naive implementation). Kernels:")
+    for k in k.K
+        print(io, "\n\t")
+        show(io, mime, k)
+    end
 end
--- a/src/mokernels/lmm.jl
+++ b/src/mokernels/lmm.jl
@@ -27,7 +27,7 @@ NaiveLMMMOKernel(k::Kernel, H::AbstractMatrix) = NaiveLMMMOKernel(Fill(k, size(H
 function (κ::NaiveLMMMOKernel)((x, px)::Tuple{Any,Int}, (y, py)::Tuple{Any,Int})
     (px > size(κ.H, 2) || py > size(κ.H, 2) || px < 1 || py < 1) &&
     error("`px` and `py` must be within the range of the number of outputs")
-    return sum(H[i, px] * κ.K[i](x, y) * H[i, py] for i in 1:length(κ.K))
+    return sum(κ.H[i, px] * κ.K[i](x, y) * κ.H[i, py] for i in 1:length(κ.K))
 end
 
 function Base.show(io::IO, k::NaiveLMMMOKernel)
--- a/src/mokernels/lmm.jl
+++ b/src/mokernels/lmm.jl
@@ -10,6 +10,7 @@ For inputs ``x, x'`` and output dimensions ``p_x, p_{x'}'``, the kernel is defin
 k\big((x, p_x), (x, p_{x'})\big) = H_{p_{x}}K(x, x')H_{p_{x'}}
 ```
 where ``K(x, x') = Diag(k_1(x, x'), ..., k_m(x, x'))`` with zero off-diagonal entries.
+``H_{p_{x}}`` is the ``p_x``-th column of ``H`` and
 ``k_1, \ldots, k_m`` are ``m`` kernels, one for each latent process, ``H`` is a
 mixing matrix of ``m`` basis vectors spanning the output space.
 
@@ -30,44 +31,9 @@ struct NaiveLMMMOKernel{Tk<:AbstractVector{<:Kernel}, Th<:AbstractMatrix} <: MOK
 end
 
 function (κ::NaiveLMMMOKernel)((x, px)::Tuple{Any,Int}, (y, py)::Tuple{Any,Int})
-    (px > size(κ.H, 1) || py > size(κ.H, 1) || px < 1 || py < 1) &&
+    (px > size(κ.H, 2) || py > size(κ.H, 2) || px < 1 || py < 1) &&
     error("`px` and `py` must be within the range of the number of outputs")
-    m = size(κ.H,2)
-    K = Diagonal(map(k -> k(x, y), κ.K))
-    return dot(H[px,:], K, H[py,:])
-end
-
-# not optimized at all, to be improved
-function kernelmatrix(k::NaiveLMMMOKernel, x::MOInput, y::MOInput)
-    x.out_dim == y.out_dim || error("`x` and `y` should have the same output dimension")
-    x.out_dim == size(k.H, 1) ||
-        error("Mixing matrix not compatible with the given multi-output inputs")
-    m = size(k.H, 2)
-    n_x = size(x.x, 1)
-    n_y = size(y.x, 1)
-    Σ = Matrix(undef, n_x, n_y)
-    for i in 1:n_x
-        for j in i:n_y
-            K_xy = Matrix(undef, m, m)
-            K_xy .= 0
-            if x.x isa Union{ColVecs,RowVecs}
-                x′ = x.x.X[i]
-            else
-                x′ = x.x[i]
-            end
-            if y.x isa Union{ColVecs,RowVecs}
-                y′ = y.x.X[j]
-            else
-                y′ = y.x[j]
-            end
-                for h in 1:m
-                    K_xy[h, h] = k.K[h](x′, y′)
-                end
-            Σ[i,j] = k.H * K_xy * k.H'
-            Σ[j,i] = Σ[i,j]
-        end
-    end
-    return Σ
+    return sum(H[i, px] * κ.K[i](x, y) * H[i, py] for i in 1:length(κ.K))
 end
 
 function Base.show(io::IO, k::NaiveLMMMOKernel)
--- a/src/mokernels/lmm.jl
+++ b/src/mokernels/lmm.jl
@@ -17,12 +17,14 @@ mixing matrix of ``m`` basis vectors spanning the output space.
 
 [^BPTHST]: Wessel P. Bruinsma, Eric Perim, Will Tebbutt, J. Scott Hosking, Arno Solin, Richard E. Turner (2020). [Scalable Exact Inference in Multi-Output Gaussian Processes](https://arxiv.org/pdf/1911.06287.pdf).
 """
-struct LinearMixingModelKernel{Tk<:AbstractVector{<:Kernel}, Th<:AbstractMatrix} <: MOKernel
+struct LinearMixingModelKernel{Tk<:AbstractVector{<:Kernel},Th<:AbstractMatrix} <: MOKernel
     K::Tk
     H::Th
 end
 
-LinearMixingModelKernel(k::Kernel, H::AbstractMatrix) = LinearMixingModelKernel(Fill(k, size(H, 1)), H)
+function LinearMixingModelKernel(k::Kernel, H::AbstractMatrix)
+    return LinearMixingModelKernel(Fill(k, size(H, 1)), H)
+end
 
 function (κ::LinearMixingModelKernel)((x, px)::Tuple{Any,Int}, (y, py)::Tuple{Any,Int})
     (px > size(κ.H, 2) || py > size(κ.H, 2) || px < 1 || py < 1) &&
--- a/src/mokernels/lmm.jl
+++ b/src/mokernels/lmm.jl
@@ -1,5 +1,5 @@
 @doc raw"""
-    NaiveLinearMixingModelKernel(g, e::MOKernel, A::AbstractMatrix)
+    LinearMixingModelKernel(g, e::MOKernel, A::AbstractMatrix)
 
 Kernel associated with the linear mixing model.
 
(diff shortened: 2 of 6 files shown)


Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "KernelFunctions"
 uuid = "ec8451be-7e33-11e9-00cf-bbf324bd1392"
-version = "0.10.5"
+version = "0.11.5"
```
devmotion, line 3: Probably this should be
```suggestion
version = "0.10.6"
```
since it is a non-breaking new feature :slightly_smiling_face:

src/mokernels/lmm.jl
```diff
@@ -0,0 +1,86 @@
...
+For inputs ``x, x'`` and output dimensions ``p_x, p_{x'}'``, the kernel is defined as[^BPTHST]
+```math
+k\big((x, p_x), (x, p_{x'})\big) = H_{p_{x}}k_{simple}(x, x')H_{p_{x'}}
+```
+where ``k_1, \ldots, k_m`` are ``m`` kernels, one for each latent process, ``H`` is a
+mixing matrix of ``m`` basis vectors spanning the output space, and ``σ²`` is noise.
+
+[^BPTHST]: Wessel P. Bruinsma, Eric Perim, Will Tebbutt, J. Scott Hosking, Arno Solin, Richard E. Turner (2020). [Scalable Exact Inference in Multi-Output Gaussian Processes](https://arxiv.org/pdf/1911.06287.pdf).
+"""
+struct NaiveLMMMOKernel{Tk<:Union{Kernel, Vector{<:Kernel}}, Th<:AbstractMatrix, F<:Float64} <: MOKernel
```
devmotion, line 17: I think it would be easier to work with if we only use `Tk<:AbstractVector{<:Kernel}`. If it is the same kernel, one could use `Fill(kernel, m)`.

src/mokernels/lmm.jl
```diff
@@ -0,0 +1,86 @@
...
+    K::Tk
+    H::Th
+    σ²::F
+    # if just a simple kernel is provided, we construct the MO kernel
+    function NaiveLMMMOKernel(k::SimpleKernel, H::AbstractMatrix, σ²)
+        σ² >= 0 || error("`σ²` must be positive")
+        m = size(H,2)
+        K = [k for i in 1:m]
+        return new{typeof(K),typeof(H),typeof(σ²)}(K,H,σ²)
+    end
```
devmotion, line 27: I guess this could just be an outer constructor that uses `Fill(k, m)`.

src/mokernels/lmm.jl
```diff
@@ -0,0 +1,86 @@
...
+    function NaiveLMMMOKernel(K::Vector{<:Kernel}, H::AbstractMatrix, σ²)
```
devmotion, line 28: I guess it would be OK to allow `AbstractVector{<:Kernel}` (and in fact needed for `Fill(kernel, m)`).

devmotion (commented):
Thanks for the PR, I just added some initial comments :slightly_smiling_face:

src/mokernels/lmm.jl
```diff
@@ -0,0 +1,86 @@
+@doc raw"""
+    NaiveLMMMOKernel(g, e::MOKernel, A::AbstractMatrix)
+
+Kernel associated with the linear mixing model.
+
+# Definition
+
+For inputs ``x, x'`` and output dimensions ``p_x, p_{x'}'``, the kernel is defined as[^BPTHST]
+```math
+k\big((x, p_x), (x, p_{x'})\big) = H_{p_{x}}k_{simple}(x, x')H_{p_{x'}}
```
devmotion, line 10: What is `k_{simple}`?

devmotion: And maybe we could use a notation that indicates more clearly that `H_{p_x}` is the `p_x`th column (I guess we want them to be organized in columns?).

src/mokernels/lmm.jl
```diff
@@ -0,0 +1,86 @@
...
+function (κ::NaiveLMMMOKernel)((x, px)::Tuple{Any,Int}, (y, py)::Tuple{Any,Int})
+    (px > size(κ.H, 1) || py > size(κ.H, 1) || px < 1 || py < 1) &&
+    error("`px` and `py` must be within the range of the number of outputs")
+    m = size(κ.H,2)
+    K = Matrix(undef, m, m)
+    K .= 0
+    for i in 1:m
+        K[i, i] = κ.K[i](x, y)
+    end
+    return H[px,:]' * K * H[py,:]
```
devmotion, line 43: I assume it is more efficient to extract columns of `H`. Also it seems this could be computed without allocating `K`? Isn't it just
```julia
sum(κ.H[i, px] * κ.H[i, py] * κ.K[i](x, y) for i in 1:m) 
```
(if we organize `H` by columns)?

src/mokernels/lmm.jl
```diff
@@ -0,0 +1,86 @@
...
+```
+where ``k_1, \ldots, k_m`` are ``m`` kernels, one for each latent process, ``H`` is a
+mixing matrix of ``m`` basis vectors spanning the output space, and ``σ²`` is noise.
+
+[^BPTHST]: Wessel P. Bruinsma, Eric Perim, Will Tebbutt, J. Scott Hosking, Arno Solin, Richard E. Turner (2020). [Scalable Exact Inference in Multi-Output Gaussian Processes](https://arxiv.org/pdf/1911.06287.pdf).
+"""
+struct NaiveLMMMOKernel{Tk<:Union{Kernel, Vector{<:Kernel}}, Th<:AbstractMatrix, F<:Float64} <: MOKernel
+    K::Tk
+    H::Th
+    σ²::F
```
devmotion, line 20: I guess it would be simpler to not include `σ²`.

src/mokernels/lmm.jl
st--: Things like Fill are why I want to spend more time coding in Julia :blue_heart: Code can be more efficient if you know it's all the same kernel (identical hyperparameters etc.) and it can just be another array at the same time.

thomasgudjonwright: Ahh good call :)

src/mokernels/lmm.jl
devmotion: No, I meant that you want to access by column and not by row (as it seems to be the case in this draft when you write `H[px, :]`).

thomasgudjonwright: Oh so redefine `H` so that `new_H[:,px] == old_H[px,:]`? Is it faster to access cols of `Matrix` types rather than rows generally?

devmotion: Yes. The reason is that Julia uses a column-major order memory layout, i.e., all elements in a column are next to each other in memory whereas this is not the case for rows: https://docs.julialang.org/en/v1/manual/performance-tips/index.html#man-performance-column-major

thomasgudjonwright: @willtebbutt do you think this is okay? It goes against the definition of the mixing matrix in the paper, blog posts, etc... but sounds like it could improve efficiency. Is it alright to do that if we document the difference?

src/mokernels/lmm.jl
thomasgudjonwright: > What is `k_{simple}`?

Thanks for the comment, rewording this to be clear

> And maybe we could use a notation that indicates more clearly that `H_{p_x}` is the `p_x`th column (I guess we want them to be organized in columns?).

This I will address after we sort out [the comment below](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/304#discussion_r652950788)

src/mokernels/lmm.jl
theogf: Yeah the noise needs to be treated by other packages like `AbstractGPs`

src/mokernels/lmm.jl
```diff
@@ -0,0 +1,87 @@
...
+    function NaiveLMMMOKernel(K::AbstractVector{<:Kernel}, H::AbstractMatrix, σ²)
+        σ² >= 0 || error("`σ²` must be positive")
+        return new{typeof(K),typeof(H),typeof(σ²)}(K,H,σ²)
+    end
+end
+
+function (κ::NaiveLMMMOKernel)((x, px)::Tuple{Any,Int}, (y, py)::Tuple{Any,Int})
+    (px > size(κ.H, 1) || py > size(κ.H, 1) || px < 1 || py < 1) &&
+    error("`px` and `py` must be within the range of the number of outputs")
+    m = size(κ.H,2)
```
theogf, line 38:
```suggestion
    m = size(κ.H, 2)
```

src/mokernels/lmm.jl
```diff
@@ -0,0 +1,87 @@
...
+    K = Matrix(undef, m, m)
+    K .= 0
```
theogf, line 40:
```suggestion
    K = zeros(m, m)
```

src/mokernels/lmm.jl
```diff
@@ -0,0 +1,87 @@
...
+    for i in 1:m
+        K[i, i] = κ.K[i](x, y)
+    end
+    return H[px,:]' * K * H[py,:] + σ²
```
theogf, line 44:
```suggestion
    return H[px, :]' * K * H[py, :] + σ²
```

src/mokernels/lmm.jl
```diff
@@ -0,0 +1,87 @@
...
+end
+
+# not optimized at all
+function kernelmatrix(k::NaiveLMMMOKernel, x::MOInput, y::MOInput)
+    x.out_dim == y.out_dim || error("`x` and `y` should have the same output dimension")
+    x.out_dim == size(k.H, 1) ||
+        error("Mixing matrix not compatible with the given multi-output inputs")
+    m = size(k.H,2)
+    n_x = size(x.x,1)
+    n_y = size(y.x,1)
```
theogf, line 54:
```suggestion
    m = size(k.H, 2)
    n_x = size(x.x, 1)
    n_y = size(y.x, 1)
```

src/mokernels/lmm.jl
```diff
@@ -0,0 +1,87 @@
...
+    Σ = Matrix(undef, n_x, n_y)
+    for i in 1:n_x
+        for j in i:n_y
+            K_xy = Matrix(undef, m, m)
+            K_xy .= 0
```
theogf, line 59:
```suggestion
            K_xy = zeros(m, m)
```

src/mokernels/lmm.jl
```diff
@@ -0,0 +1,87 @@
...
+            if x.x isa Union{ColVecs,RowVecs}
+                x′ = x.x.X[i]
```
theogf, line 61: Are you sure about this? I think you are calling the i-th element of the underlying matrix of a `ColVecs/RowVecs`

theogf (commented):
Just added some style comments and also an issue with the `kernelmatrix` computation.

src/mokernels/lmm.jl
devmotion: Regarding efficiency, it would also be good to not preallocate `K` since it is not concretely typed and e.g. just use `K = Diagonal(map(k -> k(x, y), k.K))` and `result = dot(H[:, px], K, H[:, py])`.

It would be interesting to benchmark this approach against one without allocations as suggested above (benchmarking should be done with BenchmarkTools to avoid compile time overhead).

src/mokernels/lmm.jl
thomasgudjonwright: done

src/mokernels/lmm.jl
thomasgudjonwright: done!

src/mokernels/lmm.jl
theogf: Oh yeah I missed that,  you don't even need to build a `Diagonal`, you could just do:
`K = map(k -> k(x, y), k.K); res = dot(H[:, px], K .* H[:, py])`

src/mokernels/lmm.jl
theogf: This comment is actually pointless given https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/304#discussion_r653598389

src/mokernels/lmm.jl
devmotion: The main point of the Diagonal would be that it allows you to avoid the additional allocation in `K .* H[:, py]`.

theogf: `K .*= H[:, py]` :trollface: 
But yeah I agree the `Diagonal` solution is anyway more elegant

src/mokernels/lmm.jl
thomasgudjonwright: Done!

src/mokernels/lmm.jl
thomasgudjonwright: > Regarding efficiency, it would also be good to not preallocate `K` since it is not concretely typed and e.g. just use `K = Diagonal(map(k -> k(x, y), k.K))` and `result = dot(H[:, px], K, H[:, py])`.
> 
> It would be interesting to benchmark this approach against one without allocations as suggested above (benchmarking should be done with BenchmarkTools to avoid compile time overhead).

Changed to match this, however still by row until it has been decided to redefine `H`

devmotion: I did some benchmarks with column-wise organization of `H` and the allocation-free `sum` was always the fastest:

```julia
julia> using BenchmarkTools, KernelFunctions, LinearAlgebra

julia> function f(H, kernels, x, px, y, py)
           K = map(k -> k(x, y), kernels)
           return dot(H[:, px], Diagonal(K), H[:, py])
       end

julia> function g(H, kernels, x, px, y, py)
           return sum(H[i, px] * kernels[i](x, y) * H[i, py] for i in 1:length(kernels))
       end

julia> px, py = rand(1:10, 2);

julia> x = rand(5);

julia> y = rand(5);

julia> H = rand(20, 10);

julia> kernels = map(x -> SqExponentialKernel() ∘ ScaleTransform(x), rand(20));

julia> @benchmark f($H, $kernels, $x, $px, $y, $py)
BenchmarkTools.Trial: 
  memory estimate:  720 bytes
  allocs estimate:  3
  --------------
  minimum time:     481.605 ns (0.00% GC)
  median time:      647.010 ns (0.00% GC)
  mean time:        780.019 ns (2.03% GC)
  maximum time:     10.860 μs (88.82% GC)
  --------------
  samples:          10000
  evals/sample:     195

julia> @benchmark g($H, $kernels, $x, $px, $y, $py)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     370.734 ns (0.00% GC)
  median time:      476.973 ns (0.00% GC)
  mean time:        544.421 ns (0.00% GC)
  maximum time:     1.564 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     203

julia> H = rand(50, 10);

julia> kernels = map(x -> SqExponentialKernel() ∘ ScaleTransform(x), rand(50));

julia> @benchmark f($H, $kernels, $x, $px, $y, $py)
BenchmarkTools.Trial: 
  memory estimate:  1.45 KiB
  allocs estimate:  3
  --------------
  minimum time:     1.003 μs (0.00% GC)
  median time:      1.395 μs (0.00% GC)
  mean time:        1.648 μs (1.91% GC)
  maximum time:     128.966 μs (95.61% GC)
  --------------
  samples:          10000
  evals/sample:     10

julia> @benchmark g($H, $kernels, $x, $px, $y, $py)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     894.778 ns (0.00% GC)
  median time:      1.019 μs (0.00% GC)
  mean time:        1.139 μs (0.00% GC)
  maximum time:     3.713 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     45
```

devmotion: The same can be observed if `H` is organized in rows:
```julia
julia> function f(H, kernels, x, px, y, py)
           K = map(k -> k(x, y), kernels)
           return dot(H[px, :], Diagonal(K), H[py, :])
       end

julia> function g(H, kernels, x, px, y, py)
           return sum(H[px, i] * kernels[i](x, y) * H[py, i] for i in 1:length(kernels))
       end

julia> H = rand(10, 20);

julia> kernels = map(x -> SqExponentialKernel() ∘ ScaleTransform(x), rand(20));

julia> @benchmark f($H, $kernels, $x, $px, $y, $py)
BenchmarkTools.Trial: 
  memory estimate:  720 bytes
  allocs estimate:  3
  --------------
  minimum time:     481.579 ns (0.00% GC)
  median time:      566.056 ns (0.00% GC)
  mean time:        650.247 ns (2.01% GC)
  maximum time:     7.305 μs (85.99% GC)
  --------------
  samples:          10000
  evals/sample:     195

julia> @benchmark g($H, $kernels, $x, $px, $y, $py)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     373.898 ns (0.00% GC)
  median time:      441.398 ns (0.00% GC)
  mean time:        487.026 ns (0.00% GC)
  maximum time:     1.389 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     206

julia> H = rand(10, 50);

julia> kernels = map(x -> SqExponentialKernel() ∘ ScaleTransform(x), rand(50));

julia> @benchmark f($H, $kernels, $x, $px, $y, $py)
BenchmarkTools.Trial: 
  memory estimate:  1.45 KiB
  allocs estimate:  3
  --------------
  minimum time:     1.022 μs (0.00% GC)
  median time:      1.382 μs (0.00% GC)
  mean time:        1.590 μs (2.20% GC)
  maximum time:     141.293 μs (96.39% GC)
  --------------
  samples:          10000
  evals/sample:     10

julia> @benchmark g($H, $kernels, $x, $px, $y, $py)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     905.125 ns (0.00% GC)
  median time:      1.069 μs (0.00% GC)
  mean time:        1.169 μs (0.00% GC)
  maximum time:     3.808 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     32
```

The timings for the matrix version are very similar (no clear difference) but for the faster `sum` version the results are better if `H` is organized by columns.

thomasgudjonwright: Thank you for doing this :) ! I don't think it will be a problem switching to the col `H`, but will ping Will to confirm today and hopefully get that change in

willtebbutt: > @willtebbutt do you think this is okay? It goes against the definition of the mixing matrix in the paper, blog posts, etc... but sounds like it could improve efficiency. Is it alright to do that if we document the difference?

Yeah, that's totally fine :)

src/mokernels/lmm.jl
thomasgudjonwright: I think you are right, I think for now I will just remove the `kernelmatrix` function as it seems defining it for any given kernel is just for optimization sake and this is the naive LMM so it wouldn't be optimized anyways (so it will be defined for the LMM which I plan to add next)

theogf: I think you should have a look at the implementation of another multi-output kernel : https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/master/src/mokernels/slfm.jl
Maybe this can help

thomasgudjonwright: Yeah I saw that, it breaks the cov matrix into its components nicely. I can see if theres a nice way of doing so with this covariance matrix when implementing the more efficient LMM :)

thomasgudjonwright: Do you agree that if the function isn't actually any more efficient that its worth removing for this one though (it is the naive LMM after all)?

src/mokernels/lmm.jl
thomasgudjonwright: @devmotion is there a specific way to indicate the meaning of `H_{p_x}` purely through its notation that you can think of or should I just add a note specifying that it refers to the `p_x`th col?

src/mokernels/lmm.jl
thomasgudjonwright: Updated in the most recent commit (forgot to mention it in the commit message)

src/mokernels/lmm.jl
```diff
@@ -0,0 +1,46 @@
...
+    # if just a simple kernel is provided, we construct the MO kernel
+    function NaiveLMMMOKernel(k::SimpleKernel, H::AbstractMatrix)
+        m = size(H,2)
+        K = fill(k, m)
+        return new{typeof(K),typeof(H)}(K,H)
+    end
+    function NaiveLMMMOKernel(K::AbstractVector{<:Kernel}, H::AbstractMatrix)
+        return new{typeof(K),typeof(H)}(K,H)
+    end
+end
```
devmotion, line 31: I guess this could just be one outer constructor (the other one is just the default constructor)? And does not have to be restricted to `SimpleKernel`? And one could use `Fill` instead which does not allocate?

```suggestion
end

NaiveLMMMOKernel(k::Kernel, H::AbstractMatrix) = NaiveLMMMOKernel(Fill(k, size(H, 1)), H)
```

test/mokernels/lmm.jl
```diff
@@ -0,0 +1,44 @@
...
+    rng = MersenneTwister(123)
+    FDM = FiniteDifferences.central_fdm(5, 1)
+    N = 10
+    in_dim = 3
+    out_dim = 6
+    x1 = MOInput([rand(rng, in_dim) for _ in 1:N], out_dim)
+    x2 = MOInput([rand(rng, in_dim) for _ in 1:N], out_dim)
+    H = rand(4,6)
+
+    k = NaiveLMMMOKernel(
```
devmotion, line 11: Can you add some tests for the version with a single kernel as well?

src/mokernels/lmm.jl
```diff
@@ -0,0 +1,46 @@
+@doc raw"""
+    NaiveLMMMOKernel(g, e::MOKernel, A::AbstractMatrix)
```
nickrobinson251, line 2: Does it make sense to add this as a separate `NaiveLMM` kernel, distinct from the "optimised" `LMM` kernel?

Or should it just be one kernel, the implementation of which we improve over time?

Is there something about the "optimised" version that's different? Is there a reason anyone would ever prefer to use "NaiveLMM" over the "optimised" version?

src/mokernels/lmm.jl
```diff
@@ -0,0 +1,46 @@
...
+k\big((x, p_x), (x, p_{x'})\big) = H_{p_{x}}K(x, x')H_{p_{x'}}
+```
+where ``K(x, x') = Diag(k_1(x, x'), ..., k_m(x, x'))`` with zero off-diagonal entries.
+``H_{p_{x}}`` is the ``p_x``-th column of ``H`` and
+``k_1, \ldots, k_m`` are ``m`` kernels, one for each latent process, ``H`` is a
+mixing matrix of ``m`` basis vectors spanning the output space.
+
+[^BPTHST]: Wessel P. Bruinsma, Eric Perim, Will Tebbutt, J. Scott Hosking, Arno Solin, Richard E. Turner (2020). [Scalable Exact Inference in Multi-Output Gaussian Processes](https://arxiv.org/pdf/1911.06287.pdf).
+"""
+struct NaiveLMMMOKernel{Tk<:AbstractVector{<:Kernel}, Th<:AbstractMatrix} <: MOKernel
```
nickrobinson251, line 19: naming question: Do all MO Kernels need to be prefixed `MOKernel`? It seem like it'd be nicer, and unambiguous, to call this `LLMKernel` rather than `LLMMOKernel`

src/mokernels/lmm.jl
```diff
@@ -0,0 +1,46 @@
...
+        return new{typeof(K),typeof(H)}(K,H)
+    end
+    function NaiveLMMMOKernel(K::AbstractVector{<:Kernel}, H::AbstractMatrix)
+        return new{typeof(K),typeof(H)}(K,H)
+    end
+end
+
+function (κ::NaiveLMMMOKernel)((x, px)::Tuple{Any,Int}, (y, py)::Tuple{Any,Int})
+    (px > size(κ.H, 2) || py > size(κ.H, 2) || px < 1 || py < 1) &&
+    error("`px` and `py` must be within the range of the number of outputs")
```
nickrobinson251, line 35: do we want to throw a more specific error here? perhaps a `DomainError`?

src/mokernels/lmm.jl
```diff
@@ -0,0 +1,46 @@
...
+    return sum(H[i, px] * κ.K[i](x, y) * H[i, py] for i in 1:length(κ.K))
+end
+
+function Base.show(io::IO, k::NaiveLMMMOKernel)
+    return print(io, "Linear Mixing Model Multi-Output Kernel (naive implementation)")
+end
```
nickrobinson251, line 41: i don't think we want to overload 2-arg show here, only the 3-arg plain-text method below.


The reason is that we probably want `repr` to return parseable code that recreates a NaiveLMMMOKernel (other kernels in this package seem to have overloaded 2-arg show to have non-parseable "pretty printing" output, which also seems a mistake to me)

And 3-arg show is what gets used in the repl, so probably all you want for pretty-printing. 

(But my opinion here might be at adds with the currect practices of the package, so we can move this discussion to an issue if that's better)

src/mokernels/lmm.jl
```diff
@@ -0,0 +1,46 @@
...
+
+function Base.show(io::IO, ::MIME"text/plain", k::NaiveLMMMOKernel)
+    print(io, "Linear Mixing Model Multi-Output Kernel (naive implementation)\n\tkernels (K): ")
+    return join(io, k.K, "\n\t\t")
+end
```
nickrobinson251, line 46: I don't think this is quite right as-is. e.g.
```jl
julia> lmm = KernelFunctions.NaiveLMMMOKernel([Matern32Kernel(), SqExponentialKernel()], rand(2,3))
Linear Mixing Model Multi-Output Kernel (naive implementation)
        kernels (K): Matern 3/2 Kernel (metric = Distances.Euclidean(0.0))
                Squared Exponential Kernel (metric = Distances.Euclidean(0.0))
```
perhaps we want something like
```jl
Linear Mixing Model Multi-Output Kernel (naive implementation). Kernels:
        Matern 3/2 Kernel (metric = Distances.Euclidean(0.0))
        Squared Exponential Kernel (metric = Distances.Euclidean(0.0))
```
Also, we should probably call `show` on the kernels, rather than implicitly string-ifying them via `join`:
```suggestion
function Base.show(io::IO, mime::MIME"text/plain", k::NaiveLMMMOKernel)
    print(io, "Linear Mixing Model Multi-Output Kernel (naive implementation). Kernels:")
    for k in k.K
        print(io, "\n\t")
        show(io, mime, k)
    end
end
```

src/mokernels/lmm.jl
```diff
@@ -0,0 +1,46 @@
...
+mixing matrix of ``m`` basis vectors spanning the output space.
+
+[^BPTHST]: Wessel P. Bruinsma, Eric Perim, Will Tebbutt, J. Scott Hosking, Arno Solin, Richard E. Turner (2020). [Scalable Exact Inference in Multi-Output Gaussian Processes](https://arxiv.org/pdf/1911.06287.pdf).
+"""
+struct NaiveLMMMOKernel{Tk<:AbstractVector{<:Kernel}, Th<:AbstractMatrix} <: MOKernel
+    K::Tk
+    H::Th
+    # if just a simple kernel is provided, we construct the MO kernel
+    function NaiveLMMMOKernel(k::SimpleKernel, H::AbstractMatrix)
+        m = size(H,2)
```
nickrobinson251, line 24:
```suggestion
        m = size(H, 2)
```

test/mokernels/lmm.jl
```diff
@@ -0,0 +1,44 @@
...
+    @test string(k) == "Linear Mixing Model Multi-Output Kernel (naive implementation)"
+    @test repr("text/plain", k) == (
+        "Linear Mixing Model Multi-Output Kernel (naive implementation)\n" *
+        "\tkernels (K): Matern 3/2 Kernel (metric = Euclidean(0.0))\n" *
+        "\t\tSquared Exponential Kernel (metric = Euclidean(0.0))\n" *
+        "\t\tFractional Brownian Motion Kernel (h = 0.5)\n" *
+        "\t\tMatern 3/2 Kernel (metric = Euclidean(0.0))"
+    )
+
+    # # AD test
```
nickrobinson251, line 29: i'm not familiar with how AD compatibility is tested in this package. How can i learn about that? And what's the plan for addressing this in the PR?

src/mokernels/lmm.jl
devmotion: I think `LinearMixingModelKernel` might be even better (could still add an alias if people prefer it but so far it seems for most kernels we don't use any abbreviations).

src/mokernels/lmm.jl
devmotion: Currently, `show` is implemented incorrectly. https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/244 was supposed to fix it but there was disagreement about how to print nested kernels.

src/mokernels/lmm.jl
devmotion: > than implicitly string-ifying them via join:

Does `join(io, ...)` actually create a string first before writing it to `io`? I assumed it would just write to `io` directly.

test/mokernels/lmm.jl
thomasgudjonwright: In terms of addressing it in this PR, it has been addressed by Will helping me sort out some issues and copying the tests from other MOKernel AD tests. Going forward I am certainly also keen on finding places to learn more on this and AD generally!

src/mokernels/lmm.jl
thomasgudjonwright: Good idea, done!

src/mokernels/lmm.jl
thomasgudjonwright: This is a good question. I have just been mirroring what we currently have but I can talk to Eric to ask about any fundamental differences between the two, as none pop out to me right away.

test/mokernels/lmm.jl
theogf: We have a AD test infrastructure where we compare Zygote, ForwardDiff and ReverseDiff vs the results of FiniteDiff.
Normally all you have to pass is a constructor and eventually parameters you want to test AD against. For inputs it's done automatically.
However for multi-output it's been added only recently.

src/mokernels/lmm.jl
thomasgudjonwright: I can certainly change it to `LinearMixingModelKernel` if everyone is okay with me breaking what appears to be convention with the `MOKernel` suffix! `LinearMixingModelKernel` or `LMMKernel` are both a lot nicer than the current name aha!

test/mokernels/lmm.jl
theogf: Oh my answer was to late... Here is the function used for example : https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/efca34f298277cf464a841a1ee8acf4e256871c1/test/mokernels/intrinsiccoregion.jl#L24
And here is where it's defined : https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/efca34f298277cf464a841a1ee8acf4e256871c1/test/test_utils.jl#L258

src/mokernels/lmm.jl
thomasgudjonwright: I was following what had been done before ([eg.](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/master/src/mokernels/slfm.jl#L26)) but am happy to be more specific

src/mokernels/lmm.jl
thomasgudjonwright: Done :)

test/mokernels/lmm.jl
thomasgudjonwright: Done

src/mokernels/lmm.jl
thomasgudjonwright: So my suspicions were confirmed that there are two benefits to why the naive version is useful:
1) You can see what is going on a bit more intuitively, understanding the methods behind the madness
2) It is less error prone. This point should hopefully be made false as a result of the work over the next month or so

test/mokernels/lmm.jl
thomasgudjonwright: Hey @theogf just saw this. I can certainly use this and also add it to the `slfm` tests, since it isn't used there (checked and there is already an issue for that). An issue I am running into is that `ForwardDiff` doesn't seem too happy about what is going on, although I am not sure exactly what is the underling problem. This happens for both the `slfm` and `lmm` kernels:
```
julia> test_ADs(k)
Test Summary:     | Pass  Total
FiniteDifferences |    7      7
Test Summary: | Pass  Total
Zygote        |    8      8
ForwardDiff: Test Failed at /Users/thomaswright/Documents/Dev_Folder/KernelFunctions.jl/test/test_utils.jl:48
  Expression: ≈(grad_AD, grad_FD, atol = 1.0e-8, rtol = 1.0e-5)
   Evaluated: [NaN NaN NaN; NaN NaN NaN; NaN NaN NaN] ≈ [0.15591176360047018 -1.0599515731543765 2.3445727403930037; 0.5225644767555176 2.7541574740852086 -0.27921885622899056; -0.6370212637461734 0.40638576059219833 -0.8773988979307216] (atol=1.0e-8, rtol=1.0e-5)
Stacktrace:
 [1] compare_gradient(f::Function, AD::Symbol, args::Matrix{Float64})
   @ Main ~/Documents/Dev_Folder/KernelFunctions.jl/test/test_utils.jl:48
 [2] macro expansion
   @ ~/Documents/Dev_Folder/KernelFunctions.jl/test/test_utils.jl:277 [inlined]
 [3] macro expansion
   @ /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/Test/src/Test.jl:1151 [inlined]
 [4] test_AD(AD::Symbol, k::LatentFactorMOKernel{Vector{Kernel}, IndependentMOKernel{SqExponentialKernel{Euclidean}}, Matrix{Float64}}, dims::NamedTuple{(:in, :out, :obs), Tuple{Int64, Int64, Int64}})
   @ Main ~/Documents/Dev_Folder/KernelFunctions.jl/test/test_utils.jl:260
ForwardDiff: Test Failed at /Users/thomaswright/Documents/Dev_Folder/KernelFunctions.jl/test/test_utils.jl:48
  Expression: ≈(grad_AD, grad_FD, atol = 1.0e-8, rtol = 1.0e-5)
   Evaluated: [NaN NaN NaN; NaN NaN NaN; NaN NaN NaN] ≈ [-0.10213130841060664 0.3950676414773004 0.18724131054721305; 0.03403101998124971 0.42620309981639404 0.5389335784071836; -0.575814364979744 0.0736449167318202 0.13282464788514278] (atol=1.0e-8, rtol=1.0e-5)
Stacktrace:
 [1] compare_gradient(f::Function, AD::Symbol, args::Matrix{Float64})
   @ Main ~/Documents/Dev_Folder/KernelFunctions.jl/test/test_utils.jl:48
 [2] macro expansion
   @ ~/Documents/Dev_Folder/KernelFunctions.jl/test/test_utils.jl:289 [inlined]
 [3] macro expansion
   @ /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/Test/src/Test.jl:1151 [inlined]
 [4] test_AD(AD::Symbol, k::LatentFactorMOKernel{Vector{Kernel}, IndependentMOKernel{SqExponentialKernel{Euclidean}}, Matrix{Float64}}, dims::NamedTuple{(:in, :out, :obs), Tuple{Int64, Int64, Int64}})
   @ Main ~/Documents/Dev_Folder/KernelFunctions.jl/test/test_utils.jl:260
Test Summary: | Pass  Fail  Total
ForwardDiff   |    6     2      8
ERROR: Some tests did not pass: 6 passed, 2 failed, 0 errored, 0 broken.
```

devmotion: There are some issues for some kernels such as ExponentialKernel and possible MaternKernels if you evaluate them with the same arguments. Which kernels did you try? IIRC SqExponentialKernel does not have these issues.

theogf: `Matern32Kernel` should be fine though

src/mokernels/lmm.jl
nickrobinson251: does this suggest adding a single `LMM` kernel, and making PRs to update it over time? (since we should be able to achieve the same level of robustness, just with speed/memory improvements)?

test/mokernels/lmm.jl
thomasgudjonwright: Should we not only introduce this test into the test suite once it is reliable for all `SimpleKernel`s that could be used?

src/mokernels/lmm.jl
devmotion: Maybe `H_{:,p_x}`?

test/mokernels/lmm.jl
devmotion: Since the main point here is not to test specific combinations of simple kernels but to ensure that the implementation of LMM is AD-compatible we can just choose a combination that works.

theogf: Yeah we should assume that if the AD tests for the simple kernels work, their combination should work as well

thomasgudjonwright: Done

src/mokernels/lmm.jl
thomasgudjonwright: If we can achieve parity in robustness with the optimized versions, all we lose with updating the single LMM is interpretability. Is there any downside to having two LMMs existing in `mokernel/lmm.jl`, the `NaiveLMM` as well as the `LMM`?

src/mokernels/lmm.jl
nickrobinson251: `LinearMixingModelKernel` sounds good to me!
perhaps with `const LMMKernel = LinearMixingModelKernel` too

src/mokernels/lmm.jl
nickrobinson251: cool -- i think then we can go with a single `LMMKernel` (drop the `Naive` prefix), and add optimisations to this single kernel

thomasgudjonwright: Thats true. I suppose adding the optimized version will mainly be adding a dispatch for `kernelmatrix` so it makes sense to just have the one type!

thomasgudjonwright: Done

src/mokernels/lmm.jl
thomasgudjonwright: Done :)

src/mokernels/lmm.jl
thomasgudjonwright: Is everyone happy with `show` as it stands?

src/mokernels/lmm.jl
thomasgudjonwright: For now I will leave it as is, maybe we could open an issue about making errors more specific?

src/mokernels/lmm.jl
thomasgudjonwright: Is everyone happy with its current state?

src/mokernels/lmm.jl
willtebbutt: I'm happy for this to go in as-is, provided that either a comment is added to that PR, or an issue opened so that we don't forget about it.

willtebbutt: Resolving in favour of #307

src/mokernels/lmm.jl
```diff
@@ -0,0 +1,52 @@
...
+end
+
+## Aliases ##
+
+"""
+    LMMKernel()
+
+Alias of [`LinearMixingModelKernel`](@ref).
+"""
+const LMMKernel = LinearMixingModelKernel
```
devmotion, line 52: I wonder if the alias is needed?

devmotion (commented):
Looks good. I wonder if we need the alias? Could you also update the documentation and add it to the list of multi-output kernels?

src/mokernels/lmm.jl
thomasgudjonwright: Are there any downsides to adding an alias? `LinearMixingModelKernel` is quite long.

src/mokernels/lmm.jl
nickrobinson251: > Does join(io, ...) actually create a string first before writing it to io?

i looked into it and no, it just calls `print(io, ...)` which writes straight to `io`. 

But what i meant to say is that we probably want to call 3-arg `show` on the inner kernels, rather than using `join`/`print` which are 2-arg show.

Sometimes words are harder that code 😂 sorry for the confusion. 

Let us know if you think it should be as it was before, @devmotion 

(p.s. `show`/`print` stuff is some of the most confusing stuff in all Julia-land 😵)

nickrobinson251:
This LGTM, Tom! i'll leave it to David/Will/Theo to do any final review/approval.

And thanks all for getting this PR up to scratch so fast!

src/mokernels/lmm.jl
thomasgudjonwright: I will assume this is good as is (for the sake of this PR)

thomasgudjonwright:
@devmotion @theogf @willtebbutt any more thoughts or concerns you want to discuss or have addressed?

src/mokernels/lmm.jl
devmotion: It increases the public API that can possibly be broken, increases complexity slightly since users and developers have to be aware of the alias in the first place, and could create some confusion about what the official/recommended way is for creating such kernels. However, I can see the benefit of a shorter form, so if everyone agrees that it is beneficial I am fine with it. It should just be considered and discussed in this PR since it is easy to add an alias later on but more difficult to remove it (requires deprecations and a breaking release).

thomasgudjonwright: @nickrobinson251 @willtebbutt what do you two think? The acronym is whats mainly used internally, do you see some frustration in only having the whole `LinearMixingModelKernel` available as a name?

thomasgudjonwright: @devmotion good point about easier to add than to remove later on, not something I had thought about :)

willtebbutt: I can't see there being a problem with `LinearMixingModelKernel` to be honest, so I'm happy to dispense with the alias.

willtebbutt (dismissed):
I'm happy :)

src/mokernels/lmm.jl
thomasgudjonwright: Nick is off today so I will remove this, since as @devmotion said, it can always be added after!

thomasgudjonwright: Done :)

willtebbutt (approved):

devmotion (approved):

theogf (approved):

thomasgudjonwright:
Could someone please merge this as I don't have permission to do so :)

willtebbutt:
Are you able to merge in the latest version of master to your branch @thomasgudjonwright ? A PR just went in, so it looks like you need to update your branch before we can merge.

thomasgudjonwright:
@willtebbutt should be done :)

willtebbutt:
Great -- of those who have merge rights: please merge if you see the tests pass (I'm shortly going offline until Sunday)

thomasgudjonwright:
@theogf @devmotion the three failures seem unrelated to the PR/changes and appear to show up in most PRs. Please merge if you agree :)

src/mokernels/lmm.jl
```diff
@@ -0,0 +1,45 @@
...
+where ``K(x, x') = Diag(k_1(x, x'), ..., k_m(x, x'))`` with zero off-diagonal entries.
+``H_{:,p_{x}}`` is the ``p_x``-th column (`p_x`-th output) of ``H \in \mathbb{R}^{m \times p}``
+representing ``m`` basis vectors for the ``p`` dimensional output space of ``f``.
+``k_1, \ldots, k_m`` are ``m`` kernels, one for each latent process, ``H`` is a
+mixing matrix of ``m`` basis vectors spanning the output space.
+
+[^BPTHST]: Wessel P. Bruinsma, Eric Perim, Will Tebbutt, J. Scott Hosking, Arno Solin, Richard E. Turner (2020). [Scalable Exact Inference in Multi-Output Gaussian Processes](https://arxiv.org/pdf/1911.06287.pdf).
+"""
+struct LinearMixingModelKernel{Tk<:AbstractVector{<:Kernel},Th<:AbstractMatrix} <: MOKernel
+    K::Tk
```
st--, line 21: A bit late to the party given the PR's already merged now, but I was wondering about the struct elements both concretely here and across the package more generally... I don't find `K` particularly intuitive; why not call it `kernels` (or at least `ks` to make clear it's not just one kernel)?

willtebbutt: This would make sense to me. I agree that `K` it's not at all obvious that `K` represents a collection of `Kernel`s without checking its type. Would be happy to review a PR that changed this.

nickrobinson251: if retrieving this info is useful, perhaps more robust and idiomatic to have an accessor function, rather than rely on field names?
e.g.
```julia
kernels(k::LinearMixingModelKernel) = k.K
```

devmotion: IIRC `kernels` existed but we removed it since it was not used anywhere :smile:

st--: It's more about having the code being as easily understandable and self-explanatory as possible - _not_ accessing the field names directly doesn't mean the names have to be short and cryptic; quite the opposite, no?

thomasgudjonwright: I almost made it `kernels` actually! I totally agree that a more explicit name would be helpful and thats why I went inched towards that with `K` instead of `g` which is the list of kernels in the `slfm`

PR #309, opened 2021-06-24
Use processes to run literate examples

theogf:
This PR is aiming as using difference processes for each markdown file in #303 
I created a new branch in case everything goes to :hankey:...

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -17,44 +17,30 @@ const BLACKLIST = ["deep-kernel-learning", "support-vector-machine"]
 ispath(EXAMPLES_OUT) && rm(EXAMPLES_OUT; recursive=true)
 mkpath(EXAMPLES_OUT)
 
-# preprocessor for Literate example scripts:
-#  - add Documenter @setup snippet that activates each example's own project environment
-function preprocess(content)
-    sub = SubstitutionString("""
-                     \\0
-                     #
-                     #md #
-                     #md # ```@setup @__NAME__
-                     #md # using Pkg: Pkg
-                     #md # Pkg.activate("$(EXAMPLES_SRC)/@__NAME__")
-                     #md # Pkg.instantiate()
-                     #md # ```
-                     #
-                             """)
-    return replace(content, r"^# # [^\n]*"m => sub; count=1)
-end
-
+cmd = Base.julia_cmd()
 for example in readdir(EXAMPLES_SRC)
     example ∈ BLACKLIST && continue
     exampledir = joinpath(EXAMPLES_SRC, example)
     isdir(exampledir) || continue
-    Pkg.activate(exampledir) do
-        Pkg.develop(; path=PACKAGE_DIR)
-        Pkg.instantiate()
-        filepath = joinpath(exampledir, "script.jl")
-        Literate.markdown(
-            filepath, EXAMPLES_OUT; name=example, documenter=true, preprocess=preprocess
-        )
-        Literate.notebook(
-            filepath,
-            EXAMPLES_OUT;
-            name=example,
-            documenter=true,
-            preprocess=preprocess,
-            execute=false,
-        )
-    end
+    filepath = joinpath(exampledir, "script.jl")
+    @show code = """using Literate; Literate.markdown("script.jl", "$(EXAMPLES_OUT)"; name=example, documenter=true, execute=true)"""
+    run(addenv(`$(cmd) -e $(code)`, Dict("JULIA_LOAD_PATH" => exampledir)))
+    Literate.notebook(
+        filepath,
+        EXAMPLES_OUT;
+        name=example,
+        documenter=true,
+        execute=false,
+    )
 end
+#    Pkg.activate(exampledir) do
+#        Pkg.develop(; path=PACKAGE_DIR)
+#        Pkg.instantiate()
+#        filepath = joinpath(exampledir, "script.jl")
+#        Literate.markdown(
+#            filepath, EXAMPLES_OUT; name=example, documenter=true, preprocess=preprocess
+#        )
+
 
 DocMeta.setdocmeta!(
     KernelFunctions,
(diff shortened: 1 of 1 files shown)


theogf:
No idea why but it seems to work.

docs/make.jl
```diff
@@ -17,44 +17,35 @@ const BLACKLIST = ["deep-kernel-learning", "support-vector-machine"]
...
-            documenter=true,
-            preprocess=preprocess,
-            execute=false,
-        )
-    end
+        using Literate
+        Literate.markdown("$(filepath)", "$(EXAMPLES_OUT)"; name="$(example)", documenter=true, execute=true)
+    """
+    run(
+        addenv(
```
devmotion, line 35: Maybe `withenv` would be a bit more idiomatic?

docs/make.jl
```diff
@@ -17,44 +17,35 @@ const BLACKLIST = ["deep-kernel-learning", "support-vector-machine"]
 ispath(EXAMPLES_OUT) && rm(EXAMPLES_OUT; recursive=true)
 mkpath(EXAMPLES_OUT)
 
-# preprocessor for Literate example scripts:
-#  - add Documenter @setup snippet that activates each example's own project environment
-function preprocess(content)
```
devmotion, line 22: I guess it would still be useful to preprocess the examples in a similar way as done in https://github.com/devmotion/CalibrationErrors.jl/blob/c73b2abe19b349586ca1448b55c7e697b814c03c/docs/make.jl#L19-L34 and https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/blob/9b14c286e5885c2e94edf5bc3502347126b23276/docs/make.jl#L16-L23, such that the documentation links to the notebook and vice versa.

docs/make.jl
```diff
@@ -17,44 +17,35 @@ const BLACKLIST = ["deep-kernel-learning", "support-vector-machine"]
...
-            name=example,
-            documenter=true,
-            preprocess=preprocess,
-            execute=false,
-        )
-    end
+        using Literate
+        Literate.markdown("$(filepath)", "$(EXAMPLES_OUT)"; name="$(example)", documenter=true, execute=true)
+    """
+    run(
```
devmotion, line 34: Can we parallelize this such that examples are executed in parallel?

docs/make.jl
```diff
@@ -17,44 +17,35 @@ const BLACKLIST = ["deep-kernel-learning", "support-vector-machine"]
...
-                     #md # Pkg.activate("$(EXAMPLES_SRC)/@__NAME__")
-                     #md # Pkg.instantiate()
-                     #md # ```
-                     #
-                             """)
-    return replace(content, r"^# # [^\n]*"m => sub; count=1)
-end
-
+cmd = Base.julia_cmd()
+docs_env = @__DIR__
```
devmotion, line 21: Do we actually need to inherit anything? Maybe it would be cleaner to just install `Literate` manually (or even better, add it to the dependencies of the example's Project.toml such that it is pinned and tracked as well) and not inherit anything.

Then one could even include the `notebook` run in the external process as well.

docs/make.jl
```diff
@@ -17,44 +17,35 @@ const BLACKLIST = ["deep-kernel-learning", "support-vector-machine"]
...
+        using Literate
+        Literate.markdown("$(filepath)", "$(EXAMPLES_OUT)"; name="$(example)", documenter=true, execute=true)
+    """
+    run(
+        addenv(
+            `$(cmd) --project=$(exampledir) -e $(code)`,
+            Dict("JULIA_LOAD_PATH" => load_path),
+        ),
+    )
+    Literate.notebook(filepath, EXAMPLES_OUT; name=example, documenter=true, execute=false)
```
devmotion, line 40: Maybe just include this in the external process as well? (see https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/309/files#r658620202).

docs/make.jl
theogf: This can eventually be added later. I think the main point right now is to have examples running

docs/make.jl
theogf: To be honest I don't fully understand what this does... I just copied the example from https://discourse.julialang.org/t/multiple-environments-in-documenter/63436/4

docs/make.jl
theogf: I guess we could! Does GH Actions allows this? I could not find anything in the docs

devmotion: I thought we could just use Julia's parallelization capabilities by e.g. spawning multiple processes in parallel. It seems a bit tricky to use GH actions for it since we would have to collect all outputs and only update the gh-pages branch once after everything was run and the documentation was built.

docs/make.jl
devmotion: It builds a stacked environment and adds both the packages in docs/ and examples/current_example/ to the load path (the latter are prioritized). However, I don't think we have to inherit anything if we add the Literate dependency to examples/current_example/. Then we could just run the external process, instantiate the example environment (without any additional special stuff), and generate both the Markdown and the notebook.

docs/make.jl
```diff
@@ -17,44 +17,32 @@ const BLACKLIST = ["deep-kernel-learning", "support-vector-machine"]
...
-            preprocess=preprocess,
-            execute=false,
-        )
+        using Literate
+        Literate.markdown("$(filepath)", "$(EXAMPLES_OUT)"; name="$(example)", documenter=true, execute=true)
+        Literate.notebook("$(filepath)", "$(EXAMPLES_OUT)"; name="$(example)", documenter=true, execute=true)
+    """
+    withenv("JULIA_LOAD_PATH" => load_path) do
+        run(`$(cmd) --project=$(exampledir) -e $(code)`)
     end
```
devmotion, line 37: Without stacked environments this is just
```suggestion
    run(`$(cmd) --project=$(exampledir) -e $(code)`)
```

theogf: Please feel free to apply the correct edits, I am moving in the dark here :sweat_smile:

st--:
This has been superseded by #314 - thanks for getting it started!

Issue #306, opened 2021-06-23
Split `MOInput` into two types


thomasgudjonwright:
The idea would be to introduce two new types (names open for discussion), lets call them `BlockByInputs` and `BlockByOutputs` for the purpose of this issue.

`BlockByOutputs` would be equivalent to the current `MOInput`, ie, given and input `x = [4.5, 5.5]`, we have:

```
> julia BlockByOutputs(x, 2)
4-element MOInput{Vector{Float64}}:
 (4.5, 1)
 (5.5, 1)
 (4.5, 2)
 (5.5, 2)
```

Conversely, `BlockByInputs` would look something like this for the same input:

```
> julia BlockByInputs(x, 2)
4-element MOInput{Vector{Float64}}:
 (4.5, 1)
 (4.5, 2)
 (5.5, 1)
 (5.5, 2)
```

The reasoning for this is that each will format the `kernelmatrix` differently. The `LMM` expects a kernelmatrix that would be constructed properly using the latter (also works by using `collect(MOInput(x,2))` as input to `kernelmatrix`).

willtebbutt:
(spoken to Tom offline about this)

To expand on what Tom is saying: there are two obvious conventions you could adopt for the ordering of your inputs to a multi-output GP. Suppose that you have `P` outputs and `N` "inputs", meaning that your complete input is of length `NP`. You can either
1. have `P` blocks of `N` (MOInput), or
2. `N` blocks of `P` (the one we're not currently doing)

Invenia's code currently utilises the latter, while JuliaGaussianProcesses presently only has support for the former.

We should support both conventions because
1. different situations calls for different conventions, in the same way as `ColVecs` and `RowVecs` are both necessary.
2. Invenia has lots of optimisations for the convention we don't currently support, which we really ought to be able to hook into.

willtebbutt:
I quite like `BlockByOutputs` and `BlockByInputs`, but since these are both collections of inputs (in our vocab), I wonder whether now is the right time to also solve the "inputs" vs "inputs" naming issue.

i.e. for
```julia
x = randn(2)
y = MOInput(x, 3)
```
both are valid "inputs", but it would be helpful to be able to distinguish between `x` and `y` when referring to inputs.

I wonder whether we should call `y` "inputs" and `x` "locations". We could then say things like:
"An input to a multi-ouptut GP comprises a _location_ `x` and output index `p`"

as opposed to having to say

[...]

PR #310, opened 2021-06-25
Adding two MOInput types

thomasgudjonwright:
Addresses #306.

This creates two types `IsotopicByFeatures` and `IsotopicByOutputs` to cover the two formats we could want our `MOInput` to take. 

Names for the supertype and subtypes are still to be confirmed once everyone agrees. Until then, I have left the supertype as `MOInput` as a temp name to minimize changes that were necessary.

There may be a better way to define the supertype and subtypes to allow `MOInput`s themselves to be used, but I saw that only abstract types can be subtyped. Let me know if theres a different approach for this.

WIP: removing supertype, updating docs


src/mokernels/moinput.jl
```diff
@@ -1,15 +1,23 @@
...
+    MOInput
 
-A data type to accomodate modelling multi-dimensional output data.
+An abstract type to accomodate modelling multi-dimensional output data. There are two
+subtypes that each specify a unique ordering of the dimensions.
 
-`MOInput(x, out_dim)` has length `length(x) * out_dim`.
+See [Inputs for Multiple Outputs](@ref) in the docs for more info.
+"""
+abstract type MOInput <: AbstractVector{Tuple{Any,Int}} end
```
willtebbutt, line 9: Perhaps rename `MOInput` to `AbstractIsotopicInput` or something like that?

willtebbutt: The reason being that a `Vector{Tuple{T, Int}} where T` is a perfectly valid input for a multi-output kernel, so I would prefer to avoid suggesting that users _must_ use a subtype of this type.

src/mokernels/moinput.jl
```diff
@@ -18,13 +26,40 @@ julia> MOInput(x, 2)
...
-As shown above, an `MOInput` represents a vector of tuples.
+As shown above, an `IsotopicByFeatures` represents a vector of tuples.
 The first `length(x)` elements represent the inputs for the first output, the second
 `length(x)` elements represent the inputs for the second output, etc.
+"""
 
-See [Inputs for Multiple Outputs](@ref) in the docs for more info.
+struct IsotopicByFeatures{T<:AbstractVector} <: MOInput
+    x::T
+    out_dim::Integer
```
willtebbutt, line 36: I'm not entirely sure why we made this field abstract. Could you make the type parametric instead?

src/mokernels/moinput.jl
```diff
@@ -18,13 +26,40 @@ julia> MOInput(x, 2)
...
+
+As shown above, an `IsotopicByOutputs` represents a vector of tuples.
+The first `out_dim` elements represent all outputs for the first input, the second
+`out_dim` elements represent the outputs for the second input, etc.
 """
-struct MOInput{T<:AbstractVector} <: AbstractVector{Tuple{Any,Int}}
+
+struct IsotopicByOutputs{T<:AbstractVector} <: MOInput
     x::T
     out_dim::Integer
```
willtebbutt, line 64: Same here re parametric typing.

src/mokernels/moinput.jl
```diff
@@ -39,11 +74,18 @@ Base.firstindex(inp::MOInput) = 1
 
 function Base.getindex(inp::MOInput, ind::Integer)
     if ind > 0
-        out_dim = ind ÷ length(inp.x) + 1
-        ind = ind % length(inp.x)
-        if ind == 0
-            ind = length(inp.x)
-            out_dim -= 1
+        if inp isa IsotopicByOutputs
```
willtebbutt, line 77: Probably better to use dispatch here that than an if / else. i.e. write a method of `getindex` for `IsotopicByOutputs` and `IsotopicByFeatures`?

src/mokernels/moinput.jl
devmotion, line 9: Do we need a supertype at all? `ColVecs` and `RowVecs` also don't have a special supertype, they are just subtypes of `AbstractVector{S}`. So I think `MOInput` should not be needed but we want to make sure that `IsotopicByFeatures{S,T<:AbstractVector{S}} <: AbstractVector{Tuple{S,Int}}` and similar for `IsotopicByOutputs`.

Note that currently this is not the case since we just have `<: AbstractVector{Tuple{Any,Int}}`, so the information about `S` is lost. Since I don't think there's any benefit in keeping `MOInput` I would suggest dropping it.

willtebbutt (commented):
Left some feedback -- this is already looking good.

We'll need to add MOInput to the deprecation path to avoid making this a breaking change.

@devmotion do you have thoughts on the best way to achieve that in this situation? We've essentially just renamed `MOInput` to something else, so maybe an alias is fine?

Also, we'll need to update + extend the docs around multi-output kernels a bit, in particular here: https://juliagaussianprocesses.github.io/KernelFunctions.jl/dev/api/#Inputs-for-Multiple-Outputs

src/mokernels/moinput.jl
thomasgudjonwright: I had this discussion with @willtebbutt. The reasoning for keeping the super type was eliminating a lot of repeated code (rewriting each `Base` dispatch for each, when all but `getindex` are identical) or not dispatching on `Union{IsotopicByFeatures, IsotopicByOutputs}`.

That being said, we also agreed that the route we took is by no means objectively correct so I am happy switching to one of the two directions I just mentioned above!

src/mokernels/moinput.jl
devmotion, line 77: It is better to define `Base.getindex(inp::IsotopicByOutputs, ind::Integer)` and `Base.getindex(inp::IsotopicByFeatures, ind::Integer)` separately. Then you also don't need the `else` block with the error.

src/mokernels/moinput.jl
```diff
@@ -39,11 +74,18 @@ Base.firstindex(inp::MOInput) = 1
 
 function Base.getindex(inp::MOInput, ind::Integer)
     if ind > 0
```
devmotion, line 76: This check seems strange, instead we should check the bounds properly with `@boundscheck checkbounds` (such that it can be elided with `@inbounds`): https://docs.julialang.org/en/v1/devdocs/boundscheck/

src/mokernels/moinput.jl
```diff
@@ -39,11 +74,18 @@ Base.firstindex(inp::MOInput) = 1
...
-        out_dim = ind ÷ length(inp.x) + 1
-        ind = ind % length(inp.x)
-        if ind == 0
-            ind = length(inp.x)
-            out_dim -= 1
+        if inp isa IsotopicByOutputs
+            out_dim = ind ÷ length(inp.x) + 1
+            ind = ind % length(inp.x)
```
devmotion, line 79: This seems incorrect since `ind` can become `0`. Probably one has to use `fldmod1` here (this should also be used below since probably it is faster and hopefully more readable).

BTW generally one has to be careful when reassigning variables - type changes would mess up type stability of the function.

src/mokernels/moinput.jl
devmotion: Alternatively, we could just make it `::Int`? It seems currently we mostly dispatch on `Tuple{Any,Int}` anyway.

willtebbutt: Seems fine to me. Technically breaking, but really unlikely to break anything in practice.

src/mokernels/moinput.jl
devmotion: >  The reasoning for keeping the super type was eliminating a lot of repeated code (rewriting each Base dispatch for each, when all but getindex are identical) or not dispatching on Union{IsotopicByFeatures, IsotopicByOutputs}.

I see. If it is not expected that someone else ones to hook into these methods as well, one could just define an internal union with `const` and dispatch on it instead of writing `Union{Isotopic....}` explicitly in all these cases.

If one wants it to be extendable, then probably a supertype makes sense. However, in this case you want a different supertype that keeps the information about the element type. I.e. you would want
```julia
abstract type MOInput{T} <: AbstractVector{Tuple{T,Int}}

struct IsotopicByFeatures{S,T<:AbstractVector{S}} <: MOInput{S}
    x::T
    out_dim::Int
end
...
```

devmotion:
> We'll need to add MOInput to the deprecation path to avoid making this a breaking change. @devmotion do you have thoughts on the best way to achieve this in this situation? We've essentially just renamed MOInput to something else, so maybe an alias is fine?

It would be easier if the abstract super type would have a different name. Then we could just use
```julia
Base.@deprecate_binding MOInput IsotopicByFeatures
```
However, if `MOInput` becomes an abstract type I think deprecations won't help (or even be possible) and it is always a breaking change. Any dispatch on e.g. `::Vector{MOInput{Float64}}` will always be broken since it does not cover `::Vector{IsotopicByFeatures{Float64}}`. So IMO it would be better to use a different name for the supertype. Maybe something like `IsotopicMOInput`?

willtebbutt:
> So IMO it would be better to use a different name for the supertype. Maybe something like IsotopicMOInput?

This makes sense to me. Although we probably want to pluralise it to make it clear that it's a collection of inputs, not just a single input.

src/mokernels/moinput.jl
thomasgudjonwright: Done!

thomasgudjonwright: Done!

src/mokernels/moinput.jl
thomasgudjonwright: I think this should be addressed applying @devmotion  [suggestion](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/310#discussion_r658663432) above

src/mokernels/moinput.jl
thomasgudjonwright: This should be addressed by the [suggestion](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/310#discussion_r658663432) by @devmotion above

src/mokernels/moinput.jl
thomasgudjonwright: This should be addressed by the [suggestion](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/310#discussion_r658663432) by @devmotion above

src/mokernels/moinput.jl
```diff
@@ -1,15 +1,23 @@
...
+    IsotopicMOInputs
 
-A data type to accomodate modelling multi-dimensional output data.
+An abstract type to accomodate modelling multi-dimensional output data. There are two
+subtypes that each specify a unique ordering of the dimensions.
 
-`MOInput(x, out_dim)` has length `length(x) * out_dim`.
+See [Inputs for Multiple Outputs](@ref) in the docs for more info.
+"""
+abstract type IsotopicMOInputs <: AbstractVector{Tuple{T,Int}} end
```
devmotion, line 9:
```suggestion
abstract type IsotopicMOInputs{T} <: AbstractVector{Tuple{T,Int}} end
```

src/mokernels/moinput.jl
```diff
@@ -51,7 +96,7 @@ function Base.getindex(inp::MOInput, ind::Integer)
     end
 end
 
-Base.iterate(inp::MOInput) = (inp[1], 1)
-function Base.iterate(inp::MOInput, state)
+Base.iterate(inp::IsotopicMOInputs) = (inp[1], 1)
+function Base.iterate(inp::IsotopicMOInputs, state)
```
devmotion, line 100: Is this needed? Since `IsotopicMOInputs isa AbstractVector` most methods, including `iterate`, should be defined automatically.

src/mokernels/moinput.jl
```diff
@@ -18,26 +26,63 @@ julia> MOInput(x, 2)
...
 
-Base.size(inp::MOInput, d) = d::Integer == 1 ? inp.out_dim * size(inp.x, 1) : 1
-Base.size(inp::MOInput) = (inp.out_dim * size(inp.x, 1),)
+Base.size(inp::IsotopicMOInputs, d) = d::Integer == 1 ? inp.out_dim * size(inp.x, 1) : 1
+Base.size(inp::IsotopicMOInputs) = (inp.out_dim * size(inp.x, 1),)
 
-Base.lastindex(inp::MOInput) = length(inp)
-Base.firstindex(inp::MOInput) = 1
+Base.lastindex(inp::IsotopicMOInputs) = length(inp)
+Base.firstindex(inp::IsotopicMOInputs) = 1
```
devmotion, line 83: Usually it is not necessary to define `lastindex` and `firstindex` for `AbstractVector`s. At most one has to define the indexing style but IIRC it defaults to linear indexing anyway. So I assume this could be removed?

src/mokernels/moinput.jl
```diff
@@ -18,26 +26,63 @@ julia> MOInput(x, 2)
...
-    out_dim::Integer
+    out_dim::Int
 end
 
-Base.length(inp::MOInput) = inp.out_dim * length(inp.x)
+Base.length(inp::IsotopicMOInputs) = inp.out_dim * length(inp.x)
 
-Base.size(inp::MOInput, d) = d::Integer == 1 ? inp.out_dim * size(inp.x, 1) : 1
-Base.size(inp::MOInput) = (inp.out_dim * size(inp.x, 1),)
+Base.size(inp::IsotopicMOInputs, d) = d::Integer == 1 ? inp.out_dim * size(inp.x, 1) : 1
```
devmotion, line 79: I think this is not needed? It should be covered by https://github.com/JuliaLang/julia/blob/ed4c44fbafbbace523ebdc9440bf1f542157048a/base/abstractarray.jl#L42.

src/mokernels/moinput.jl
willtebbutt: I'm happy with IsotopicMOInputs., so let's go with that.

src/mokernels/moinput.jl
```diff
@@ -18,26 +26,63 @@ julia> MOInput(x, 2)
...
-struct MOInput{T<:AbstractVector} <: AbstractVector{Tuple{Any,Int}}
+
+struct IsotopicByOutputs{S,T<:AbstractVector{S}} <: IsotopicMOInputs{S}
     x::T
-    out_dim::Integer
+    out_dim::Int
 end
 
-Base.length(inp::MOInput) = inp.out_dim * length(inp.x)
+Base.length(inp::IsotopicMOInputs) = inp.out_dim * length(inp.x)
```
devmotion, line 77: This definition, and the definition of e.g. `size` and `kernelmatrix` below, all assume that `x.x` and `x.out_dim` exist and have the same meaning as the fields in `IsotopicByFeatures` and `IsotopicByOutputs`. It would be more correct to either define these methods only for `IsotopicByFeatures` and `IsotopicByOutputs` or to make it completely clear that `IsotopicMOInputs` have to have these fields and have to behave in this way. In both cases the question is how useful the supertype actually is - in the first case a union would be more helpful for reducing the amount of code and in the second case it is not clear if there are any other subtypes possible with this restriction.

willtebbutt: Yeah, I think you're right. I was arguing in favour of the supertype over the union type offline with Tom, but you make a good point. I agree that a union is more appropriate in this case, since we're really just doing this for the sake of reducing the amount of code needed.

thomasgudjonwright: Yeah this was a concern of mine. I am not sure how best to define the supertype and subtypes to leave no assumptions for the fields of `IsotopicMOInputs`. Its an argument for not going the supertype direction, unless there exists a way to subtype a `struct` that I don't know about or something like that

thomasgudjonwright: Should I add a `const` for the union or are we not bothered repeating `Union{IsotopicByFeatures, IsotopicByOutputs}` a few times?

src/mokernels/moinput.jl
thomasgudjonwright: So I think your first point relates to the `IsotopicByOutputs` which is the new version of the old `MOInput`. Maybe whoever wrote that has a better perspective than me for why its the way it is. Otherwise, I am happy to try and get the same functionality while improving the stability of the arithmetic.

Your second point relates to the use of `round` and the `Int` typecast? If so I am removing them, they aren't needed with `\div`, I had it when I accidentally was using `/`

src/mokernels/moinput.jl
thomasgudjonwright: Based off the tests that we currently run, it appears we need to define:
```
Base.size(inp::IsotopicMOInputs)
Base.iterate(inp::IsotopicMOInputs)
Base.iterate(inp::IsotopicMOInputs, state)
```
Given `const IsotopicMOInputs = Union{IsotopicByFeatures, IsotopicByOutputs}`

Updating accordingly

devmotion: I don't think we should define `iterate` but remove these tests. I don't see any benefit in defining this custom iteration procedure, the default for vectors should be fine.

src/mokernels/moinput.jl
```diff
@@ -51,7 +75,20 @@ function Base.getindex(inp::MOInput, ind::Integer)
...
-Base.iterate(inp::MOInput) = (inp[1], 1)
-function Base.iterate(inp::MOInput, state)
+function Base.getindex(inp::IsotopicByFeatures, ind::Integer)
+    if ind > 0
+        out_dim = (ind - 1) % inp.out_dim + 1
+        ind = (ind - 1) ÷ inp.out_dim + 1
+        return (inp.x[ind], out_dim::Int)
+    else
+        throw(BoundsError(string("Trying to access at ", ind)))
+    end
```
devmotion, line 85: As per https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/310/files#r658653385, I suggest
```suggestion
    @boundscheck checkbounds(inp, ind)
    feature_index, output_index = fldmod1(ind, inp.out_dim)
    feature = @inbounds inp.x[feature_index]
    return feature, output_index
```

The other `getindex` implementation should be updated in a similar way.

src/mokernels/moinput.jl
```diff
@@ -51,7 +75,20 @@ function Base.getindex(inp::MOInput, ind::Integer)
...
+end
+
+Base.size(inp::IsotopicMOInputs) = (inp.out_dim * size(inp.x, 1),)
```
devmotion, line 88: A bit simpler:
```suggestion
Base.size(inp::IsotopicMOInputs) = (inp.out_dim * length(inp.x),)
```

src/mokernels/moinput.jl
thomasgudjonwright: Given we are removing the supertype this has been adjusted to:

```
struct IsotopicByFeatures{S,T<:AbstractVector{S}} <: AbstractVector{Tuple{T,Int}}
    x::T
    out_dim::Int
end
```

thomasgudjonwright:
> > We'll need to add MOInput to the deprecation path to avoid making this a breaking change. @devmotion do you have thoughts on the best way to achieve this in this situation? We've essentially just renamed MOInput to something else, so maybe an alias is fine?
> 
> It would be easier if the abstract super type would have a different name. Then we could just use
> 
> ```julia
> Base.@deprecate_binding MOInput IsotopicByFeatures
> ```
> 
> However, if `MOInput` becomes an abstract type I think deprecations won't help (or even be possible) and it is always a breaking change. Any dispatch on e.g. `::Vector{MOInput{Float64}}` will always be broken since it does not cover `::Vector{IsotopicByFeatures{Float64}}`. So IMO it would be better to use a different name for the supertype. Maybe something like `IsotopicMOInput`?

If we are dropping the supertype does this change how `MOInput` gets deprecated? It is now a breaking change no matter what I assume.

src/mokernels/moinput.jl
thomasgudjonwright: Good find, thanks!

src/mokernels/moinput.jl
thomasgudjonwright: Happy to do so if you think thats best. I would be curious to hear the original reason for including `iterate` for the `MOInput` definition if it is never used in practice in case it was included for a specific reason before we remove it. Do you know who first defined the `MOInput`?

src/mokernels/moinput.jl
thomasgudjonwright: This should be fixed by removing the supertype, resolving

src/mokernels/moinput.jl
devmotion: IIRC it was implemented by @sharanry. I suggested before that it should be removed and there was an umerged PR (https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/169). I don't think there's anything to worry here, the AbstractArray interface is a bit confusing and so it just happened that too many methods were defined initially.

src/mokernels/moinput.jl
thomasgudjonwright: Done!

src/mokernels/moinput.jl
thomasgudjonwright: Done :)

src/mokernels/moinput.jl
thomasgudjonwright: Thank you for this! It has been implemented in both dispatches :)

docs/src/api.md
```diff
@@ -58,13 +58,16 @@ For an explanation of this design choice, see [the design notes on multi-output
...
+KernelFunctions.jl provides the following types for situations in which all outputs are observed all of the time:
 ```@docs
-MOInput
+IsotopicByFeatures
+IsotopicByOutputs
 ```
-As with [`ColVecs`](@ref) and [`RowVecs`](@ref) for vector-valued input spaces, this
-type enables specialised implementations of e.g. [`kernelmatrix`](@ref) for
-[`MOInput`](@ref)s in some situations.
+For[`IsotopicByFeatures`](@ref) i
```
willtebbutt, line 66: Is this a leftover line?

test/mokernels/moinput.jl
```diff
@@ -1,19 +1,40 @@
...
-    @test mgpi[5] == (x[1], 2)
-    @test mgpi[7] == (x[3], 2)
-    @test all([(x_, i) for i in 1:3 for x_ in x] .== mgpi)
+        @test mgpi_bo[2] == (x[2], 1)
+        @test mgpi_bo[5] == (x[1], 2)
+        @test mgpi_bo[7] == (x[3], 2)
+        @test all([(x_, i) for i in 1:3 for x_ in x] .== mgpi_bo)
+    end
+
+    mgpi_bf = IsotopicByFeatures(x, 3)
```
willtebbutt, line 22: Can this be moved inside the testset below?

test/mokernels/moinput.jl
```diff
@@ -1,19 +1,40 @@
 @testset "moinput" begin
     x = [rand(5) for _ in 1:4]
-    mgpi = MOInput(x, 3)
+    mgpi_bo = IsotopicByOutputs(x, 3)
```
willtebbutt, line 3: Can this be moved inside the testset below?

willtebbutt (commented):
Done a quick pass. A couple of really minor things.

willtebbutt:
> If we are dropping the supertype does this change how MOInput gets deprecated? It is now a breaking change no matter what I assume.

I would imagine we can just alias `MOInput` to `IsotopicByOutputs` and be done with it -- I wouldn't have thought this would be breaking then? Or am I missing something?

src/mokernels/moinput.jl
```diff
@@ -18,40 +16,63 @@ julia> MOInput(x, 2)
...
+
+As shown above, an `IsotopicByFeatures` represents a vector of tuples.
 The first `length(x)` elements represent the inputs for the first output, the second
 `length(x)` elements represent the inputs for the second output, etc.
 
 See [Inputs for Multiple Outputs](@ref) in the docs for more info.
 """
-struct MOInput{T<:AbstractVector} <: AbstractVector{Tuple{Any,Int}}
+
+struct IsotopicByFeatures{S,T<:AbstractVector{S}} <: AbstractVector{Tuple{T,Int}}
```
devmotion, line 29: This should be
```suggestion
struct IsotopicByFeatures{S,T<:AbstractVector{S}} <: AbstractVector{Tuple{S,Int}}
```
shouldn't it?

src/mokernels/moinput.jl
```diff
@@ -18,40 +16,63 @@ julia> MOInput(x, 2)
...
+ (3, 1)
+ (3, 2)
+```
+
+As shown above, an `IsotopicByOutputs` represents a vector of tuples.
+The first `out_dim` elements represent all outputs for the first input, the second
+`out_dim` elements represent the outputs for the second input, etc.
+"""
+
+struct IsotopicByOutputs{S,T<:AbstractVector{S}} <: AbstractVector{Tuple{T,Int}}
```
devmotion, line 57: Same here:
```suggestion
struct IsotopicByOutputs{S,T<:AbstractVector{S}} <: AbstractVector{Tuple{S,Int}}
```

devmotion: Maybe we should add a test for it?

st--:
Is it just me who finds the current convention rather unintuitive? "By features" makes me think "grouped by features" (same for the initial "block by featuies"), so I would expect all outputs for the first feature, then all outputs for the second feature, etc. - which is instead what `IsotopicByOutputs` constructs.

Also, I'm wondering if it'd be useful to include an `MO` in the names to clarify what these objects are about (there are a lot of items in the namespace already)...

thomasgudjonwright:
> Is it just me who finds the current convention rather unintuitive? "By features" makes me think "grouped by features" (same for the initial "block by featuies"), so I would expect all outputs for the first feature, then all outputs for the second feature, etc. - which is instead what `IsotopicByOutputs` constructs.
> 
> Also, I'm wondering if it'd be useful to include an `MO` in the names to clarify what these objects are about (there are a lot of items in the namespace already)...

I think it currently works in the way you claim is more intuitive (I agree, the opposite would be quite confusing!) 

```
julia> o = IsotopicByOutputs(x,2)
6-element IsotopicByOutputs{SubArray{Float64, 1, Matrix{Float64}, Tuple{Int64, Base.Slice{Base.OneTo{Int64}}}, true}, RowVecs{Float64, Matrix{Float64}, SubArray{Float64, 1, Matrix{Float64}, Tuple{Int64, Base.Slice{Base.OneTo{Int64}}}, true}}}:
 ([0.48177990377024393, 0.8080869947045819, 0.01357827855442939], 1)
 ([0.46941035398309006, 0.5079452944317515, 0.3991885361162597], 1)
 ([0.9541462929301829, 0.35490339688118633, 0.3247450403266279], 1)
 ([0.48177990377024393, 0.8080869947045819, 0.01357827855442939], 2)
 ([0.46941035398309006, 0.5079452944317515, 0.3991885361162597], 2)
 ([0.9541462929301829, 0.35490339688118633, 0.3247450403266279], 2)

julia> f = IsotopicByFeatures(x,2)
6-element IsotopicByFeatures{SubArray{Float64, 1, Matrix{Float64}, Tuple{Int64, Base.Slice{Base.OneTo{Int64}}}, true}, RowVecs{Float64, Matrix{Float64}, SubArray{Float64, 1, Matrix{Float64}, Tuple{Int64, Base.Slice{Base.OneTo{Int64}}}, true}}}:
 ([0.48177990377024393, 0.8080869947045819, 0.01357827855442939], 1)
 ([0.48177990377024393, 0.8080869947045819, 0.01357827855442939], 2)
 ([0.46941035398309006, 0.5079452944317515, 0.3991885361162597], 1)
 ([0.46941035398309006, 0.5079452944317515, 0.3991885361162597], 2)
 ([0.9541462929301829, 0.35490339688118633, 0.3247450403266279], 1)
 ([0.9541462929301829, 0.35490339688118633, 0.3247450403266279], 2)
```

src/mokernels/moinput.jl
thomasgudjonwright: resolving

src/mokernels/moinput.jl
thomasgudjonwright: resolving

src/mokernels/moinput.jl
thomasgudjonwright: this has been addressed, resolving

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "KernelFunctions"
 uuid = "ec8451be-7e33-11e9-00cf-bbf324bd1392"
-version = "0.10.5"
+version = "0.11.5"
```
nickrobinson251, line 3:
```suggestion
version = "0.10.6"
```

This is none breaking if we've just renamed `MOInput` -> `IsotopicByFeatures`

We just need to add `Base.@deprecate_binding MOInput IsotopicByFeatures`

devmotion: Actually, this is breaking since we changed the type parameters (and even the type of the fields). Any dispatches of the form `::MOInput{T}` and `::Vector{MOInput{T}}` won't work anymore.

devmotion:
```suggestion
version = "0.11.0"
```

src/mokernels/moinput.jl
```diff
@@ -1,15 +1,13 @@
...
-`MOInput(x, out_dim)` has length `length(x) * out_dim`.
+`IsotopicByFeatures(x, out_dim)` has length `length(x) * out_dim`.
 
 ```jldoctest
 julia> x = [1, 2, 3];
 
-julia> MOInput(x, 2)
-6-element MOInput{Vector{Int64}}:
+julia> IsotopicByFeatures(x, 2)
+6-element IsotopicByFeatures{Vector{Int64}}:
```
st--, line 10:
```suggestion
julia> IsotopicByOutputs(x, 2)
6-element IsotopicByOutputs{Vector{Int64}}:
```

st--:
@thomasgudjonwright 
> I think it currently works in the way you claim is more intuitive (I agree, the opposite would be quite confusing!)

that doesn't seem to be consistent with the docstring examples though?

Project.toml
thomasgudjonwright: Oops aha, lazy type 😆  thanks!

src/mokernels/moinput.jl
thomasgudjonwright: Ah now I get your source of confusion for the two, sorry about that! Good find :)

docs/src/api.md
thomasgudjonwright: Whoops, indeed it is!

src/mokernels/moinput.jl
thomasgudjonwright: I think this should be fixed now!

src/mokernels/moinput.jl
thomasgudjonwright: I think you're right. `S` is the type of a single observation and `T` is the vector of those. Re your related [comment](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/310#discussion_r659760809), I will add a test!

src/mokernels/moinput.jl
thomasgudjonwright: Wasn't exactly sure what a test for this should look like, let me know if you had something different in mind and I will fix it! :)

src/mokernels/moinput.jl
st--: Ah but why did this not break the doctests ? :thinking: :confused:

src/mokernels/moinput.jl
```diff
@@ -1,15 +1,45 @@
...
+
+Accomodates modelling multi-dimensional output data.
+
+As shown above, an `IsotopicByFeatures` represents a vector of tuples.
+The first `out_dim` elements represent all outputs for the first input, the second
+`out_dim` elements represent the outputs for the second input, etc.
+
+See [Inputs for Multiple Outputs](@ref) in the docs for more info.
+"""
+
```
willtebbutt, line 27: Should there be this line break here? Will this docstring actually be associated with `IsotropicByFeatures`?

src/mokernels/moinput.jl
```diff
@@ -18,40 +48,32 @@ julia> MOInput(x, 2)
...
+Accomodates modelling multi-dimensional output data.
+
+As shown above, an `IsotopicByOutputs` represents a vector of tuples.
 The first `length(x)` elements represent the inputs for the first output, the second
 `length(x)` elements represent the inputs for the second output, etc.
-
-See [Inputs for Multiple Outputs](@ref) in the docs for more info.
 """
-struct MOInput{T<:AbstractVector} <: AbstractVector{Tuple{Any,Int}}
+
```
willtebbutt, line 57: Same here -- does the line break matter?

src/mokernels/moinput.jl
```diff
@@ -18,40 +48,32 @@ julia> MOInput(x, 2)
  (3, 2)
 ```
 
-As shown above, an `MOInput` represents a vector of tuples.
+Accomodates modelling multi-dimensional output data.
```
willtebbutt, line 51:
```suggestion
Accomodates modelling multi-dimensional output data where all outputs are always observed.
```

willtebbutt (commented):
Just a couple of comments.

src/mokernels/moinput.jl
thomasgudjonwright: Resolving

src/mokernels/moinput.jl
thomasgudjonwright: They weren't being included because of the line break noticed by @willtebbutt [here](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/310#pullrequestreview-696036211). Being fixed now :)

src/mokernels/moinput.jl
thomasgudjonwright: Nice find! Thanks!!

src/mokernels/moinput.jl
thomasgudjonwright: This is the case for types right? I will also add it to the `IsotopicByFeatures` docstring

st--:
```diff
-Accomodates
+Accommodates
```

src/KernelFunctions.jl
```diff
@@ -35,7 +35,7 @@ export spectral_mixture_kernel, spectral_mixture_product_kernel
 
 export ColVecs, RowVecs
 
-export MOInput
+export IsotopicByFeatures, IsotopicByOutputs
```
st--, line 38:
```suggestion
export MOIsotopicByFeatures, MOIsotopicByOutputs
```
?

src/KernelFunctions.jl
```diff
@@ -35,7 +35,7 @@ export spectral_mixture_kernel, spectral_mixture_product_kernel
...
 export IndependentMOKernel,
     LatentFactorMOKernel, IntrinsicCoregionMOKernel, LinearMixingModelKernel
```
st--, line 40:
```suggestion
    LatentFactorMOKernel, IntrinsicCoregionMOKernel, LinearMixingMOKernel
```
(can be a separate PR)

src/KernelFunctions.jl
thomasgudjonwright: Are you suggesting we change the names of these (this would of course require changes in more locations)?

src/KernelFunctions.jl
thomasgudjonwright: Same as above question regarding changing the name. There was a [thread](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/304#discussion_r655369071) about the naming of this `MOKernel` in the PR that added it.

It is called the Linear Mixing Model or LMM so I think either one of those needs to be in the name. Either way we can maybe discuss that in a separate issue :)

src/KernelFunctions.jl
devmotion: Personally I am not a fan of the MO acronym since in my opinion it is not a common abbreviation and hence does not explain clearly the purpose or meaning of the kernels/inputs/etc.

src/KernelFunctions.jl
st--: It's also called Intrinsic Coregionalization Model and [Semiparametric] Latent Factor Model - the "Model" doesn't add much, but distinguishing between multi-output kernels and single-output kernels seems useful to me:)

st--: fwiw I'd be against the abbreviations...

thomasgudjonwright: I agree adding the `MO` could be beneficial. However I believe it is important to maintain `Linear Mixing Model` in its entirety as this is the name that it was assigned by those who developed it and will be using it (of course not using it exclusively). I could be wrong though, @willtebbutt @nickrobinson251 may also have thoughts

thomasgudjonwright: Re abbreviations I think that would just potentially be an alias added (discussed in the last PR, dropped for now)

src/KernelFunctions.jl
st--: > Are you suggesting we change the names of these (this would of course require changes in more locations)?

yes that was the idea, will make that more explicit next time:)

"MO" is reasonably common in ML modelling literature.

Do you have alternative suggestions for how to clarify that this is multi-output specific ?

src/KernelFunctions.jl
st--: It's only a *model* once you do something with the kernel, so I don't think it makes sense to include "Model" as part of the name. It *is* a multi-output kernel that is created by linearly mixing of kernel matrices.

willtebbutt: Agreed -- for example, I would be completely fine with `LinearMixingKernel` or something like that.

src/KernelFunctions.jl
willtebbutt: Does the use of the word `Isotopic` not make it clear that it's got something to do with multiple outputs, or at least sounds sufficiently specific to motivate a reader to look at the docstring?

src/KernelFunctions.jl
st--: Well, it comes down to whether you think saving the 10% extra characters from "LinearMixingMOKernel" vs "LinearMixingKernel" offsets the corresponding loss in explicitness. Users of the package can always use their own aliases to make their own code shorter. But KernelFunctions only exposes a single name space that has a lot of entities already, so making it easier for people to distinguish between single-output and multi-output kernels is valuable. (In my experience, in practice you often want to do quite different things for multi-output *models* so easily knowing whether a kernel is multi-output or not is useful.)

devmotion: I would argue that you can already distinguish them since (currently) all multi-output kernels are a subtype of `MOKernel` (which maybe should rather be called `MultiOutputKernel`).

src/KernelFunctions.jl
devmotion: I think maybe it would be more important to clearly point out that these are specific types of inputs, something like `IsotopicInputsByFeatures`, than mentioning "MO"/multi-output.

src/KernelFunctions.jl
st--: I'd be happier with that if there was a SingleOutputKernel as well ... I'm not sure if MOKernel being a subtype of Kernel is really a good idea...

willtebbutt: Is it not okay to view a multi-output kernel as a special case of a regular kernel?

src/KernelFunctions.jl
st--: Yeah including "Inputs" in the name might be useful too

src/mokernels/moinput.jl
```diff
@@ -1,15 +1,44 @@
...
+Accommodates modelling multi-dimensional output data where all outputs are always observed.
+
+As shown above, an `IsotopicByFeatures` represents a vector of tuples.
+The first `out_dim` elements represent all outputs for the first input, the second
+`out_dim` elements represent the outputs for the second input, etc.
+
+See [Inputs for Multiple Outputs](@ref) in the docs for more info.
+"""
+struct IsotopicByFeatures{S,T<:AbstractVector{S}} <: AbstractVector{Tuple{S,Int}}
+    x::T
```
st--, line 28: How about
```suggestion
    xs::T
```
or even better
```suggestion
    features::T
```
to clarify what this is?

src/mokernels/moinput.jl
```diff
@@ -1,15 +1,44 @@
...
+    out_dim::Int
```
st--, line 29:
```suggestion
    output_dim::Int
```
?
(both for the other struct as well of course)

test/mokernels/moinput.jl
```diff
@@ -1,19 +1,45 @@
...
-
-    @test mgpi[2] == (x[2], 1)
-    @test mgpi[5] == (x[1], 2)
-    @test mgpi[7] == (x[3], 2)
-    @test all([(x_, i) for i in 1:3 for x_ in x] .== mgpi)
+    type_1 = AbstractVector{Tuple{Vector{Float64},Int}}
+    type_2 = AbstractVector{Tuple{AbstractVector{Vector{Float64}},Int}}
+
+    @testset "isotopicbyoutputs" begin
+        mgpi_bo = IsotopicByOutputs(x, 3)
```
st--, line 7: I can guess that `bo` means "by output", but what does the rest of it `mgpi` mean?
How about just calling it `ibo` (and `ibf` below), if you want something short that is distinguishable between the two testset blocks ?

src/mokernels/moinput.jl
thomasgudjonwright: If we make this change (which I am happy to do), I think this becomes more breaking, since I think the plan was to add an alias for `IsotopicByOutput` to `MOInput` (the old name) which uses `x` and `out_dim`.

test/mokernels/moinput.jl
thomasgudjonwright: I have no idea what `mgpi` stands for, it was what the previous test called it but I agree, will change it :)

src/mokernels/moinput.jl
st--: Ah, I thought it was going to be a breaking change anyways. In that case we can add it to the list for cleaning up struct field names more generally, in a separate PR

willtebbutt:
What is people's opinion on adding helper functions to construct these, and not exporting the types directly? e.g. `isotopic_by_output` and `isotopic_by_feature`. At the minute, the implementations would just be
```julia
isotopic_by_output = IsotopicByOutput
isotopic_by_feature = IsotopicByFeature
```
(possibly a `const` is needed).

The rationale is that at some point we _might_ want to change the implementation of these types. I'm specifically imagining that we might well introduce some kind of generic cartesian product input type, of which these would be special cases. In that situation, we could keep these functions the same (since they're reasonably interpretable) and just change the type that implements them behind the scenes. In practice, I would expect this would make these changes less breaking because user-facing code wouldn't change.

thomasgudjonwright:
> What is people's opinion on adding helper functions to construct these, and not exporting the types directly? e.g. `isotopic_by_output` and `isotopic_by_feature`. At the minute, the implementations would just be
> 
> ```julia
> isotopic_by_output = IsotopicByOutput
> isotopic_by_feature = IsotopicByFeature
> ```
> 
> (possibly a `const` is needed).
> 
> The rationale is that at some point we _might_ want to change the implementation of these types. I'm specifically imagining that we might well introduce some kind of generic cartesian product input type, of which these would be special cases. In that situation, we could keep these functions the same (since they're reasonably interpretable) and just change the type that implements them behind the scenes. In practice, I would expect this would make these changes less breaking because user-facing code wouldn't change.

If the types get redefined is it not likely that these functions would also require editing that would be user-facing? Such as adjusting arguments

willtebbutt:
>If the types get redefined is it not likely that these functions would also require editing that would be user-facing? Such as adjusting arguments

This is my point -- I don't think they would. Ultimately you'd still just need to specify the features and the number of outputs, so you'd just have to change the implementation of the function.

thomasgudjonwright:
> > If the types get redefined is it not likely that these functions would also require editing that would be user-facing? Such as adjusting arguments
> 
> This is my point -- I don't think they would. Ultimately you'd still just need to specify the features and the number of outputs, so you'd just have to change the implementation of the function.

Does this make the `MOInput` alias for `IsotopicByOutputs` redundant? If we are choosing to use functions to build the objects it doesn't really make sense to keep `MOInput`. This would make this breaking again though

willtebbutt:
Hmmm I've actually completely lost track of what aspect of this PR is breaking. Could you remind me? If we need a minor version bump, then we might as well drop the `MOInputs` type now (unless anyone objects)

thomasgudjonwright:
> Hmmm I've actually completely lost track of what aspect of this PR is breaking. Could you remind me? If we need a minor version bump, then we might as well drop the `MOInputs` type now (unless anyone objects)

We are dropping the `MOInput`. We were trying to avoid this being breaking before by creating an alias `const MOInput = IsotopicByOutputs` but given we are not exporting those objects and using functions (like `isotopic_by_outputs(...)`), it seems strange to me to allow for people to create the objects via `MOInput(...)`

docs/src/api.md
```diff
@@ -58,13 +58,21 @@ For an explanation of this design choice, see [the design notes on multi-output
 An input to a multi-output `Kernel` should be a `Tuple{T, Int}`, whose first element specifies a location in the domain of the multi-output GP, and whose second element specifies which output the inputs corresponds to.
 The type of collections of inputs for multi-output GPs is therefore `AbstractVector{<:Tuple{T, Int}}`.
 
-KernelFunctions.jl provides the following type or situations in which all outputs are observed all of the time:
+KernelFunctions.jl provides the following types for situations in which all outputs are observed all of the time:
 ```@docs
-MOInput
+IsotopicByFeatures
```
thomasgudjonwright, line 63: Is it okay to still include these docs in this way if they aren't exported (ie a part of the API) if I clarify as I have below about how to implement them via their helper methods?

thomasgudjonwright:
Also the failing required test : `CI / Julia 1.3 - Others (pull_request)` has me confused. In Julia 1.3 would `x = [1, 2, 3]` return an `Array`? I don't quite know how to go about making it pass

docs/src/api.md
st--: If package users weren't going to see it it wouldn't make any sense to document it to expose it in the documentation either, then.

But actually this made me question @willtebbutt's suggestion of not exposing them: how would a downstream user then be able to implement their own multi-output kernels, if they can't dispatch on the input types ?

Maybe we should just mark them as "experimental, liable to change" and not worry too much about breaking things a bit later (it's clearly not 1.0 yet)?

thomasgudjonwright: Can they not still use it via `KernelFunctions.IsotopicByOutputs` for example?

willtebbutt:
> We are dropping the MOInput. We were trying to avoid this being breaking before by creating an alias const MOInput = IsotopicByOutputs but given we are not exporting those objects and using functions (like isotopic_by_outputs(...)), it seems strange to me to allow for people to create the objects via MOInput(...)

I agree that this will be slightly strange, but it's a temporary thing since we're deprecating `MOInput`, so I don't think it matters.

> Also the failing required test : CI / Julia 1.3 - Others (pull_request) has me confused. In Julia 1.3 would x = [1, 2, 3] return an Array? I don't quite know how to go about making it pass

This looks to me like Julia changed the way it printed stuff for 1.6 or something (since `Vector` and `Array{T, 1}` are the same type). @devmotion any idea how we can avoid this causing problems?

docs/src/api.md
willtebbutt: Hmm yes, good point @st--. I _think_ it's fine though, for the reasons that @thomasgudjonwright suggests.

devmotion:
> This looks to me like Julia changed the way it printed stuff for 1.6 or something (since `Vector` and `Array{T, 1}` are the same type). @devmotion any idea how we can avoid this causing problems?

Yes, it changed but https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/10ad9ee17c0e8d78df746672a0071f99c4ed63e4/test/runtests.jl#L180-L187 was added to fix these issues. Is there maybe another problem with the doctest? Or do we have to add another regex?

thomasgudjonwright:
> > This looks to me like Julia changed the way it printed stuff for 1.6 or something (since `Vector` and `Array{T, 1}` are the same type). @devmotion any idea how we can avoid this causing problems?
> 
> Yes, it changed but
> 
> https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/10ad9ee17c0e8d78df746672a0071f99c4ed63e4/test/runtests.jl#L180-L187
> 
> was added to fix these issues. Is there maybe another problem with the doctest? Or do we have to add another regex?

It is evaluating `6-element KernelFunctions.IsotopicByOutputs{Int64,Array{Int64,1}}:` instead of
`6-element KernelFunctions.IsotopicByOutputs{Int64, Vector{Int64}}:`.

With those regexs is it just the space that is the issue maybe?

src/KernelFunctions.jl
thomasgudjonwright: Discussion bumped to an [issue](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/332)
:)

src/mokernels/moinput.jl
willtebbutt: @thomasgudjonwright I would be happy for this change to be made in this PR, since this is a new type. Are you happy to make it?

willtebbutt:
@thomasgudjonwright looks like this PR just need to have the version number increment turned into a patch bump rather than a minor version bump, address the remaining comment from @st-- , and sort out the CI. Am I missing anything?

src/mokernels/moinput.jl
thomasgudjonwright: Resolving along with thread above for reason discussed in above thread. Discussion noted in a new [issue](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/332)

src/mokernels/moinput.jl
thomasgudjonwright: Discussion moved to a new [issue](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/332)

src/KernelFunctions.jl
thomasgudjonwright: Discussion moved to a new [issue](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/332)

src/KernelFunctions.jl
thomasgudjonwright: Discussion moved to a new [issue](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/332)

thomasgudjonwright:
> @thomasgudjonwright looks like this PR just need to have the version number increment turned into a patch bump rather than a minor version bump, address the remaining comment from @st-- , and sort out the CI. Am I missing anything?

Ahh good catch, just edited the version :) Which comment are you talking about? I just resolved all that related to naming linking the new [issue](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/332) to continue that discussion.

For the CI issue I will play with the regex and see if I can solve it before @devmotion is back

willtebbutt:
> Which comment are you talking about? I just resolved all that related to naming linking the new issue to continue that discussion.

Fair enough -- as long as we don't drop it I don't mind.

theogf:
Ok my Regex hunch tells me you should replace 
```
         r"(Array{[a-zA-Z0-9]+,\s?1}|Vector{[a-zA-Z0-9]+})", 
         r"(Array{[a-zA-Z0-9]+,\s?2}|Matrix{[a-zA-Z0-9]+})",
```
by 
```
         r"(Array{[a-zA-Z0-9]+,\s?1}|\s?Vector{[a-zA-Z0-9]+})", 
         r"(Array{[a-zA-Z0-9]+,\s?2}|\s?Matrix{[a-zA-Z0-9]+})",
```

src/mokernels/moinput.jl
thomasgudjonwright: @willtebbutt do we want any disparity between the two types though? Since we couldn't do the same for `IsotopicByOutputs` without breaking any `MOInputs` currently in the wild, right?

docs/Manifest.toml
```diff
@@ -119,7 +119,7 @@ version = "0.21.1"
 deps = ["ChainRulesCore", "Compat", "CompositionsBase", "Distances", "FillArrays", "Functors", "LinearAlgebra", "Random", "Requires", "SpecialFunctions", "StatsBase", "StatsFuns", "TensorCore", "Test", "ZygoteRules"]
 path = ".."
 uuid = "ec8451be-7e33-11e9-00cf-bbf324bd1392"
-version = "0.10.6"
+version = "0.11.0"
```
st--, line 122: should be consistent with main Project.toml ...
```suggestion
version = "0.10.7"
```

src/mokernels/moinput.jl
```diff
@@ -1,15 +1,44 @@
...
+ (1, 2)
+ (2, 1)
+ (2, 2)
+ (3, 1)
+ (3, 2)
+```
+
+Accommodates modelling multi-dimensional output data where all outputs are always observed.
+
+As shown above, an `IsotopicByFeatures` represents a vector of tuples.
```
st--, line 21:
```suggestion
As shown above, a `MOInputIsotopicByFeatures` represents a vector of tuples.
```

src/mokernels/moinput.jl
```diff
@@ -18,40 +47,83 @@ julia> MOInput(x, 2)
...
 The first `length(x)` elements represent the inputs for the first output, the second
 `length(x)` elements represent the inputs for the second output, etc.
-
-See [Inputs for Multiple Outputs](@ref) in the docs for more info.
 """
-struct MOInput{T<:AbstractVector} <: AbstractVector{Tuple{Any,Int}}
+struct MOInputIsotopicByOutputs{S,T<:AbstractVector{S}} <: AbstractVector{Tuple{S,Int}}
     x::T
-    out_dim::Integer
+    out_dim::Int
```
st--, line 58: :+1:

src/mokernels/moinput.jl
```diff
@@ -18,40 +47,83 @@ julia> MOInput(x, 2)
...
+    @boundscheck checkbounds(inp, ind)
+    feature_index, output_index = fldmod1(ind, inp.out_dim)
+    feature = @inbounds inp.x[feature_index]
+    return feature, output_index
 end
+
+Base.size(inp::IsotopicMOInputs) = (inp.out_dim * length(inp.x),)
+
+"""
+    MOInput(x::AbstractVector, out_dim::Integer)
```
st--, line 80:
```suggestion
    MOInput(x::AbstractVector, out_dim::Integer)

```

Also, would it be good to mention here that this is just for backwards compatibility? (Or otherwise clarify relation to MOInputIsotopicByOutputs)

src/mokernels/moinput.jl
```diff
@@ -18,40 +47,83 @@ julia> MOInput(x, 2)
...
+A data type to accomodate modelling multi-dimensional output data.
```
st--, line 81:
```suggestion
A data type to accommodate modelling multi-dimensional output data.
```

src/mokernels/moinput.jl
```diff
@@ -18,40 +47,83 @@ julia> MOInput(x, 2)
...
+The first `length(x)` elements represent the inputs for the first output, the second
+`length(x)` elements represent the inputs for the second output, etc.
+See [Inputs for Multiple Outputs](@ref) in the docs for more info.
+"""
+const MOInput = MOInputIsotopicByOutputs
+
+"""
+    isotopic_by_outputs(x::AbstractVector, out_dim::Integer)
+
+Helper function to construct [`IsotopicByOutputs`](@ref).
```
st--, line 106:
```suggestion
Helper function to construct [`MOInputIsotopicByOutputs`](@ref).
```

src/mokernels/moinput.jl
```diff
@@ -18,40 +47,83 @@ julia> MOInput(x, 2)
...
+```
+As shown above, an `MOInput` represents a vector of tuples.
+The first `length(x)` elements represent the inputs for the first output, the second
+`length(x)` elements represent the inputs for the second output, etc.
+See [Inputs for Multiple Outputs](@ref) in the docs for more info.
+"""
+const MOInput = MOInputIsotopicByOutputs
+
+"""
+    isotopic_by_outputs(x::AbstractVector, out_dim::Integer)
```
st--, line 104:
```suggestion
    isotopic_by_outputs(features::AbstractVector, output_dim::Integer)
```
? if we want to change the struct field names later anyways to be more explicit...

st--:
I don't understand why coverage would decrease (-0.08% overall), if the only changed source file has 100% coverage now... :monocle_face:

src/mokernels/moinput.jl
```diff
@@ -18,40 +47,83 @@ julia> MOInput(x, 2)
...
+Helper function to construct [`IsotopicByOutputs`](@ref).
+
+```jldoctest isotopic_by_outputs
+julia> x = [1, 2, 3];
+
+julia> KernelFunctions.isotopic_by_outputs(x, 2) == KernelFunctions.MOInputIsotopicByOutputs(x, 2)
+true
+```
+"""
+const isotopic_by_outputs = MOInputIsotopicByOutputs
```
st--, line 115: it's too early, I belatedly realized that this might also require a thin wrapper function instead of the const alias:
```suggestion
isotopic_by_outputs(features, output_dim) = MOInputIsotopicByOutputs(features, output_dim)
```

src/mokernels/moinput.jl
```diff
@@ -18,40 +47,83 @@ julia> MOInput(x, 2)
...
+Helper function to construct [`IsotopicByFeatures`](@ref).
+
+```jldoctest isotopic_by_features
+julia> x = [1, 2, 3];
+
+julia> KernelFunctions.isotopic_by_features(x, 2) == KernelFunctions.MOInputIsotopicByFeatures(x, 2)
+true
+```
+"""
+const isotopic_by_features = MOInputIsotopicByFeatures
```
st--, line 129:
```suggestion
isotopic_by_features(features, output_dim) = MOInputIsotopicByFeatures(features, output_dim)
```

st-- (commented):
Looks good, just left a couple small comments!

docs/Manifest.toml
thomasgudjonwright: Good catch!

src/mokernels/moinput.jl
thomasgudjonwright: 👍🏻

src/mokernels/moinput.jl
thomasgudjonwright: Doesn't using those argument names make this breaking?

src/mokernels/moinput.jl
thomasgudjonwright: Should these helpers just be removed until we come up with how we want to export the types following the plan discussed in the meeting?

src/mokernels/moinput.jl
thomasgudjonwright: Does this not make it breaking?

src/mokernels/moinput.jl
thomasgudjonwright: Definitely can add that. Just since it isn't exported it won't be in the docs so they will have to look at the source code docstring to see what the `MOInputIsotopicByOutputs` is if its mentioned

src/mokernels/moinput.jl
thomasgudjonwright: 👍🏻

src/mokernels/moinput.jl
thomasgudjonwright: Thanks :)

src/mokernels/moinput.jl
thomasgudjonwright: Should these helpers just be removed for now until we know how exactly we want to export these types as discussed in the meeting?

src/mokernels/moinput.jl
```diff
@@ -18,40 +47,83 @@ julia> MOInput(x, 2)
...
+julia> x = [1, 2, 3];
+
+julia> KernelFunctions.isotopic_by_outputs(x, 2) == KernelFunctions.MOInputIsotopicByOutputs(x, 2)
+true
+```
+"""
+const isotopic_by_outputs = MOInputIsotopicByOutputs
+
+"""
+    isotopic_by_features(x::AbstractVector, out_dim::Integer)
```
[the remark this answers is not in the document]
thomasgudjonwright: This discussion is moved [here](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/332)

src/mokernels/moinput.jl
thomasgudjonwright: Should we hold off on mentioning the backwards compatibility until we explicitly make the other types public by exporting them (or helper functions that build them)?

willtebbutt: Probably not a bad idea to add a note that says "as of version ... this feature is deprecated, and will be removed in the next breaking release"

thomasgudjonwright: But isn't it not deprecated at the moment? If we deprecate it now, we have no public multi output input types. 

Also is the plan to have it completely removed? I thought the idea was to leave it as an alias to minimize breaking changes but I may have misunderstood that it was just a short term plan

thomasgudjonwright: Note added, resolving

docs/Manifest.toml
```diff
@@ -119,7 +119,7 @@ version = "0.21.1"
 deps = ["ChainRulesCore", "Compat", "CompositionsBase", "Distances", "FillArrays", "Functors", "LinearAlgebra", "Random", "Requires", "SpecialFunctions", "StatsBase", "StatsFuns", "TensorCore", "Test", "ZygoteRules"]
 path = ".."
 uuid = "ec8451be-7e33-11e9-00cf-bbf324bd1392"
-version = "0.10.6"
+version = "0.10.7"
```
willtebbutt, line 122:
```suggestion
version = "0.10.8"
```
?

willtebbutt (approved):
Other than the patch versions appearing to not be aligned, LGTM!

docs/Manifest.toml
thomasgudjonwright: Good catch, thanks!

willtebbutt (approved):

PR #314, opened 2021-06-28
Run examples with separate processes

devmotion:
This is an (arguably) simpler alternative to https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/309 and (partially) https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/303.

Examples are run asynchronously with different processes and no load path trickery is needed. CompatHelper is updated such that example environment are updated as well. I also updated and fixed the SVM example to demonstrate that the pipeline works as expected (the other examples are not blacklisted but since they are not moved to a separate folder they are not run currently).

*Edit:* I did not base this PR on #303 intentionally since in this way there are no changes hidden in the other PR and no major unrelated rewriting is needed. The PR is self-contained and as minimal as possible. I assumed that the other examples would be fixed in separate PRs, e.g., by rebasing #303.


devmotion:
This should be good to go.

Documenter: https://juliagaussianprocesses.github.io/KernelFunctions.jl/previews/PR314/examples/svm/
Notebook: https://nbviewer.jupyter.org/github/JuliaGaussianProcesses/KernelFunctions.jl/blob/gh-pages/previews/PR314/examples/svm.ipynb

docs/make.jl
```diff
@@ -1,12 +1,41 @@
...
+            stdin=devnull,
+            stdout=devnull,
+            stderr=stderr,
+        );
+        wait=false,
+    )
+end
+
+# Check that all examples were run successfully
+if !all(success, processes)
```
willtebbutt, line 25: Does calling `success` here wait until the process has completed? (I'm just trying to ensure that I've understood this correctly).

i.e. by calling `all(success, processes)`, are you implicitly ensuring that all of the processes have completed _and_ then check that they returned `0`?

examples/svm/script.jl
```diff
@@ -0,0 +1,41 @@
...
+
+# Create a 2D grid:
+xgrid = range(floor(Int, minimum(X)), ceil(Int, maximum(X)); length=100)
+Xgrid = ColVecs(mapreduce(collect, hcat, Iterators.product(xgrid, xgrid)));
+
+# Create kernel function:
+k = SqExponentialKernel() ∘ ScaleTransform(2.0)
+
+# Optimal prediction:
+function f(x, X, k, λ)
```
willtebbutt, line 35: Maybe put this on a single line?

examples/svm/script.jl
```diff
@@ -0,0 +1,41 @@
...
+    return kernelmatrix(k, x, X) / (kernelmatrix(k, X) + exp(λ) * I) * y
```
willtebbutt, line 36: Also, is this not just kernel ridge regression, rather than support vector regression? Possibly I'm forgetting something basic here, because I've not looked at SVM stuff for many years...

willtebbutt (commented):
Some minor comments. Would be helpful for others to review as well, as they've spent more time looking at this than I.

docs/make.jl
willtebbutt: Great, thanks for clarifying for me.

examples/svm/script.jl
theogf: Haha yes I had the exact same comment on Slack 😅

devmotion: Haha yeah I just copied and fixed the existing example to show that the setup works without thinking about the content at all :smile: Maybe we should just remove the example?

theogf: That's why we we switched from SVM to kernel ridge regression on the other PR

devmotion: This "SVM" example is also part of #303 (which is less optimized/updated than the version here). There exists another kernel ridge regression and a deep kernel learning example that are also updated in #303.

devmotion: Therefore my suggestion would be to just completely remove the SVM example (and also not to keep the current version). The current version showed that the setup works and other examples can then be updated in separate PRs.

st--: :woman_shrugging: happy to delete it as well and only put it back once we've done it properly. I would just not get too hung up on the contents of the example notebooks, let's get the doc build in order, and then we can rescue the actual content from #234

.github/workflows/CompatHelper.yml
```diff
@@ -1,16 +1,33 @@
...
+          import Pkg
+          name = "CompatHelper"
+          uuid = "aa819f21-2bde-4658-8897-bab36330d9b7"
+          version = "2"
+          Pkg.add(; name, uuid, version)
+        shell: julia --color=yes {0}
+      - name: "Run CompatHelper"
+        run: |
+          import CompatHelper
+          subdirs = ["", "test", "docs"]
```
st--, line 23: What does the empty `""` do here? If it's referring to the repo base directory, why would it need to be passed to _sub_dirs?

.github/workflows/CompatHelper.yml
```diff
@@ -1,16 +1,33 @@
...
     runs-on: ubuntu-latest
     steps:
-      - name: Pkg.add("CompatHelper")
-        run: julia -e 'using Pkg; Pkg.add("CompatHelper")'
-      - name: CompatHelper.main()
+      - name: "Install CompatHelper"
+        run: |
+          import Pkg
+          name = "CompatHelper"
+          uuid = "aa819f21-2bde-4658-8897-bab36330d9b7"
```
st--, line 16: Why do you want to change to pinning CompatHelper? is there any reason not to want the latest version (with potential bugfixes etc)?

.github/workflows/docs.yml
```diff
@@ -19,6 +19,7 @@ jobs:
         run: julia --project=docs/ -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate()'
       - name: Build and deploy
         env:
+          GKSwstype: nul
```
st--, line 22:
```suggestion
          GKSwstype: nul # turn off GR's interactive plotting for notebooks
```

docs/make.jl
```diff
@@ -1,12 +1,41 @@
...
+    error("some examples were not run successfully")
```
st--, line 26: Would there be any way of indicating _which_ script failed ?

docs/make.jl
```diff
@@ -31,8 +60,11 @@ makedocs(;
         "metrics.md",
         "create_kernel.md",
         "API" => "api.md",
-        "Examples" => "example.md",
         "Design" => "design.md",
+        "Examples" =>
+            map(filter!(filename -> endswith(filename, ".md"), readdir(EXAMPLES_OUT))) do x
+                return joinpath("examples", x)
+            end,
```
st--, line 67: nice:) much easier to read than the other version

docs/literate.jl
```diff
@@ -0,0 +1,56 @@
+# Retrieve filename of literate script
+if length(ARGS) != 2
+    error("please specify the literate script and the output directory")
+end
+const SCRIPTJL = ARGS[1]
+const OUTDIR = ARGS[2]
+
+# Activate environment
+using Pkg: Pkg
+Pkg.activate(dirname(SCRIPTJL))
```
st--, line 10: Requiring the examples scripts to include Literate in their dependencies when they don't actually use it seems weird; maybe that's the best we can do (and of course we can then clarify that in the examples/README.md), but why not add them here as well:
```suggestion
Pkg.activate(dirname(SCRIPTJL))
Pkg.add("Literate")
```
this should add it to the correct Project.toml (e.g. when I write a new examples script and locally run make.jl), but be a no-op if it's already there.

.github/workflows/CompatHelper.yml
devmotion: The UUID is just an additional guarantee that it is the correct package but does not fix the version. The version is only pinned to 2.x.x but no specific version, i.e., it has to be updated only if there is a breaking release of CompatHelper.

In general, this is just copied from the documentation of CompatHelper: https://juliaregistries.github.io/CompatHelper.jl/dev/

docs/make.jl
devmotion: I redirected stderr of the processes to the default stderr, and hence errors of the individual processes are printed in the terminal regardless of `success`. The check here mainly ensures that all processes are done and that the script errors if a process failed.

docs/literate.jl
devmotion: I thought about this as well and I am still not sure what's the best approach. The advantage of adding them to the Project.toml + Manifest.toml is that it is correctly tracked and updated by CompatHelper and that it is guaranteed that the installation of Literate does not downgrade any other packages or mess with the package setup which would impact reproducibility if it is not part of Project.toml + Manifest.toml (otherwise it might only be installed when the CI tests are run).

.github/workflows/CompatHelper.yml
devmotion: Otherwise the base directory is not updated anymore. The default setting is `subdirs = [""]`: https://juliaregistries.github.io/CompatHelper.jl/dev/options/#Packages-in-subdirectories

examples/svm/script.jl
devmotion: Done :+1:

examples/svm/script.jl
devmotion: I moved the updated version back to `examples/svm.jl` since it is still an improvement even though it's not actually SVM :smile:

willtebbutt (approved):
This looks good to me now.

Perhaps renaming the svm file to `kernel_ridge_regression.jl` or something would make sense while we're here though 🤷

docs/literate.jl
st--: Would
```suggestion
Pkg.activate(dirname(SCRIPTJL))
Pkg.add("Literate"; preserve=PRESERVE_ALL)
```
or similar ([Pkg.add docs](https://pkgdocs.julialang.org/v1/api/#Pkg.add)) not satisfy that?

st-- (approved):
Looks good to me; I would still move the Literate dependency from examples scripts themselves into the literate.jl script, but happy to propose that change in #303 after rebasing and continuing the discussion there:)

PR #316, opened 2021-06-30
Example notebook: Gaussian process priors

st--:


devmotion:
Some high-level comment/idea: Since this is not how we want users to create or work with Gaussian processes, I think it would be nice to have to both the "naive"/basic KernelFunctions approach and the AbstractGP-based implementation together in an example to avoid any incorrect impressions and so that users know how they should do it. However, if both AbstractGPs and KernelFunctions are part of the example, I think the example fits better into either the AbstractGPs repo or the JuliaGaussianProcesses webpage. Otherwise we use packages that depend on KernelFunctions and hence introduce a circular dependency that makes it difficult to update the example without delay if there are breaking changes in KernelFunctions.

st--:
How about leaving this example in KernelFunctions but clarifying all you're saying in an intro paragraph and linking to the corresponding "proper" example in AbstractGPs docs?

devmotion:
This should definitely be done if the example is added to KernelFunctions. My worry is that people do not notice and read the paragraph and just copy whatever code there is on this page about Gaussian processes. I assume this is less of a problem with the "correct" code is on the same page as well.

st--:
> My worry is that people do not notice and read the paragraph and just copy whatever code there is on this page about Gaussian processes. 

It might be helpful if you expand on what you're worried about so we can discuss likelihood, impact, and mitigations:)?

devmotion:
I am worried that they copy and use the code but are not aware of how one is supposed to work with Gaussian processes.

willtebbutt:
In my opinion it would be sufficient to add one of those hint things that you can add in documenter saying "this is an illustrative example, you should probably be using an AbstractGP of some kind or another if you actually want to do something with a GP"

devmotion:
On the other hand, we can always link to AbstractGPs or the JuliaGaussianProcesses webpage from the documentation, so the example would still be discoverable from the KernelFunctions docs even if it is part of AbstractGPs or the webpage.

examples/gaussian-process-priors/script.jl
```diff
@@ -0,0 +1,109 @@
...
+
+function visualize(k::Kernel; xref=0.0)
+    K = kernelmatrix(k, X)
+    f = mvn_sample(K)
+
+    p_kernel_2d = heatmap(
+        K;
+        yflip=true,
+        colorbar=false,
+        ylabel=string(nameof(typeof(k))),
```
st--, line 66: this doesn't actually show up in the plot, any idea how to get it to display ?

devmotion: I guess it might be disabled due to `framestyle=:none`? Related to this, could we add labels `x` and `x'` along the axes instead of the name of the kernel?

st--: I'd be very happy with that if we can figure out how to otherwise get the kernel label on the left hand side (so that it works better as a table with rows for the different kernels)!

st--: I've added the `x` label; I can't get Plots.jl to process the inline LaTeX if I want to have the kernel name in one line and the `x'` in a second line in the ylabel. I think it's clear enough as it is though, with the ticklabels at least and the heatmaps all being symmetric, so I think this is good enough for now

examples/gaussian-process-priors/script.jl
```diff
@@ -0,0 +1,130 @@
...
+xlim = (-5, 5)
+X = range(xlim...; length=num_inputs);
+
+# Given a kernel `k`, we can compute the kernel matrix as `K = kernelmatrix(k, X)`.`
+
+##
+# ## Random samples
+#
+# To sample from the multivariate normal distribution $p(\mathbf{f}) = \mathcal{MVN}(0, \mathrm{K})$, we could make use of Distributions.jl and call `rand(MvNormal(K))`.
+# Alternatively, we could use the AbstractGPs.jl package and construct a `GP` object which we evaluate at the points of interest and from which we can then sample: `rand(GP(k)(X))`.
```
theogf, line 41:
```suggestion
# Alternatively, we could use the [AbstractGPs.jl](https://github.com/JuliaGaussianProcesses/AbstractGPs.jl) package and construct a `GP` object which we evaluate at the points of interest and from which we can then sample: `rand(GP(k)(X))`.
```

examples/gaussian-process-priors/script.jl
```diff
@@ -0,0 +1,130 @@
...
+    L = cholesky(K + 1e-6 * I)
+    f = L.L * v
+    return f
+end;
+
+##
+# ## Visualization
+# We now define a function that visualizes a kernel for us.
+
+function visualize(k::Kernel; xref=0.0)
```
theogf, line 61: How about having two xref ? like 0.0 and 1.0. This would make the `LinearKernel` and `Neural...` more interesting to visualize

theogf: And I mean to have them on the same plot

st--: it would be quite boring for all the stationary kernels. how about simply changing the xref to 1.0 for the linear and NN kernels?

st--: done

examples/gaussian-process-priors/script.jl
```diff
@@ -0,0 +1,128 @@
...
+# The function values $\mathbf{f} = \{f(x_n)\}_{n=1}^N$ of the GP at a finite number $N$ of points $X = \{x_n\}_{n=1}^N$ follow a multivariate normal distribution $\mathbf{f} \sim \mathcal{MVN}(\mathbf{m}, \mathrm{K})$ with mean vector $\mathbf{m}$ and covariance matrix $\mathrm{K}$, where
+# ```math
+# \begin{aligned}
+#   \mathbf{m}_i &= m(x_i) \\
+#   \mathrm{K}_{i,j} &= k(x_i, x_j)
+# \end{aligned}
+# ```
+# with $1 \le i, j \le N$.
+#
+# We can visualize the infinite-dimensional GP by evaluating it on a fine grid to approximate the dense real line:
```
willtebbutt, line 30: What is meant by "the dense real line" here?

willtebbutt (dismissed):
LGTM. Just a quick clarification.

examples/gaussian-process-priors/script.jl
st--: the "full" (uncountably infinite) `\mathcal{R}` (in contrast to the finite `X \subset \mathcal{R}` with `|X| = N`)

theogf (dismissed):

willtebbutt (approved):

PR #317, opened 2021-06-30
Example notebook: training kernel parameters

st--:


willtebbutt:
@st-- how would you like to progress with this PR? The example generally looks reasonable to me, so I'm happy to review it for details if that's what you'd like. Just wondering if there's any specific kind of feedback that you're interested in.

Crown421:
I have finally managed to work on this and have something that could be nice. I am not sure if I can push to this PR, or should/need to open a new one?

willtebbutt:
I would imagine it would be safer to implement in a new PR, unless @st-- thinks otherwise?

st--:
#430 is much better, thanks @Crown421 !

PR #320, opened 2021-07-01
Example notebook: kernel ridge regression

st--:


examples/kernel-ridge-regression/script.jl
```diff
@@ -1,66 +1,159 @@
...
 
+plot([regularized_fit_and_plot(18, lambda) for lambda in [1e-4, 1e-2, 0.1, 10]]...)
+
+# Instead of constructing the feature matrix explicitly, we can use *kernels* to replace inner products of feature vectors with a kernel evaluation: $\langle \phi(x), \phi(x') \rangle = k(x, x')$ or $\mathrm{X} \mathrm{X}^\top = \mathrm{K}$, where $\mathrm{K}_{ij} = k(x_i, x_j)$.
+#
+# To apply this "kernel trick" to ridge regression, we can rewrite the ridge estimate for the weights
+# $$
+# \mathbf{w} = (\mathrm{X}^\top \mathrm{X} + \lambda \mathbb{1})^{-1} \mathrm{X}^\top \mathbf{y}
+# $$
+# using the [matrix inversion lemma](https://tlienart.github.io/pub/csml/mtheory/matinvlem.html#basic_lemmas)
```
devmotion, line 112: Maybe rather link to the Wikipedia article about the [Woodbury matrix identity](https://en.wikipedia.org/wiki/Woodbury_matrix_identity)?

willtebbutt (dismissed):
I have no objection to this being merged once CI is made to pass.

examples/kernel-ridge-regression/script.jl
```diff
@@ -1,66 +1,163 @@
...
+plot!(x_test, y_pred; label="linear fit")
 
-loss(k, λ) = (ŷ -> sum(y - ŷ) / length(y) + exp(λ[1]) * norm(ŷ))(f(X, k, λ))
-loss(k, λ)
+# ## Featurization
+# We can improve the fit by including additional features, i.e. generalizing to $\mathrm{X} = (\phi(x_n))_{n=1}^N$, where $\phi(x)$ constructs a feature vector for each input $x$. Here we include powers of the input, $\phi(x) = (1, x, x^2, \dots, x^d)$:
 
+function featurize_poly(x; degree=1)
+    xcols = [x .^ d for d in 0:degree]
+    return hcat(xcols...)
```
devmotion, line 59: A bit simpler and without splatting:
```suggestion
    return mapreduce(d -> x .^ d, hcat, 0:degree)
```
Alternatively
```suggestion
    return repeat(x, 1, degree + 1) .^ (0:degree)'
```
or without additional allocations

```suggestion
    y = repeat(x, 1, degree + 1)
    y .= y .^ (0:degree)
    return y
```

examples/kernel-ridge-regression/script.jl
```diff
@@ -1,66 +1,163 @@
...
 #
-# !!! warning
-#     This example is under construction
-
-# Setup
+# Building on linear regression, we can fit non-linear data sets by introducing a feature space. In a higher-dimensional feature space, we can overfit the data; ridge regression introduces regularization to avoid this. In this notebook we show how we can use KernelFunctions.jl for *kernel* ridge regression.
 
+## Loading and setup of required packages
 using KernelFunctions
-using MLDataUtils
```
devmotion, line 9: Can you update the Project.toml and Manifest.toml files as well?

examples/kernel-ridge-regression/script.jl
```diff
@@ -1,66 +1,163 @@
...
+seed!(42);
 
-#
+# ## Toy data
+# Here we use a one-dimensional toy problem. We generate data using the fourth-order polynomial $f(x) = (x+4)(x+1)(x-1)(x-3)$:
+
+f_truth(x) = (x + 4) * (x + 1) * (x - 1) * (x - 3)
+
+x_train = collect(-5:0.5:5)
+x_test = collect(-5:0.1:5)
```
devmotion, line 23: It seems a bit strange that we can't work with the ranges?

examples/kernel-ridge-regression/script.jl
```diff
@@ -1,66 +1,163 @@
...
+
+noise = rand(Uniform(-10, 10), size(x_train))
```
devmotion, line 25:
```suggestion
noise = rand(Uniform(-10, 10), length(x_train))
```

examples/kernel-ridge-regression/script.jl
```diff
@@ -1,66 +1,163 @@
...
+
+function featurized_fit_and_plot(degree)
+    X = featurize_poly(x_train; degree=degree)
+    Xstar = featurize_poly(x_test; degree=degree)
+    y_pred = linear_regression(X, y_train, Xstar)
+    scatter(x_train, y_train; legend=false, title="fit of order $degree")
+    return plot!(x_test, y_pred)
+end
+
+plot([featurized_fit_and_plot(degree) for degree in 1:4]...)
```
devmotion, line 70:
```suggestion
plot((featurized_fit_and_plot(degree) for degree in 1:4)...)
```

examples/kernel-ridge-regression/script.jl
devmotion: I still think it would be more natural to link to Wikipedia?

examples/kernel-ridge-regression/script.jl
```diff
@@ -1,66 +1,163 @@
...
-    push!(plots, p)
+function regularized_fit_and_plot(degree, lambda)
+    X = featurize_poly(x_train; degree=degree)
+    Xstar = featurize_poly(x_test; degree=degree)
+    y_pred = ridge_regression(X, y_train, Xstar, lambda)
+    scatter(x_train, y_train; legend=false, title="\$\\lambda=$lambda\$")
+    return plot!(x_test, y_pred)
 end
 
+plot([regularized_fit_and_plot(18, lambda) for lambda in [1e-4, 1e-2, 0.1, 10]]...)
```
devmotion, line 107:
```suggestion
plot((regularized_fit_and_plot(18, lambda) for lambda in (1e-4, 1e-2, 0.1, 10))...)
```

examples/kernel-ridge-regression/script.jl
```diff
@@ -1,66 +1,163 @@
...
+    else
+        title = string(nameof(typeof(kernel)))
+    end
+    scatter(x_train, y_train; label=nothing)
+    p = plot!(
+        x_test,
+        y_pred;
+        label=nothing,
+        title=title,
+        #title=string(raw"$\lambda=", lambda, raw"$")
```
devmotion, line 154:
```suggestion
```

examples/kernel-ridge-regression/script.jl
```diff
@@ -1,66 +1,163 @@
...
+# Now, instead of explicitly constructing features, we can simply pass in a `PolynomialKernel` object:
+
+function kernelized_fit_and_plot(kernel, lambda=1e-4)
+    y_pred = kernel_ridge_regression(kernel, x_train, y_train, x_test, lambda)
+    if kernel isa PolynomialKernel
+        title = string("order ", kernel.degree)
+    else
+        title = string(nameof(typeof(kernel)))
+    end
+    scatter(x_train, y_train; label=nothing)
```
devmotion, line 148: I didn't know that `nothing` works, I always use `label=""` to hide the label for a specific series. However, it seems you want to remove labels for all series here in which caseI guess you can just use `legend=false`

```suggestion
    scatter(x_train, y_train; legend=false)
```
is simpler.

examples/kernel-ridge-regression/script.jl
```diff
@@ -1,66 +1,163 @@
...
+    p = plot!(
+        x_test,
+        y_pred;
+        label=nothing,
```
devmotion, line 152: This is not needed if `legend=false`:

```suggestion
```

examples/kernel-ridge-regression/script.jl
```diff
@@ -1,66 +1,163 @@
...
+        title=title,
+        #title=string(raw"$\lambda=", lambda, raw"$")
+    )
+    return p
+end
+
+plot([kernelized_fit_and_plot(PolynomialKernel(; degree=degree, c=1)) for degree in 1:4]...)
```
devmotion, line 159:
```suggestion
plot((kernelized_fit_and_plot(PolynomialKernel(; degree=degree, c=1)) for degree in 1:4)...)
```

examples/kernel-ridge-regression/script.jl
st--: good catch, done

examples/kernel-ridge-regression/script.jl
st--: I found the blog post more instructive/helpful than the wikipedia article (and the latter is straightforward to find for anyone who wants it)..

examples/kernel-ridge-regression/script.jl
st--: huh, maybe we can, I'll check

theogf: Additionally, it would be nice to extend the limits of `x_test` so see some extrapolation

st--: done

examples/kernel-ridge-regression/script.jl
```diff
@@ -1,66 +1,162 @@
...
+    y_pred = ridge_regression(X, y_train, Xstar, lambda)
+    scatter(x_train, y_train; legend=false, title="\$\\lambda=$lambda\$")
+    return plot!(x_test, y_pred)
 end
 
+plot((regularized_fit_and_plot(20, lambda) for lambda in (1e-3, 1e-2, 1e-1, 1))...)
+
+##
+# ## Kernel ridge regression
+# Instead of constructing the feature matrix explicitly, we can use *kernels* to replace inner products of feature vectors with a kernel evaluation: $\langle \phi(x), \phi(x') \rangle = k(x, x')$ or $\mathrm{X} \mathrm{X}^\top = \mathrm{K}$, where $\mathrm{K}_{ij} = k(x_i, x_j)$.
```
theogf, line 115: Here talking about XX^\top=K is a bit misleading, it does not sound like we are using any feature mapping, maybe \phi(X)\phi(X)^\top

st--: well, `X` is the "extended" `x` with feature columns, each column of X is one \phi applied to the input x

st--: but it's a bit ambiguous, you're right. maybe call "featurized X" `\tilde{X}` instead?

st--: NB: `XX' = K` is exact for the LinearKernel k(x,x') = xx' :)

theogf (approved):

examples/kernel-ridge-regression/script.jl
theogf: With no constant :)

PR #322, opened 2021-07-01
Example notebook: deep kernel learning

st--:


examples/deep-kernel-learning/script.jl
```diff
@@ -1,68 +1,78 @@
...
+# We use a couple of useful packages to plot and optimize
+# the different hyper-parameters
 using KernelFunctions
-using MLDataUtils
-using Zygote
 using Flux
 using Distributions, LinearAlgebra
 using Plots
+using ProgressMeter
+using AbstractGPs
```
devmotion, line 10: This also introduces a circular dependency and hence leads to the same problems as mentioned in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/316#issuecomment-871452800. It might be better to move it to AbstractGPs or the JuliaGaussianProcesses webpage. We can always link to these examples from the documentation, so it would still be possible to make it discoverable from the KernelFunctions docs.

st--: Now that each example has its own project environment & pinning, shouldn't this no longer be a concern? The example depends on KernelFunctions and AbstractGPs, and AbstractGPs depends on KernelFunctions, but neither KernelFunctions nor AbstractGPs depend on the example.

devmotion: > AbstractGPs depends on KernelFunctions

This is exactly the problem and the circular dependency I was referring to. If we make breaking changes in KernelFunctions, the example will break - there is no version of AbstractGPs that is compatible with this version yet. Hence either we have to switch to an old version of KernelFunctions and there will be a disconnect between the version of KernelFunctions discussed in the documentation and used in the example until AbstractGPs is updated and the example is fixed or we have to remove the example for a while. Both alternatives seem a bit annoying.

This problem is the main reason why I think it is unfortunate to use downstream packages in the documentation of KernelFunctions and these examples should rather go into AbstractGPs where we would not have to deal with these problems.

st--: A breaking change doesn't imply the example *must* break, just that it's not guaranteed to not break... In any case, I'm fine with moving this notebook over to AbstractGPs. Just wanted to save the current state out of #234.

devmotion: Often the example itself won't break but this is not relevant: AbstractGPs is just not compatible with KernelFunctions before and when (and possibly for some time after) a breaking release of KernelFunctions is made. A new release of AbstractGPs is needed with updated compatibility bounds and possibly some fixes, even if the example itself does not require any changes.

Issue #315, opened 2021-06-30
Parallelise tests in build


st--:
> Currently the rest of the build takes ~40 minutes anyways so that shouldn't make it longer overall.

IMO this is a motivation for splitting the tests into different groups that are run in parallel (similar what we do in the SciML ecosystem: https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/test/runtests.jl) rather than running the examples in serial.

_Originally posted by @devmotion in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/303#issuecomment-870543998_

4aHxKzD:
Related to but a different issue is not being able to run sections of the tests separately through the command line. I`ve been thinking of implementing the code that allows to use a [`Pkg.Test` feature](https://github.com/JuliaLang/Pkg.jl/pull/1226) that was added for this purpose. I think with this potential contributors will have an easier time when experimenting stuff to add to the package.

4aHxKzD:
Maybe both ideas can be implemented at the same time?

devmotion:
This is already part of my suggestion above. You would just set e.g. `ENV["GROUP"] = "BaseKernels"`.

4aHxKzD:
Oh that also works with comand line arguments?

devmotion:
There are no command line arguments involved here. It is based on an environment variable that you can set in your shell before starting Julia or in the Julia REPL with `ENV[...] = ...` or `withenv(... => ...) do ....`.

PR #323, opened 2021-07-01
Break tests into different groups

devmotion:
Fixes https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/315.

Probably it can/should be optimized what groups and in particular how many of them we use.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -19,7 +19,9 @@ using KernelFunctions: SimpleKernel, metric, kappa, ColVecs, RowVecs, TestUtils
 
 using KernelFunctions.TestUtils: test_interface
 
-const GROUP = get(ENV, "GROUP", "All")
+# The GROUP is used to run different sets of tests in parallel on the GitHub Actions CI.
+# If you want to introduce a new group, ensure you also add it to .github/workflows/ci.yml
+const GROUP = get(ENV, "GROUP", "")
 
 # Writing tests:
 # 1. The file structure of the test should match precisely the file structure of src.
@@ -56,7 +58,7 @@ const GROUP = get(ENV, "GROUP", "All")
 include("test_utils.jl")
 
 @testset "KernelFunctions" begin
-    if GROUP == "All" || GROUP == "Transform"
+    if GROUP == "" || GROUP == "Transform"
         @testset "transform" begin
             include(joinpath("transform", "transform.jl"))
             print(" ")
@@ -78,7 +80,7 @@ include("test_utils.jl")
         @info "Ran tests on Transform"
     end
 
-    if GROUP == "All" || GROUP == "BaseKernels"
+    if GROUP == "" || GROUP == "BaseKernels"
         @testset "basekernels" begin
             include(joinpath("basekernels", "constant.jl"))
             print(" ")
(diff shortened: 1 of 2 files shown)


devmotion:
Maybe we could also drop tests on MacOS and Windows? There are no OS-specific features in KernelFunctions, so I don't think the additional tests provide any additional information. And if there are actually OS-specific bug reports at some we can still reevaluate.

willtebbutt:
> Maybe we could also drop tests on MacOS and Windows?

I am in favour of this.

st-- (approved):
Looks good to me. Just a thought, to minimise the risk of someone at some point introducing a typo in `"All"`, how about changing it to `GROUP = get(ENV, "GROUP", "")` and then simply testing `if GROUP == "" || ...`? The intent is the same ("" == "no specific group").

test/runtests.jl
```diff
@@ -19,6 +19,8 @@ using KernelFunctions: SimpleKernel, metric, kappa, ColVecs, RowVecs, TestUtils
 
 using KernelFunctions.TestUtils: test_interface
 
+const GROUP = get(ENV, "GROUP", "All")
```
st--, line 22: I can't comment on the test-explanation below, but I'd add a comment along the lines of
```suggestion
const GROUP = get(ENV, "GROUP", "All")
# The GROUP is used to run different sets of tests in parallel on the GitHub Actions CI. If you want to introduce a new group, ensure you also add it to .github/workflows/ci.yml
```

test/runtests.jl
```diff
@@ -54,123 +56,135 @@ using KernelFunctions.TestUtils: test_interface
...
+            include(joinpath("distances", "dotproduct.jl"))
+            include(joinpath("distances", "delta.jl"))
+            include(joinpath("distances", "sinus.jl"))
+        end
+        @info "Ran tests on Distances"
 
-    @testset "approximations" begin
-        include(joinpath("approximations", "nystrom.jl"))
-    end
+        if GROUP == "All" || GROUP == "Matrix"
```
st--, line 151: This doesn't seem to make sense; to execute this line, GROUP can never be Matrix

devmotion: Oh I forgot to remove this line when I merged the two groups :grimacing:

st-- (approved):

devmotion:
@st-- I noticed that the coveralls requires some explicit handling of the parallel runs, otherwise the coverage reports are completely messed up. I just updated the action according to https://github.com/coverallsapp/github-action#complete-parallel-job-example. Can you re-approve if you are happy with the PR?

st-- (approved):
I'm happy with it but there still seem to be some test issues...

devmotion:
The test error with the Julia nightly is caused by https://github.com/JuliaArrays/AxisArrays.jl/issues/200 and not related to KernelFunctions.

st--:
@devmotion why would coverage have decreased? does this mean some tests got missed out in the split ?

devmotion:
I don't think the drop is real, I can't find any coverage changes in the more detailed output of coveralls. I assume it is just related to the change from non-parallel to parallel tests and uploads.

PR #324, opened 2021-07-01
Work around https://github.com/JuliaLang/Pkg.jl/issues/2219

devmotion:
@st-- I finally figured out what the problem with CUDA was and why the documentation fails to build on the master branch: Pkg does not have a global lock and hence it is not safe to install packages in parallel (https://github.com/JuliaLang/Pkg.jl/issues/2219). This PR should fix the issues.

I also changed the literate.jl script a bit to avoid inconsistencies. Now the name of the example has to be provided which ensures that the preprocessing is always correct.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -22,4 +22,5 @@ jobs:
           GKSwstype: nul # turn off GR's interactive plotting for notebooks
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # For authentication with GitHub Actions token
           DOCUMENTER_KEY: ${{ secrets.DOCUMENTER_KEY }} # For authentication with SSH deploy key
+          JULIA_DEBUG: Documenter # Print `@debug` statements (https://github.com/JuliaDocs/Documenter.jl/issues/955)
         run: julia --project=docs/ docs/make.jl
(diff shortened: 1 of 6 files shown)


docs/literate.jl
```diff
@@ -1,15 +1,16 @@
...
 
 # Activate environment
-using Pkg: Pkg
-Pkg.activate(dirname(SCRIPTJL))
 # Note that each example's Project.toml must include Literate as a dependency
-Pkg.instantiate()
+using Pkg: Pkg
+const EXAMPLEPATH = joinpath(@__DIR__, "..", "examples", EXAMPLE)
+Pkg.activate(EXAMPLEPATH)
+
```
devmotion, line 13:
```suggestion
Pkg.instantiate()
```

.github/workflows/docs.yml
```diff
@@ -22,4 +22,5 @@ jobs:
           GKSwstype: nul # turn off GR's interactive plotting for notebooks
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # For authentication with GitHub Actions token
           DOCUMENTER_KEY: ${{ secrets.DOCUMENTER_KEY }} # For authentication with SSH deploy key
+          JULIA_DEBUG: Documenter # Print `@debug` statements (https://github.com/JuliaDocs/Documenter.jl/issues/955)
```
st--, line 25: much neater :+1:

st-- (approved):
Great! Thanks for figuring this out:)

PR #333, opened 2021-07-08
Revert "CompatHelper: bump compat for "FillArrays" to "0.12""

devmotion:
FillArrays 0.12 was not tested, it seems some dependency holds it back.


st-- (approved):
Woops, hadn't spotted that it didn't actually install 0.12 for the tests!

Issue #107, opened 2020-04-27
Make ARDTransform a subtype of LinearTransform?


devmotion:
As discussed in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/106, the ARDTransform could be viewed as a special LinearTransform with a diagonal matrix. Do we still want to keep these separate types?

theogf:
I would strongly push for keeping `ARDTransform` separate from `LinearTransform`.
It is indeed a special case but it is a very used one. In the same sense that `SqExponentialKernel` is just a special case of the `GammaSqExponentialKernel`

willtebbutt:
I agree that users should have easy access an ARD transform, but I think it's very debatable whether or not this means we need to maintain an entirely separate transform from the `LinearTransform` unless there are good performance reasons to do so.

We could just alias `const ARDTransform = LinearTransform{<:Diagonal}` (or whatever the correct parametrisation is) and no longer have to maintain the corresponding code as a consequence.

For me at least, the difference between this and the `SqExponentialKernel` case is that here whether or not something is a ARD is determined by the type of its argument, so we know everything at compile time and can specialise appropriately using dispatch.

devmotion:
I agree. IMO there's no particular reason for not unifying the underlying implementation (which a user shouldn't deal with or be concerned about). As @willtebbutt says, one can introduce an alias and still provide constructors for vectors to users, even if one makes ARDTransform a subtype of LinearTransform.

theogf:
I am not sure I see the point of it but okay...

PR #335, opened 2021-07-09
with_lengthscale

### LINKED ISSUE #326: lengthscale transform
There is currently ScaleTransform, ARDTransform, and LinearTransform. These could all be unified under a LinearTransform with potentially specialised matrices (e.g. a scaled identity matrix or a diagonal matrix for ScaleTransform or ARDTransform, respectively).

Moreover, other kernel/GP packages generally parametrise the kernels with the rather more interpretable _lengthscale_ rather than the inverse lengthscale as KernelFunctions.jl does at the moment: it would be good to have an easy way for users to directly specify a lengthscale (ARD or not).

st--:
To allow for a more natural construction of kernels. Provides the solution discussed in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/217#issuecomment-760995803 and resolves #326 (the other half of #326 is a duplicate of #107...)

### DIFF
--- a/src/kernels/transformedkernel.jl
+++ b/src/kernels/transformedkernel.jl
@@ -72,6 +72,10 @@ Base.:∘(k::TransformedKernel, t::Transform) = TransformedKernel(k.kernel, k.tr
 Base.:∘(k::Kernel, ::IdentityTransform) = k
 Base.:∘(k::TransformedKernel, ::IdentityTransform) = k
 
+function Base.isequal(k::TransformedKernel, k2::TransformedKernel)
+    return isequal(k.kernel, k2.kernel) && isequal(k.transform, k2.transform)
+end
+
 Base.show(io::IO, κ::TransformedKernel) = printshifted(io, κ, 0)
 
 function printshifted(io::IO, κ::TransformedKernel, shift::Int)
--- a/src/transform/ardtransform.jl
+++ b/src/transform/ardtransform.jl
@@ -15,6 +15,7 @@ true
 struct ARDTransform{Tv<:AbstractVector{<:Real}} <: Transform
     v::Tv
 end
+Base.==(a::ARDTransform, b::ARDTransform) = a.v == b.v
 
 """
     ARDTransform(s::Real, dims::Integer)
--- a/src/transform/ardtransform.jl
+++ b/src/transform/ardtransform.jl
@@ -15,7 +15,7 @@ true
 struct ARDTransform{Tv<:AbstractVector{<:Real}} <: Transform
     v::Tv
 end
-Base.==(a::ARDTransform, b::ARDTransform) = a.v == b.v
+Base .== (a::ARDTransform, b::ARDTransform) = a.v == b.v
 
 """
     ARDTransform(s::Real, dims::Integer)
--- a/src/transform/ardtransform.jl
+++ b/src/transform/ardtransform.jl
@@ -15,7 +15,6 @@ true
 struct ARDTransform{Tv<:AbstractVector{<:Real}} <: Transform
     v::Tv
 end
-Base .== (a::ARDTransform, b::ARDTransform) = a.v == b.v
 
 """
     ARDTransform(s::Real, dims::Integer)
--- a/src/transform/scaletransform.jl
+++ b/src/transform/scaletransform.jl
@@ -15,7 +15,7 @@ true
 struct ScaleTransform{T<:Real} <: Transform
     s::Vector{T}
 end
-Base.==(a::ScaleTransform, b::ScaleTransform) = a.s == b.s
+Base .== (a::ScaleTransform, b::ScaleTransform) = a.s == b.s
 
 function ScaleTransform(s::T=1.0) where {T<:Real}
     return ScaleTransform{T}([s])
--- a/src/transform/scaletransform.jl
+++ b/src/transform/scaletransform.jl
@@ -15,7 +15,6 @@ true
 struct ScaleTransform{T<:Real} <: Transform
     s::Vector{T}
 end
-Base .== (a::ScaleTransform, b::ScaleTransform) = a.s == b.s
 
 function ScaleTransform(s::T=1.0) where {T<:Real}
     return ScaleTransform{T}([s])
--- a/src/transform/scaletransform.jl
+++ b/src/transform/scaletransform.jl
@@ -15,6 +15,7 @@ true
 struct ScaleTransform{T<:Real} <: Transform
     s::Vector{T}
 end
+Base.==(a::ScaleTransform, b::ScaleTransform) = a.s == b.s
 
 function ScaleTransform(s::T=1.0) where {T<:Real}
     return ScaleTransform{T}([s])
--- a/src/transform/with_lengthscale.jl
+++ b/src/transform/with_lengthscale.jl
@@ -23,11 +23,10 @@ julia> ℓ = [0.5, 2.5];
 julia> isequal(SqExponentialKernel() ∘ ARDTransform(inv.(ℓ)), with_lengthscale(SqExponentialKernel(), ℓ))
 true
 ```
-
 """
 function with_lengthscale(kernel::Kernel, lengthscale::Real)
     return compose(kernel, ScaleTransform(inv(lengthscale)))
 end
 function with_lengthscale(kernel::Kernel, lengthscales::AbstractVector{<:Real})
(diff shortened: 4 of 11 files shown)


src/transform/with_lengthscale.jl
```diff
@@ -0,0 +1,34 @@
+"""
+    with_lengthscale(kernel::Kernel, lengthscale::Real)
+    with_lengthscale(kernel::Kernel, lengthscales::AbstractVector{<:Real})
+
+Constructs a transformed kernel with `lengthscale`. If `lengthscales` is passed
+as a vector instead, constructs an "ARD" kernel with different lengthscales for
+each dimension.
```
willtebbutt, line 7:
```suggestion
Constructs a transformed kernel with `lengthscale`.
If a vector `lengthscales` is passed instead, constructs an "ARD" kernel with different lengthscales for each dimension.
```
Minor grammar thing -- I think this way makes more sense to me.

src/transform/ardtransform.jl
```diff
@@ -15,6 +15,7 @@ true
 struct ARDTransform{Tv<:AbstractVector{<:Real}} <: Transform
     v::Tv
 end
+Base .== (a::ARDTransform, b::ARDTransform) = a.v == b.v
```
devmotion, line 18:
```suggestion
Base.:(==)(a::ARDTransform, b::ARDTransform) = a.v == b.v
```

src/transform/scaletransform.jl
```diff
@@ -15,6 +15,7 @@ true
 struct ScaleTransform{T<:Real} <: Transform
     s::Vector{T}
 end
+Base .== (a::ScaleTransform, b::ScaleTransform) = a.s == b.s
```
devmotion, line 18:
```suggestion
Base.:(==)(a::ScaleTransform, b::ScaleTransform) = a.s == b.s
```

src/transform/with_lengthscale.jl
```diff
@@ -0,0 +1,33 @@
+"""
+    with_lengthscale(kernel::Kernel, lengthscale::Real)
+    with_lengthscale(kernel::Kernel, lengthscales::AbstractVector{<:Real})
```
devmotion, line 3: It might be a bit easier, both to read and write, if one adds separate docstrings to the two methods. They will both show up if one looks up `with_lengthscale`.

src/transform/with_lengthscale.jl
```diff
@@ -0,0 +1,33 @@
...
+
+Constructs a transformed kernel with `lengthscale`.
+If a vector `lengthscales` is passed instead, constructs an "ARD" kernel with different lengthscales for each dimension.
```
devmotion, line 6: The convention is to use the imperative form in the first line of the docstring:
```suggestion
Construct a transformed kernel with `lengthscale`.
```

src/transform/with_lengthscale.jl
```diff
@@ -0,0 +1,33 @@
...
+julia> SqExponentialKernel() ∘ ARDTransform(inv.(ℓ)) == with_lengthscale(SqExponentialKernel(), ℓ)
+true
+```
+
+"""
+function with_lengthscale(kernel::Kernel, lengthscale::Real)
+    return compose(kernel, ScaleTransform(inv(lengthscale)))
+end
+function with_lengthscale(kernel::Kernel, lengthscales::AbstractVector{<:Real})
+    return compose(kernel, ARDTransform(inv.(lengthscales)))
```
devmotion, line 32: The broadcasting machinery is not needed it seems:
```suggestion
    return kernel ∘ ARDTransform(map(inv, lengthscales))
```

st--: why would you prefer `map` instead of the broadcasting?

devmotion: The broadcasting machinery is very heavy and puts a lot of additional stress on the compiler, so in general I avoid it if there is no actual broadcasting of dimensions needed.

src/kernels/transformedkernel.jl
```diff
@@ -72,6 +72,10 @@ Base.:∘(k::TransformedKernel, t::Transform) = TransformedKernel(k.kernel, k.tr
 Base.:∘(k::Kernel, ::IdentityTransform) = k
 Base.:∘(k::TransformedKernel, ::IdentityTransform) = k
 
+function Base.isequal(k::TransformedKernel, k2::TransformedKernel)
+    return isequal(k.kernel, k2.kernel) && isequal(k.transform, k2.transform)
+end
```
devmotion, line 77: If one defines `isequal` one has to define `hash` as well since otherwise the two are not consistent. In general, if one does not care about floating point comparisons one should just define `==` instead of `isequal` since `isequal` falls back to `==`. Here it probably would make sense to define both since it would lead to different comparisons of the kernel parameters and parameters of the transform.

Basically, most often one wants the recursive implementation in https://github.com/andrewcooke/AutoHashEquals.jl/issues/18.

st--: I don't follow why one cannot define `isequal` without defining `hash`; it can be easy to check containers for equality without being able to construct a hash function. Based on the docstrings of `isequal` and `==`, it seems like the former is the one to use for containers. But in any case that seems out of scope of this PR.

devmotion: > I don't follow why one cannot define isequal without defining hash; it can be easy to check containers for equality without being able to construct a hash function.

It would lead to an inconsistency. It is explained in the docstring of `isequal` (and a bit shorter in the one for `==`):

```julia
  isequal is the comparison function used by hash tables (Dict). isequal(x,y) must imply that hash(x) == hash(y).

  This typically means that types for which a custom == or isequal method exists must implement a corresponding hash method (and vice versa). Collections typically implement
  isequal by calling isequal recursively on all contents.
```

> Based on the docstrings of isequal and ==, it seems like the former is the one to use for containers.

This is not correct, e.g., both are defined for arrays: https://github.com/JuliaLang/julia/blob/d85d8a91b79021668a855ab496ff3d79c2de5f4b/base/abstractarray.jl#L1842 and https://github.com/JuliaLang/julia/blob/d85d8a91b79021668a855ab496ff3d79c2de5f4b/base/abstractarray.jl#L1818 The difference is the floating point comparisons, and hence for anything involving floating point numbers it makes sense to define both.

> But in any case that seems out of scope of this PR.

Just defining `isequal` leads to incorrect behaviour for dictionaries, so at least one should add `hash` as well in this PR. And then I guess one can just define `==` as well?

st--: > Just defining isequal leads to incorrect behaviour for dictionaries,

could you give an example of what you mean ?

> so at least one should add hash as well in this PR. And then I guess one can just define == as well?

There were already plenty of `isequal` implementations in the codebase; it was just the one for `TransformedKernel` that was missing to be able to write the doctests. I think "add `hash` for all of them" is beyond the scope of this PR. If you don't want this PR merged without adding all of those, please feel free to take over and do the work.

devmotion: > could you give an example of what you mean ?

Dictionaries use the `hash` function to search for keys but not `isequal`. Thus if the `hash` function is not consistent with `isequal`, objects that are equal according to `isequal` can be considered as different by the `hash` function, and hence the same key could appear multiple times. And vice versa, objects might seem to be equal according to `hash` and their values are replaced even though actually they are different according to `isequal`.

> There were already plenty of isequal implementations in the codebase; it was just the one for TransformedKernel that was missing to be able to write the doctests. I think "add hash for all of them" is beyond the scope of this PR. If you don't want this PR merged without adding all of those, please feel free to take over and do the work.

Existing implementations should definitely not be fixed in this PR. However, I think we should not add a broken/inconsistent implementation of equality checks in this PR. In my opinion, either we should not add any equality checks at all and use different tests (e.g., just check the types or the lengthscale explicitly) or we should define the equality checks correctly (i.e., add `isequal`, `hash`, and `==`).

st--: please take over the PR then, this was how far I was prepared to take it:)

devmotion: I opened https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/336, I guess it is cleaner to just not define `isequal` in this PR.

devmotion:
Fixed by https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/336.

PR #336, opened 2021-07-09
Add `with_lengthscale` (alternative/extension of #335)

devmotion:
This PR is an alternative/extension of #335 that avoids the issues in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/335#discussion_r666911524. It seems cleaner to fix and complete equality definitions in a separate PR. In addition to the doctests the PR adds some regular tests.


willtebbutt (approved):
LGTM

Issue #339, opened 2021-07-19
Use dev in examples?


willtebbutt:
Out examples each have a different environment, and each has it's own dependency on AbstractGPs. Currently, this dependency is not tied to the currently checked-out branch via `dev ../..` or whatever. I wonder whether we should do this?

Reasons in favour:
1. allows us to run the examples in CI to ensure that they still pass when we make changes.
1. it ensures that our examples run on whatever the most recent version of 
1. if we _do_ run our examples as part of CI, it'll give us a few examples of stuff running in the wild, which is extra coverage, which is probably a good thing

I also don't recall whether we've already had this conversation and ruled this out...

devmotion:
The examples (or at least some of them) use the checked out development version: https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/fbce558b9362fa7e6d4c810ec7a98de09f3990b8/examples/kernel-ridge-regression/Manifest.toml#L284

willtebbutt:
Oh -- I extrapolated from a single data point! https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/fbce558b9362fa7e6d4c810ec7a98de09f3990b8/examples/gaussian-process-priors/Manifest.toml#L282

I'll make a PR to correct.

PR #340, opened 2021-07-19
Fix Manifest for GP prior example

willtebbutt:
As discussed in #339

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -44,6 +44,7 @@ From a Julia REPL started in your example script's directory, you can run
 ```julia
 julia> ] activate .
 julia> ] add Literate
+julia> # the following line adds a dependency on KernelFunctions that is based on the local directories, not a hash:
 julia> ] dev ../..
 julia> # add any other example-specific dependencies
 ```
--- a/examples/README.md
+++ b/examples/README.md
@@ -44,7 +44,7 @@ From a Julia REPL started in your example script's directory, you can run
 ```julia
 julia> ] activate .
 julia> ] add Literate
-julia> ] add KernelFunctions
+julia> ] dev ../..
 julia> # add any other example-specific dependencies
 ```
 to generate the project files.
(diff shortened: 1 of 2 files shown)


willtebbutt:
Test failure is entirely unrelated to this PR. Not sure what's causing it, but it's a separate issue.

theogf (approved):

willtebbutt:
Hmm I don't appear to be able to merge. Possibly because the FBM kernel AD tests are being annoying?

examples/gaussian-process-priors/Manifest.toml
```diff
@@ -281,9 +281,9 @@ version = "2.1.0+0"
 
 [[KernelFunctions]]
 deps = ["ChainRulesCore", "Compat", "CompositionsBase", "Distances", "FillArrays", "Functors", "LinearAlgebra", "Random", "Requires", "SpecialFunctions", "StatsBase", "StatsFuns", "TensorCore", "Test", "ZygoteRules"]
-git-tree-sha1 = "e8b5ba31b6d18695fd46bfcd8557682839023195"
+path = "../.."
 uuid = "ec8451be-7e33-11e9-00cf-bbf324bd1392"
-version = "0.10.6"
+version = "0.10.8"
```
st--, line 286: will we have to keep bumping this version and keep it in sync with actual package version ?

st-- (approved):

st--:
@willtebbutt we should also update the examples/README.md instructions for adding new examples, do you want to just add that in here ?

examples/gaussian-process-priors/Manifest.toml
willtebbutt: Hmmm it seems to be fine running it locally -- I think we might just want to add a preamble to each of the example scripts along the lines of
```julia
using Pkg
Pkg.activate(@__DIR__)
Pkg.resolve()
Pkg.instantiate()
```
which should sort it out.

willtebbutt: I'm going to leave this for now actually -- it seems to run fine locally when the KernelFunctions version is out of sync, provided that it's dev-ed correctly.

examples/README.md
```diff
@@ -44,7 +44,7 @@ From a Julia REPL started in your example script's directory, you can run
 ```julia
 julia> ] activate .
 julia> ] add Literate
-julia> ] add KernelFunctions
+julia> ] dev ../..
```
st--, line 47:
```suggestion
julia> # the following line adds a dependency on KernelFunctions that is based on the local directories, not a hash:
julia> ] dev ../..
```

st-- (approved):

PR #342, opened 2021-07-20
Don't test FBM on Zygote

willtebbutt:
We've been CI failures. I believe I've diagnosed the problem, and have opened an issue about it: https://github.com/FluxML/Zygote.jl/issues/1036

In the mean time, we need our CI to pass, so I've added some tests which are broken, and will be fixed when the issue in question is fixed (so we'll know about it when it happens).

### DIFF
--- a/test/basekernels/fbm.jl
+++ b/test/basekernels/fbm.jl
@@ -13,7 +13,13 @@
 
     test_interface(k)
     @test repr(k) == "Fractional Brownian Motion Kernel (h = $(h))"
-    test_ADs(FBMKernel; ADs=[:ReverseDiff, :Zygote])
-    @test_broken "Tests failing for kernelmatrix(k, x) for ForwardDiff"
+    test_ADs(FBMKernel; ADs=[:ReverseDiff])
+
+    # Tests failing for ForwardDiff
+    @test_broken !isinf(ForwardDiff.gradient(x -> x[1]^x[2], [0.0, 0.9])[1])
+
+    # Zygote also doesn't work.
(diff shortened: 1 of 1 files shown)


test/basekernels/fbm.jl
```diff
@@ -13,7 +13,14 @@
...
     test_interface(k)
     @test repr(k) == "Fractional Brownian Motion Kernel (h = $(h))"
-    test_ADs(FBMKernel; ADs=[:ReverseDiff, :Zygote])
-    @test_broken "Tests failing for kernelmatrix(k, x) for ForwardDiff"
+    test_ADs(FBMKernel; ADs=[:ReverseDiff])
+
+    # Tests failing for ForwardDiff and Zygote.
+    # Related to: https://github.com/FluxML/Zygote.jl/issues/1036
+    @test_broken !isinf(ForwardDiff.gradient(x -> x[1]^x[2], [0.0, 0.9])[1])
+    @test_broken !isinf(
```
st--, line 21: Looks like you might be missing a line:
```suggestion
    f(x, y) = x ^ y
    @test_broken !isinf(
```

willtebbutt: Indeed I am -- should have been more careful

st-- (dismissed):

willtebbutt:
Eurgh -- our CI on 1.3 gets a different version of Zygote from 1.6. I'm not entirely sure why though...

willtebbutt:
@st-- I had to make some changes. If you could re-review, that would be grand.

st-- (approved):

PR #345, opened 2021-07-27
Multi-Output helper function

willtebbutt:
Implements a function to make it possible to work with our input types for multi-output GPs more straightforwardly in the isotopic case.

It's entirely possible that we'll want more helper functions as we move forwards, so this should be seen as a first step in that direction, rather than something that will solve all useability problems.

edit: I've added a note to the docs. Unless I'm missing something obvious, we don't currently have a multi-output example, so I'm not going to create one in this PR to showcase this functionality.

edit2: also note that I'm testing only via doctests. It feels a bit weird, but it's not obviously to me that it's obviously a bad idea, so maybe it's okay?

### DIFF
--- a/src/mokernels/moinput.jl
+++ b/src/mokernels/moinput.jl
@@ -103,3 +103,72 @@ See [Inputs for Multiple Outputs](@ref) in the docs for more info.
 and removed in version 0.12.
 """
 const MOInput = MOInputIsotopicByOutputs
+
+
+"""
+    prepare_isotopic_multi_output_data(x::AbstractVector, Y::AbstractMatrix{<:Real})
+
+Utility functionality to convert a collection of inputs `x` and a matrix of vector-valued
+outputs `Y` into a canonical format needed by `KernelFunctions`.
+
+If `length(x)` is `N`, then `Y` must be either of size `N x P` or `P x N`, where `P` is the
+number of outputs.
+
+If `size(Y)` is `N x P`, then
+```jldoctest
+julia> N, P = 10, 5;
+
+julia> x = ColVecs(randn(3, N));
+
+julia> Y = randn(N, P);
+
+julia> x_canon, y_canon = prepare_isotopic_multi_output_data(x, Y);
+
+julia> x_canon isa KernelFunctions.MOInputIsotopicByOutputs
+true
+
+julia> length(x_canon) == N * P
+true
+
+julia> y_canon isa AbstractVector{<:Real}
+true
+
+julia> length(y_canon) == length(x_canon)
+true
+```
+
+If `size(Y)` is `P x N`, then
+```julia
+julia> N, P = 10, 5;
+
+julia> x = ColVecs(randn(3, N));
+
+julia> Y = randn(P, N);
+
+julia> x_canon, y_canon = prepare_isotopic_multi_output_data(x, Y);
+
+julia> x_canon isa KernelFunctions.MOInputIsotopicByFeatures
+true
+
+julia> length(x_canon) == N * P
+true
+
+julia> y_canon isa AbstractVector{<:Real}
+true
+
+julia> length(y_canon) == length(x_canon)
+true
+```
+
+An `ArgumentError` is thrown if the sizes don't match.
+"""
+function prepare_isotopic_multi_output_data(x::AbstractVector, Y::AbstractMatrix{<:Real})
+    N = length(x)
+    if size(Y, 1) == N
+        return MOInputIsotopicByOutputs(x, size(Y, 2)), vec(Y)
+    elseif size(Y, 2) == N
+        return MOInputIsotopicByFeatures(x, size(Y, 1)), vec(Y)
+    else
+        throw(ArgumentError("Length of x not compatible with size of Y."))
+    end
+end
--- a/src/mokernels/moinput.jl
+++ b/src/mokernels/moinput.jl
@@ -105,46 +105,24 @@ and removed in version 0.12.
 const MOInput = MOInputIsotopicByOutputs
 
 """
-    prepare_isotopic_multi_output_data(x::AbstractVector, Y::AbstractMatrix{<:Real})
+    prepare_isotopic_multi_output_data(x::AbstractVector, y::ColVecs)
 
-Utility functionality to convert a collection of inputs `x` and a matrix of vector-valued
-outputs `Y` into a canonical format needed by `KernelFunctions`.
+Utility functionality to convert a collection of `N = length(x)` inputs `x`, and a
+vector-of-vectors `y` (efficiently represented by a `ColVecs`) into a format suitable for
+use with multi-output kernels.
 
-If `length(x)` is `N`, then `Y` must be either of size `N x P` or `P x N`, where `P` is the
-number of outputs.
+`y[n]` is the vector-valued output corresponding to the input `x[n]`.
+Consequently, it is necessary that `length(x) == length(y)`.
 
-If `size(Y)` is `N x P`, then
-```jldoctest
-julia> N, P = 10, 5;
-
-julia> x = ColVecs(randn(3, N));
-
-julia> Y = randn(N, P);
-
-julia> x_canon, y_canon = prepare_isotopic_multi_output_data(x, Y);
-
-julia> x_canon isa KernelFunctions.MOInputIsotopicByOutputs
-true
-
-julia> length(x_canon) == N * P
-true
-
-julia> y_canon isa AbstractVector{<:Real}
-true
-
-julia> length(y_canon) == length(x_canon)
-true
-```
-
-If `size(Y)` is `P x N`, then
+For example:
 ```julia
 julia> N, P = 10, 5;
 
-julia> x = ColVecs(randn(3, N));
+julia> x = randn(N);
 
-julia> Y = randn(P, N);
+julia> y = ColVecs(randn(P, N));
 
-julia> x_canon, y_canon = prepare_isotopic_multi_output_data(x, Y);
+julia> x_canon, y_canon = prepare_isotopic_multi_output_data(x, y);
 
 julia> x_canon isa KernelFunctions.MOInputIsotopicByFeatures
 true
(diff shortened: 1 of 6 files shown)


theogf:
Did not go into details but the obvious edge case not considered here is if N == P

src/mokernels/moinput.jl
```diff
@@ -103,3 +103,71 @@ See [Inputs for Multiple Outputs](@ref) in the docs for more info.
...
+function prepare_isotopic_multi_output_data(x::AbstractVector, Y::AbstractMatrix{<:Real})
+    N = length(x)
+    if size(Y, 1) == N
+        return MOInputIsotopicByOutputs(x, size(Y, 2)), vec(Y)
+    elseif size(Y, 2) == N
+        return MOInputIsotopicByFeatures(x, size(Y, 1)), vec(Y)
+    else
+        throw(ArgumentError("Length of x not compatible with size of Y."))
+    end
+end
```
st--, line 173: Seconded @theogf's comment, going by the shape can fail (silently) for N==P. I was wondering if we could instead re-use `ColVecs`/`RowVecs`, but I now think that's an orthogonal property - one is whether rows or columns of Y match x (and, except for N==P, you can identify that from the size); the _other_ is whether you want to vectorize it by outputs or by features.

src/mokernels/moinput.jl
```diff
@@ -103,3 +103,71 @@ See [Inputs for Multiple Outputs](@ref) in the docs for more info.
...
+julia> y_canon isa AbstractVector{<:Real}
+true
+
+julia> length(y_canon) == length(x_canon)
+true
+```
+
+An `ArgumentError` is thrown if the sizes don't match.
+"""
+function prepare_isotopic_multi_output_data(x::AbstractVector, Y::AbstractMatrix{<:Real})
```
st--, line 164:
```suggestion
function prepare_isotopic_multi_output_data(x::AbstractVector, Y::AbstractMatrix)
```
or what's the reason for constraining to <:Real ?

willtebbutt: Hmm good point. Probaby unnecessary at this point.

src/mokernels/moinput.jl
willtebbutt: The `ColVecs` vs `RowVecs` idea makes sense to me actually -- you could just require that `length(y) == length(x)`, where `y isa Union{ColVecs, RowVecs}`, then you've disambiguated successfully. Do you not think this makes sense?

As regards whether to vectorize by outputs or by features, I had been thinking you do whatever `vec` does. This has the advantage that it won't need to allocate new memory (it just reinterprets the existing data) and that it's definitely going to do something that creates a valid collection of inputs, but the disadvantage of not being completely explicit. Do you think a different appoach might be better? The only way I could think to make it more explicit would be to add an additional argument to the helper that lets you be more specific -- at that point though, I _suspect_ that you're no better off than just doing it manually.

st--: ah yeah, I think that would be a good solution then - you can anyways check what you got by looking at the return type:)

src/mokernels/moinput.jl
```diff
@@ -103,3 +103,71 @@ See [Inputs for Multiple Outputs](@ref) in the docs for more info.
...
+function prepare_isotopic_multi_output_data(x::AbstractVector, Y::AbstractMatrix)
+    N = length(x)
+    if size(Y, 1) == N
+        return MOInputIsotopicByOutputs(x, size(Y, 2)), vec(Y)
+    elseif size(Y, 2) == N
+        return MOInputIsotopicByFeatures(x, size(Y, 1)), vec(Y)
+    else
+        throw(ArgumentError("Length of x not compatible with size of Y."))
+    end
+end
```
st--, line 173:
```suggestion
function prepare_isotopic_multi_output_data(x::AbstractVector, Y::Union{RowVecs, ColVecs})
    if length(x) != length(Y)
        throw(ArgumentError("Length of x not compatible with length of Y."))
    end
    if Y isa RowVecs
        num_outputs = size(Y.X, 2)
        return MOInputIsotopicByOutputs(x, num_outputs), vec(Y.X)
    elseif Y isa ColVecs
        num_outputs = size(Y.X, 1)
        return MOInputIsotopicByFeatures(x, num_outputs), vec(Y.X)
    end
end
```

theogf: Should we also allow `Y::AbstractVector` though?

willtebbutt: Didn't see this until after I pushed a change -- I've gone with two methods rather than one, and docstringed them separately.

st--: @theogf 

> Should we also allow Y::AbstractVector though?

what would that look like? AbstractVector{<:AbstractVector}? how would you then encode which one you want ? I would assume in the isotopic case you'll generally have your Y observations as some kind of matrix...

st-- (commented):
Generally looks good to me, but I think it'd be super helpful to include a concrete example with explicit numbers in the docstrings, as even being somewhat familiar with all this it keeps doing my head in. Do doctests count towards code coverage? It might make more sense to move the "consistency checks" to actual tests.

src/mokernels/moinput.jl
```diff
@@ -103,3 +103,79 @@ See [Inputs for Multiple Outputs](@ref) in the docs for more info.
...
+true
+
+julia> length(x_canon) == N * P
+true
+
+julia> y_canon isa AbstractVector{<:Real}
+true
+
+julia> length(y_canon) == length(x_canon)
+true
```
st--, line 175:
```suggestion
julia> x = [1, 2, 3];

julia> Y = [11 12; 21 22; 31 32]
3×2 Matrix{Int64}:
 11  12
 21  22
 31  32

julia> inputs, outputs = prepare_isotopic_multi_output_data(x, RowVecs(Y));

julia> inputs
6-element KernelFunctions.MOInputIsotopicByOutputs{Int64, Vector{Int64}}:
 (1, 1)
 (2, 1)
 (3, 1)
 (1, 2)
 (2, 2)
 (3, 2)

julia> outputs
6-element Vector{Int64}:
 11
 21
 31
 12
 22
 32
```

st--: and move the current doctest contents into actual tests

src/mokernels/moinput.jl
```diff
@@ -103,3 +103,79 @@ See [Inputs for Multiple Outputs](@ref) in the docs for more info.
...
```
st--, line 137: as for RowVecs case (with N=3, P=2), adapted to Y'

st--: and original doctests moved into actual tests

src/mokernels/moinput.jl
willtebbutt: Oooo very nice -- this is much better.

willtebbutt:
Eurgh -- looks like 1.3 doesn't like the doctests. I feel like we must have encountered this before, any thoughts @theogf @st-- ?

src/mokernels/moinput.jl
```diff
@@ -103,3 +103,94 @@ See [Inputs for Multiple Outputs](@ref) in the docs for more info.
...
+    prepare_isotopic_multi_output_data(x::AbstractVector, y::ColVecs)
+
+Utility functionality to convert a collection of `N = length(x)` inputs `x`, and a
+vector-of-vectors `y` (efficiently represented by a `ColVecs`) into a format suitable for
+use with multi-output kernels.
+
+`y[n]` is the vector-valued output corresponding to the input `x[n]`.
+Consequently, it is necessary that `length(x) == length(y)`.
+
+For example, if outputs are initially stored in a `num_outputs x N` matrix:
```
st--, line 117:
```suggestion
For example, if outputs are initially stored in a `num_outputs × N` matrix:
```

src/mokernels/moinput.jl
```diff
@@ -103,3 +103,94 @@ See [Inputs for Multiple Outputs](@ref) in the docs for more info.
...
+    prepare_isotopic_multi_output_data(x::AbstractVector, y::RowVecs)
+
+Utility functionality to convert a collection of `N = length(x)` inputs `x` and output
+vectors `y` (efficiently represented by a `RowVecs`) into a format suitable for
+use with multi-output kernels.
+
+`y[n]` is the vector-valued output corresponding to the input `x[n]`.
+Consequently, it is necessary that `length(x) == length(y)`.
+
+For example, if outputs are initial stored in an `N x num_outputs` matrix:
```
st--, line 162:
```suggestion
For example, if outputs are initial stored in an `N × num_outputs` matrix:
```

test/mokernels/moinput.jl
```diff
@@ -44,4 +44,35 @@
         @test ibf[7] == (x[3], 1)
         @test all([(x_, i) for x_ in x for i in 1:3] .== ibf)
     end
+
+    @testset "prepare_isotopic_multi_output_data" begin
+        @testset "ColVecs" begin
+            N = 10
+            P = 5
```
st--, line 51: It doesn't matter much here, but in general I like to use prime numbers for array dimensions because that makes it a lot more obvious what's going on when something fails.
```suggestion
            N = 5
            P = 3
```

st-- (dismissed):
Looks good to me; no idea why the doctest fails - it should be caught by the filter! @theogf was going to have a look into that

theogf:
Just trying something now

theogf:
It works locally at least!

test/mokernels/moinput.jl
```diff
@@ -44,4 +44,35 @@
...
+            x_canon, y_canon = prepare_isotopic_multi_output_data(x, y)
+
+            @test x_canon isa KernelFunctions.MOInputIsotopicByFeatures
+            @test length(x_canon) == N * P
+            @test y_canon isa AbstractVector{<:Real}
+            @test length(y_canon) == length(x_canon)
+        end
+        @testset "RowVecs" begin
+            N = 10
+            P = 5
```
st--, line 65:
```suggestion
            N = 5
            P = 3
```

st-- (approved):

willtebbutt:
@theogf your regex magic seems to have sorted it. Have bumped patch -- will merge once CI passes.

PR #349, opened 2021-08-02
Adding vcat and tests

### LINKED ISSUE #348: Add `Base.vcat` method to MO input types

thomasgudjonwright:
Closes #348.

Not sure if we want to default to a specific type of the two inputs are different, or error as I have done at this point.


willtebbutt:
> Not sure if we want to default to a specific type of the two inputs are different, or error as I have done at this point.

I suspect that we just want two methods here -- one for each isotopic type.  The reason being that `vcat` _is_ defined for both of them via a fallback definition -- we just want to accelerate it in cases where it's possible to accelerate.

test/mokernels/moinput.jl
```diff
@@ -3,46 +3,57 @@
...
-        @test lastindex(ibo) == 12
-        @test firstindex(ibo) == 1
-        @test_throws BoundsError ibo[0]
-
-        @test ibo[2] == (x[2], 1)
-        @test ibo[5] == (x[1], 2)
-        @test ibo[7] == (x[3], 2)
-        @test all([(x_, i) for i in 1:3 for x_ in x] .== ibo)
-    end
+    @testset "isotopicbyoutputs_and_isotopicbyfeatures" begin
```
willtebbutt, line 6: I'm a bit confused by the changes here. Could you explain a bit why it's helpful to join the tests for the two isotopic types into the same testset?

thomasgudjonwright:
> > Not sure if we want to default to a specific type of the two inputs are different, or error as I have done at this point.
> 
> I suspect that we just want two methods here -- one for each isotopic type. The reason being that `vcat` _is_ defined for both of them via a fallback definition -- we just want to accelerate it in cases where it's possible to accelerate.

Ahh okay and then just let it error by itself if someone tries to pass it one of each type as the inputs?

test/mokernels/moinput.jl
thomasgudjonwright: I figured it would save redefining the same values (`ibo` and `ibf`). Also since itincluded one of each type I thought it didn't belong in either of the individual testsets.

If we separate the two cases however and remove the type check, letting it error on its own, we can remove the nested testsets

willtebbutt:
>Ahh okay and then just let it error by itself if someone tries to pass it one of each type as the inputs?

Not quite -- it'll hit the slow fallback method, and produce a valid result, it will just do it slowly.

willtebbutt (approved):
LGTM. Just needs a patch bump.

PR #352, opened 2021-08-06
remove StatsFuns dependency

st--:
Instead, depend explicitly on IrrationalConstants for logtwo and two\pi (for the newcomer: what's the value of doing that over simply writing log(2) or 2\pi, anyways?), and on LogExpFunctions for softplus - StatsFuns was simply reexporting all of these. We are still held back from LogExpFunctions=0.3 compatibility by SpecialFunctions (@torfjelde is working on that already), but otherwise good to go. Will bump version once we can bump the dependencies.


willtebbutt (approved):
This looks reasonable to me -- very happy to see this merged once SpecialFunctions is updated.

st--:
Waiting for https://github.com/JuliaMath/SpecialFunctions.jl/pull/333 (or @torfjelde's replacement PR).

PR #353, opened 2021-08-08
Heterotopic MO kernel helper

willtebbutt:
I think I like this feature. It's a bit trivial at the minute, but I _think_ it probably helps a bit with the cognitive load of using multi-output kernels in a heterotopic setting.

If nothing else, suggesting that users use this function to get their multi-output data in a format that KernelFunctions likes will help to future proof us a bit in case we change implementation details later on (e.g. maybe using `StructArray`s to represent the inputs that are returned from this function)


src/mokernels/moinput.jl
```diff
@@ -189,8 +191,65 @@ julia> outputs
...
+    # Ensure validity of arguments.
+    if length(x) != length(y)
+        throw(ArgumentError("length(x) != length(y)"))
+    end
+    if length(x) != length(output_indices)
+        throw(ArgumentError("length(x) != length(output_indices)"))
+    end
+
+    # Construct inputs and outputs for multi-output kernel.
+    x_mogp = map(tuple, x, output_indices)
```
thomasgudjonwright, line 253: It can always be added later, but do we want to make a `MOInputHeterotopic` type?

src/mokernels/moinput.jl
```diff
@@ -143,6 +143,8 @@ julia> outputs
  3.1
  3.2
 ```
+
+See also [`prepare_heterotopic_multi_output_data`](@ref).
 """
 function prepare_isotopic_multi_output_data(x::AbstractVector, y::ColVecs)
     length(x) == length(y) || throw(ArgumentError("length(x) not equal to length(y)."))
```
thomasgudjonwright, line 150: This is more related to the last MR, but should we be providing any explanation as to why `Y` being a `ColVecs` or `RowVecs` affects the output? It looks like it is just being used as an indicator here really, but some people seeing this might think theres is some intrinsic connection between `ColVecs`  and `MOInputIsotopicByFeatures` (unless there is one that I am missing!)

thomasgudjonwright (commented):
Great idea :) Just a couple small comments that are really more a matter of opinion! Feel free to leave them as is!

src/mokernels/moinput.jl
st--: You're right, it's mainly used as a way of disambiguating, and because of which way `vec(y)` ends up ordering observations.

src/mokernels/moinput.jl
st--: As an alias for `Tuple{T, Int}`?

src/mokernels/moinput.jl
willtebbutt: I've opened an issue about this: https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/356

@thomasgudjonwright please resolve this conversation if you're happy with that resolution for now.

thomasgudjonwright: Sorry for missing this until now! Thanks for opening the issue :)

src/mokernels/moinput.jl
thomasgudjonwright: Probably not. More likely a subtype similarly to how `MOInputIsotopicByFeatures` and `MOInputIsotopicByOutputs` are set up. I think this would be better since we could then define some convenient helper functionality (eg. `get_output_input(X::MOInputHeterotopic, p::Int)` could return the observations for output `p`).

This is slightly tangent to what this MR accomplishes though so lets resolve this thread for now so this can go through :) My unresolved threads have held this up for far too long!

willtebbutt: Please do open an issue about this Tom, would be good to figure out whether this is a thing that we want to do.

docs/src/api.md
```diff
@@ -58,13 +58,19 @@ For an explanation of this design choice, see [the design notes on multi-output
...
+values throughout their code, without having to worry further about correctly formatting
+their data for KernelFunctions' sake:
 ```@docs
 prepare_isotopic_multi_output_data(x::AbstractVector, y::ColVecs)
 prepare_isotopic_multi_output_data(x::AbstractVector, y::RowVecs)
+prepare_heterotopic_multi_output_data
 ```
 
-The input types that it constructs can also be constructed manually:
+The input types returned by `prepare_isotopic_multi_output_data` can also be constructed manually:
```
thomasgudjonwright, line 73: Only remaining thought is that this might be misleading as `MOInput` is an alias for `MOInputIsotopicByOutputs`, so technically `prepare_isotopic_multi_output_data` could return something different from what will be shown in the docs here. Maybe I am being too persnickety! Otherwise LGTM, approved :)

willtebbutt:
Patch bumped. Will merge when CI passes again.

PR #354, opened 2021-08-09
Some improvement and additions to MO kernels

Crown421:
I have been using KernelFunctions in my research, but have been building some more or less hacky solution on top of it to facilitate Multi-Output GPs. 

Here I would like to move some of that work into this package to streamline my own code and get it ready for publication. 

The following is a start, with more things to come soon. I have added tests as I thought appropriate, but ran into a segfault caused by the AD test function. I hope that this was a local problem.


st--:
Thanks for your interest in KernelFunctions:) could you provide a bit more context on what hadn't been working for you / what you're trying to achieve / work around with your changes?

src/mokernels/independent.jl
```diff
@@ -23,20 +23,39 @@ struct IndependentMOKernel{Tkernel<:Kernel} <: MOKernel
     kernel::Tkernel
 end
 
+# kernel function should be symmetric
+# would really like (κ::IndependentMOKernel)((x, px)::Tuple{T,Int}, (y, py)::Tuple{T,Int}) where T, but seems to cause autodiff problems
```
st--, line 27: what are the problems you get when you force first and second argument to have the same type?

src/mokernels/independent.jl
```diff
@@ -23,20 +23,39 @@ struct IndependentMOKernel{Tkernel<:Kernel} <: MOKernel
...
 function (κ::IndependentMOKernel)((x, px)::Tuple{Any,Int}, (y, py)::Tuple{Any,Int})
     if px == py
         return κ.kernel(x, y)
     else
-        return 0.0
+        retType = Base.return_types(κ.kernel, (typeof(x), typeof(y)))[1]
```
st--, line 32:
```suggestion
        retType, = Base.return_types(κ.kernel, (typeof(x), typeof(y)))
```
encodes in code that we expect the kernel to return a single value, makes it easier to read & also errors instead of just a silent bug in case kernel() happens to return more than one value..

src/mokernels/independent.jl
Crown421: It seemed like the AD test failed as a result, so I reverted that change. However, I can't be sure, because the test fails with a SegFault even after reverting it.

Crown421:
> Thanks for your interest in KernelFunctions:) could you provide a bit more context on what hadn't been working for you / what you're trying to achieve / work around with your changes?

Sorry, of course. For this set of changes, my changes aim to make sure that the kernels are type stable, as well as propose a specific implementation for the `kernelmatrix` function that is substantially faster than the general method. 

In additional commits, I would like to introduce a new type of MO kernel, the Group-Invariant-Matrix (GIM) kernel (Reisert, 2007).

src/mokernels/independent.jl
Crown421: It appears from the CI that the AD segfault crash I saw locally is a broader issue, that makes me feel both happier and sadder.

src/mokernels/independent.jl
Crown421: I did not know about this syntax, thank you.

src/mokernels/independent.jl
```diff
@@ -23,20 +23,39 @@ struct IndependentMOKernel{Tkernel<:Kernel} <: MOKernel
...
+        return zero(retType)
```
willtebbutt, line 33: Using `Base.return_types` always makes me nervous, because there's no guarantee that it will infer the correct type. If possible, it's generally considered best practice to rely on the promotion system in this kind of situation.
```suggestion
        return false
```
be sufficient for your use case? `false` gets promoted to `0f0` when it encounters a `Float32` and `0.0` when it encounters a `Float64`, so maybe that's the best thing to do here?

The whole thing won't be type-stable of course, but it's only a small type-union, so hopefully it's okay

Crown421: In a away, this becomes far less of an issue if you are ok with my proposal for a specific implementation for the kernelmatrix (see kernelmatrix2 and the top level script).

I just think that if possible, type stability is a nice thing to have.

willtebbutt: Yeah -- I think we'll probably be fine with them. Would you mind posting the timing results so that we don't all have to run them locally?

> I just think that if possible, type stability is a nice thing to have.

Totally agree in general. In this particular case, using `Base.return_types` here means that in order for this function to work at all, `κ.kernel` _must_ infer. If the underlying kernel is type-unstable, and `Base.return_types` returns `Any`, `zero(Any)` will error.

Crown421: Of course, have added the results to the script. 

Ideally, all possible underlying kernels would be type stable and tested as well. But I see how that is a problem to 100% guarantee at this point.

src/mokernels/intrinsiccoregion.jl
```diff
@@ -34,10 +34,37 @@ function IntrinsicCoregionMOKernel(; kernel::Kernel, B::AbstractMatrix)
...
+function IntrinsicCoregionMOKernel(kernel::Kernel, B::AbstractMatrix)
+    return IntrinsicCoregionMOKernel{typeof(kernel),typeof(B)}(kernel, B)
+end
+
+function (k::IntrinsicCoregionMOKernel)((x, px)::Tuple{Any,Int}, (y, py)::Tuple{Any,Int}) 
     return k.B[px, py] * k.kernel(x, y)
 end
 
+# convenience function
+function (k::IntrinsicCoregionMOKernel)(x::Vector{T}, y::Vector{T}) where T <: Real
```
willtebbutt, line 46: I don't think we'll be able to accept this particular method. I get the motivation for it, but the API currently has well-defined semantics for kernel evaluation, and I'd rather keep it that way.

We could consider a new convenience function that provides this functionality though. It would need a name though...

willtebbutt:
Thanks for uploading the timing data -- pretty clear win with the new versions. This kind of optimisation is exactly the kind of PR we're keen to have, so please rename `kernelmatrix2` to `kernelmatrix` when you get a minute so that we can push this PR forward.

src/mokernels/intrinsiccoregion.jl
st--: Something along the lines of "covariance over outputs"... I find it quite helpful to have verbose-but-explicit names in cases like this when we're a bit unsure what to name things. So how about `cov_of_outputs` maybe?

Also am I missing something - it seems `x::Vector{T}` is a bit too restrictive: my understanding is that you want to evaluate the covariance matrix between all outputs at two points, but it could be scalars as well (or a completely different type!) no?

src/mokernels/independent.jl
st--: :) that's what I love about PR reviews, learning how to write better code!
and I learned that this is actually all superseded by what @willtebbutt wrote below ;)

src/mokernels/independent.jl
```diff
@@ -23,20 +23,39 @@ struct IndependentMOKernel{Tkernel<:Kernel} <: MOKernel
...
     temp = k.kernel.(x.x, permutedims(y.x))
     return cat((temp for _ in 1:(y.out_dim))...; dims=(1, 2))
 end
 
+export kernelmatrix2
+function kernelmatrix2(k::IndependentMOKernel, x::MOInputIsotopicByFeatures, y::MOInputIsotopicByFeatures)
+    @assert x.out_dim == y.out_dim
+    Ktmp = kernelmatrix(k.kernel, x.x, y.x)
+    mtype = eltype(Ktmp)
+    kron(Ktmp, Matrix{mtype}(I, x.out_dim, x.out_dim))
```
st--, line 49: What would the timing be like when using lazy Kronecker products as implemented by https://github.com/MichielStock/Kronecker.jl/ ? It'd be great to let downstream uses benefit from structure in the kernel matrix: )

@Crown421 if you've not already looked into this, would you be willing to do that in this PR already ?

src/mokernels/independent.jl
```diff
@@ -23,20 +23,39 @@ struct IndependentMOKernel{Tkernel<:Kernel} <: MOKernel
...
     if px == py
         return κ.kernel(x, y)
     else
-        return 0.0
+        retType = Base.return_types(κ.kernel, (typeof(x), typeof(y)))[1]
+        return zero(retType)
     end
 end
 
+# this function never gets called it seems
```
st--, line 37: @thomasgudjonwright @willtebbutt can you confirm this? If it's not used, better to delete it altogether perhaps?

thomasgudjonwright: AFAICT this won't get used if the `kernelmatrix2` dispatches get renamed to `kernelmatrix` as [suggested ](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/354#issuecomment-895942796)by @willtebbutt  above!

willtebbutt: Yeah -- I would be surprised if it doesn't get called in the current release, but since `MOInput` is an alias for `MOInputIsotopicByOutputs` now, I agree that it's redundant in light of the proposed changes, and should be removed.

temporary_script.jl
```diff
@@ -0,0 +1,79 @@
...
+
+xMOF = KernelFunctions.MOInputIsotopicByFeatures(x, dims.out)
+xMOO  = KernelFunctions.MOInputIsotopicByOutputs(x, dims.out)
+
+indk = IndependentMOKernel(GaussianKernel())
+
+Kind1 = kernelmatrix(indk, xMOF, xMOF)
+Kind2 = kernelmatrix2(indk, xMOF, xMOF)
+
+Kind1 ≈ Kind2
```
st--, line 18: How about turning this into an actual test using `invoke` (after renaming `kernelmatrix2` back to `kernelmatrix`)?
e.g.
```suggestion
K = kernelmatrix(indk, xMOF, xMOF)
K_fallback = invoke(kernelmatrix, Tuple{Kernel, AbstractVector, AbstractVector})(indk, xMOF, xMOF)

@test K ≈ K_fallback
```

test/mokernels/intrinsiccoregion.jl
```diff
@@ -7,16 +7,27 @@
...
     @test icoregionkernel.B == B
     @test icoregionkernel.kernel == kernel
     @test icoregionkernel(X[1], X[1]) ≈ B[X[1][2], X[1][2]] * kernel(X[1][1], X[1][1])
     @test icoregionkernel(X[1], X[end]) ≈ B[X[1][2], X[end][2]] * kernel(X[1][1], X[end][1])
 
+    # test convenience function using kronecker product
+    @test icoregionkernel(X.x[1], X.x[2]) ≈ icoregionkernel.kernel(X.x[1], X.x[2])*B
+
+    # kernelmatrix
+    @test kernelmatrix(icoregionkernel, X) ≈ kron(kernelmatrix(kernel, X.x), B)
```
st--, line 29: I would replace this with a test against the generic (slow) fallback as mentioned in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/354/files#r685965795

st--:
How about similar definitions for `kernelmatrix!` using `kron!`?

This is probably a source of inconsistency that is likely to pop up again in the future: would it perhaps be better to generally define `kernelmatrix` in terms of `kernelmatrix!`, then people only have to implement `kernelmatrix!` - unless of course you want to return lazy versions (e.g. Kronecker.jl) in the not-in-place method?

temporary_script.jl
st--: I take it back, this should already be covered by
```julia
    @test kernelmatrix(k, x0, x2) ≈ [k(xl, xr) for xl in x0, xr in x2]
```
in src/test_utils.jl

test/mokernels/intrinsiccoregion.jl
st--: or remove it as it should be superseded by the explicit elementwise test in `test_interface`

willtebbutt:
That would be a nice way to do it, but unfortunately `kernelmatrix!` tends not to play nicely with reverse-mode AD due to a lack of support for in-place operations. Consequently, we need the `kernelmatrix` implementations to be free of in-place operations.

src/mokernels/intrinsiccoregion.jl
```diff
@@ -34,10 +34,37 @@ function IntrinsicCoregionMOKernel(; kernel::Kernel, B::AbstractMatrix)
     return IntrinsicCoregionMOKernel{typeof(kernel),typeof(B)}(kernel, B)
 end
 
-function (k::IntrinsicCoregionMOKernel)((x, px)::Tuple{Any,Int}, (y, py)::Tuple{Any,Int})
+function IntrinsicCoregionMOKernel(kernel::Kernel, B::AbstractMatrix)
+    return IntrinsicCoregionMOKernel{typeof(kernel),typeof(B)}(kernel, B)
+end
```
st--, line 39: was there any reason for not having had this in the first place @willtebbutt? :thinking:

willtebbutt: I don't think so. This is a helpful addition.

src/mokernels/independent.jl
Crown421: I had been thinking about it, because the inverse of a kronecker product can be so very efficient, but it would mean another dependency or new output type.
If that is ok for this package, I am very happy to include it here.

src/mokernels/intrinsiccoregion.jl
Crown421: I suspected as much, based on the MOInputs type and the corresponding text in the documentation. 

I am coming from the view of matrix kernels, where K(x,y) is a matrix, as opposed to a scalar kernel k(x,y). From this perspective, maybe something like `matrixkernel(ker::MOK, x, y) where MOK <: MOKernel` for the convenience function?

On the matter of `Vector{T}`, I don't really work with heterogeneous inputs, so I am not sure what would be best here.

willtebbutt: Any of these names seem reasonable to me.

> On the matter of Vector{T}, I don't really work with heterogeneous inputs, so I am not sure what would be best here.

I think it's fine to leave the arguments un-typed, because in general we don't restrict the types of features. So something like
```julia
matrixkernel(ker::MOKernel, x, y) = ...
```
would be totally fine as far as I'm concerned.

Crown421:
Worked in most of the changes, will look at the lazy Kronecker product now.

src/mokernels/intrinsiccoregion.jl
Crown421: Maybe something like this?
```Julia
matrixkernel(ker::MOKernel, x::T, y::T)
```
Conceptually, I really like kernel functions to be symmetric.

Crown421:
It seems that Kronecker.jl is already an optional dependency of KernelFunctions. I have to think about how to implement the matrix kernels in a way that works with the optional dependency.

willtebbutt:
> It seems that Kronecker.jl is already an optional dependency of KernelFunctions. I have to think about how to implement the matrix kernels in a way that works with the optional dependency.

Please do feel free not to worry about it for now if it's not high on your priority list -- we can always revisit it later.

src/mokernels/independent.jl
st--: Ah, we already half-depend on Kronecker.jl - there is `kernelkronmat` in src/matrix/kernelkroneckermat.jl (sic! some more of that typo that need fixing as well) - maybe for now the easiest would be to add implementations of those (plus `iskroncompatible` for the MO kernel?) thoughts @willtebbutt ?

st--: half-depend = optional dependency based on Require.jl

st--:
@Crown421 you'll have to accept reviewdog's formatting suggestions (or run JuliaFormatter locally):)

src/mokernels/independent.jl
Crown421: Playing around with it some more, and `return false` does not work. 

`k.(x, permutedims(x)` returns a `Matrix{Real}` and multiplying the resulting matrix with a `rand(...)` vector yields a `Vector{Any}`, which I think is not great. 

I think I would rather see if some tests can be written to check type stability for all scalar kernels, but then I am not sure what happens to heterogeneous inputs.

Crown421:
Ok, did some reworking using traits, now there is less code duplication, and getting the lazy Kronecker product has to be made explicit.

src/mokernels/independent.jl
```diff
@@ -23,18 +23,63 @@ struct IndependentMOKernel{Tkernel<:Kernel} <: MOKernel
...
+    return kron(B, Ktmp)
+end
+
+function kernelmatrix(
+    k::IndependentMOKernel, x::MOI, y::MOI; matrixtype::MT=ExplicitKroneckerKernelMatrix()
+) where {MOI<:AbstractMOInput,MT<:KroneckerKernelMatrix}
+    @assert x.out_dim == y.out_dim
+    Ktmp = kernelmatrix(k.kernel, x.x, y.x)
+    mtype = eltype(Ktmp)
+    return _kronkernelmatrix(Ktmp, Matrix{mtype}(I, x.out_dim, x.out_dim), x, matrixtype)
```
st--, line 56: Could we avoid the out_dim x out_dim allocation by using FillArrays?

src/mokernels/independent.jl
```diff
@@ -23,18 +23,63 @@ struct IndependentMOKernel{Tkernel<:Kernel} <: MOKernel
...
+    k::IndependentMOKernel,
+    x::MOInputIsotopicByFeatures,
+    y::MOInputIsotopicByFeatures,
+)
     @assert x.out_dim == y.out_dim
-    temp = k.kernel.(x.x, permutedims(y.x))
-    return cat((temp for _ in 1:(y.out_dim))...; dims=(1, 2))
+    Ktmp = kernelmatrix(k.kernel, x.x, y.x)
+    mtype = eltype(Ktmp)
+    return _kronkernelmatrix!(K, Ktmp, Matrix{mtype}(I, x.out_dim, x.out_dim), x)
```
st--, line 82: What's the benefit of having the `_kronkernelmatrix!` redirect in this case instead of directly calling `kron!`?

src/mokernels/independent.jl
```diff
@@ -23,18 +23,63 @@ struct IndependentMOKernel{Tkernel<:Kernel} <: MOKernel
...
+
+function _kronkernelmatrix!(K, Ktmp, B, ::MOInputIsotopicByFeatures)
+    return kron!(K, Ktmp, B)
+end
+
+function _kronkernelmatrix!(K, Ktmp, B, ::MOInputIsotopicByOutputs)
+    return kron!(K, B, Ktmp)
+end
+
+function kernelmatrix!(
```
st--, line 73: Should we advertise in the docstring that this method still does _some_ allocation (because we have to construct `Ktmp`)?

src/mokernels/moinput.jl
```diff
@@ -24,7 +26,7 @@ The first `out_dim` elements represent all outputs for the first input, the seco
 
 See [Inputs for Multiple Outputs](@ref) in the docs for more info.
 """
-struct MOInputIsotopicByFeatures{S,T<:AbstractVector{S}} <: AbstractVector{Tuple{S,Int}}
+struct MOInputIsotopicByFeatures{S,T<:AbstractVector{S}} <: AbstractMOInput{Tuple{S,Int}}
```
st--, line 29: What are the pros/cons of including the tuple construct in the abstract type parameterisation? Alternatively we could also have
```suggestion
struct MOInputIsotopicByFeatures{S,T<:AbstractVector{Tuple{S,Int}}} <: AbstractMOInput{S}
```

src/mokernels/moinput.jl
```diff
@@ -1,3 +1,5 @@
+abstract type AbstractMOInput{T} <: AbstractVector{T} end
```
st--, line 1: What are the pros/cons of including the tuple construct in the abstract type parameterisation? Alternatively we could also have
```suggestion
abstract type AbstractMOInput{T} <: AbstractVector{Tuple{T,Int}} end
```

src/mokernels/independent.jl
Crown421: Note that I am dispatching on the type of the MOInput, as that affects the order in the Kronecker product.

src/mokernels/intrinsiccoregion.jl
```diff
@@ -34,10 +34,54 @@ function IntrinsicCoregionMOKernel(; kernel::Kernel, B::AbstractMatrix)
...
+
 function (k::IntrinsicCoregionMOKernel)((x, px)::Tuple{Any,Int}, (y, py)::Tuple{Any,Int})
     return k.B[px, py] * k.kernel(x, y)
 end
 
+# convenience function
+export matrixkernel
+function matrixkernel(
+    k::IntrinsicCoregionMOKernel, x::Vector{T}, y::Vector{T}
```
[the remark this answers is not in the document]
Crown421: Agree on the types. I had some documentation improvements planned for today.

src/mokernels/moinput.jl
Crown421: Sure, I am fine with that. I started with 
```julia 
abstract type AbstractMOInputabstract{Tuple{T,Int}} <: AbstractVector{Tuple{T,Int}} end
```
which does not work, so I went very simple to get it to work.

src/mokernels/independent.jl
Crown421: Sure, that makes sense to me.

src/mokernels/independent.jl
```diff
@@ -23,18 +23,63 @@ struct IndependentMOKernel{Tkernel<:Kernel} <: MOKernel
...
+end
+
+function _kronkernelmatrix(
+    Ktmp, B, ::MOInputIsotopicByOutputs, ::ExplicitKroneckerKernelMatrix
+)
+    return kron(B, Ktmp)
+end
+
+function kernelmatrix(
+    k::IndependentMOKernel, x::MOI, y::MOI; matrixtype::MT=ExplicitKroneckerKernelMatrix()
```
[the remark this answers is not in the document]
Crown421: That actually makes more sense to me, instead of an explicit keyword, make it an explicit method.

src/mokernels/moinput.jl
Crown421: Tried the suggestion, it does precompile, but then it fails the tests, as `collect` no longer knows what to do.

src/mokernels/independent.jl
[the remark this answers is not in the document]
Crown421: made a change, probably cleaner.

src/mokernels/independent.jl
willtebbutt: Thinking about this a bit more, this suggestion would actually be totally fine for `kernelmatrix_diag`, because `px == py` is always true, and given that you've added `kernelmatrix`, I _think_ this fallback will only ever get hit for `kernelmatrix_diag` in general usage (unless someone is calling the kernel directly, which is generally useful for debugging but isn't what people should be doing if they need performance.

src/mokernels/intrinsiccoregion.jl
willtebbutt: I agree that they ought to be symmetric in the sense that `k(x, y) == k(y, x)` (up to numerics), but I don't believe that extends to imposing type-constraints on the arguments. For example, `SEKernel()(5.0, 3f0)` seems like a perfectly reasonable thing to evaluate to me (both arguments represent real numbers, just a different levels of precision), but using the above diagonal dispatch constraint would preclude it. I'm sure there are other examples.

More generally, there's nothing to say that a kernel _can't_ be defined on some really weird domain that is represented in Julia by multiple types. For example, there's nothing to _stop_ someone defining a kernel
```julia
function (k::MyKernel)(x::Union{Real, AbstractVector}, y::Union{Real, AbstractVector})
    # some code
end
```
For such a kernel, something like `k(5.0, randn(3))` is perfectly reasonable.

src/mokernels/moinput.jl
```diff
@@ -1,3 +1,5 @@
+abstract type AbstractMOInput{T} <: AbstractVector{Tuple{T,Int}} end
```
willtebbutt, line 1: Could we make this an alias, rather than a type? The problem with this subtyping is that it's potentially confusing for users: for some type `T`, `Vector{Tuple{T, Int}}` is a perfectly valid collection of inputs for any multi-output kernel. However, under the current definition it would not be a subtype of `AbstractMOInput`, possibly suggesting that it's not a valid input to a multi-output kernel.

Is there a particular reason that this supertype is necessary?

Also, `AbstractMOInputs` rather than `AbstractMOInput` would be my preference, because it represents a collection of inputs rather than a single input (I think @st-- pointed this out in another PR at some point).

Crown421: My reasoning was that the specialized `kernelmatrix` only works with the `MOInputsIsotropics`, possibly not for general `Vector{Tuple{T, Int}}`.
I have changed this now to using a `Union`, which should do the same thing, without needing the abstract type.

src/mokernels/independent.jl
Crown421: Sorry, I missed these comments. I like the suggestion, it also returns the correct result. Have changed it.

src/mokernels/intrinsiccoregion.jl
Crown421: As someone who is mostly a mathematician, I am puzzled as to why someone would do something like that, but I agree that it should remain possible. Have changed.

Crown421:
I noticed, that `kron!` makes the test fail, because it only exists from Julia 1.6 up. What should be done here? Comment the code/tests out until the compatibility of this package is raised to 1.6,  maybe?

src/mokernels/intrinsiccoregion.jl
willtebbutt: I mainly work with GPs, so I'll attempt to give a GP example. Apologies in advance for the unholy mixture of maths and Julia.

The main place where I've seen interesting collections of input types is when you want to treat a collection of related GPs as a single GP on an extended input domain.

Consider a GP `f1`, with index set R^2, and another GP `f2` given by integrating over one of the dimensions of `f1`:
```julia
f2(x) := ∫ f1([x, y]) dμ(y)
```
The index set of `f2` is just R.

You can treat these two related GPs as a single GP, and the way that such a process would be represented in Stheno.jl would be via a single GP `f` with index set
```julia
{(x, :f1) | x in R} ∪ {(x, :f2) | x in R^2}
```
where `:f1` and `:f2` are just symbols. Then for some input `(x, symbol)`, `f` is something like
```julia
f((x, symbol)) = symbol == :f1 ? f1(x) : f2(x)
```
So the kernel of `f` would have quite an interesting domain, not obviously representable by a single type.

Crown421:
I notice that the documentation of MO kernels is somewhat inconsistent, in most cases the number of outputs is indicated by ``m``, but in the `LinearMizingModelKernel`, ``m`` is the number of basis kernels and ``p`` is the number of outputs. 
Are there objections to making this consistent?

willtebbutt:
> I notice that the documentation of MO kernels is somewhat inconsistent, in most cases the number of outputs is indicated by m, but in the LinearMizingModelKernel, m is the number of basis kernels and p is the number of outputs.
Are there objections to making this consistent?

None whatsoever, I'm pro- consistency.

> I noticed, that kron! makes the test fail, because it only exists from Julia 1.6 up. What should be done here? Comment the code/tests out until the compatibility of this package is raised to 1.6, maybe?

I _think_ it's probably fine to do something like
```julia
if VERSION >= v"1.6"
    # code for kernelmatrix!
end
```
in both the source. In older versions it should just hit the fallback and be slow.

test/mokernels/intrinsiccoregion.jl
```diff
@@ -2,25 +2,50 @@
...
+    icoregionkernel2 = IntrinsicCoregionMOKernel(kernel, B)
+    @test icoregionkernel == icoregionkernel2
+
     @test icoregionkernel.B == B
     @test icoregionkernel.kernel == kernel
     @test icoregionkernel(X[1], X[1]) ≈ B[X[1][2], X[1][2]] * kernel(X[1][1], X[1][1])
     @test icoregionkernel(X[1], X[end]) ≈ B[X[1][2], X[end][2]] * kernel(X[1][1], X[end][1])
 
+    # test convenience function using kronecker product
+    @test matrixkernel(icoregionkernel, X.x[1], X.x[2]) ≈
```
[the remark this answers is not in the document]
Crown421: That seems sensible, will check out in more detail tomorrow.

src/mokernels/intrinsiccoregion.jl
Crown421: Very interesting, I have never come across something like this. Need to look into it in more detail.

src/mokernels/lmm.jl
```diff
@@ -21,6 +22,7 @@ struct LinearMixingModelKernel{Tk<:AbstractVector{<:Kernel},Th<:AbstractMatrix}
     K::Tk
     H::Th
 end
+# does it maybe make sense to check that length(K) and size(H,1) are equal?
```
st--, line 25: Yeah I'd be happy with adding a check of that:)

src/mokernels/lmm.jl
```diff
@@ -1,7 +1,8 @@
 @doc raw"""
-    LinearMixingModelKernel(g, e::MOKernel, A::AbstractMatrix)
+    LinearMixingModelKernel(k::Kernel, H::AbstractMatrix)
+    LinearMixingModelKernel(Tk::AbstractVector{<:Kernel},Th::AbstractMatrix)
 
-Kernel associated with the linear mixing model.
+Kernel associated with the linear mixing model, taking a vector of `m` kernels and a `m` by `p` matrix H for a function with `p` outputs. Also accepts a single kernel `k` for use across all `m` basis vectors.
```
st--, line 5:
```suggestion
Kernel associated with the linear mixing model, taking a vector of `m` kernels and a `m × p` matrix H for a function with `p` outputs. Also accepts a single kernel `k` for use across all `m` basis vectors. 
```

src/matrix/kernelkroneckermat.jl
```diff
@@ -25,3 +25,35 @@ end
...
+
+@inline ismatrixkroncompatible(κ::MOKernel) = false # Default return for kernels
+@inline ismatrixkroncompatible(κ::IndependentMOKernel) = true
+@inline ismatrixkroncompatible(κ::IntrinsicCoregionMOKernel) = true
+
+function _kroneckerkernelmatrix(Ktmp, B, ::MOInputIsotopicByFeatures)
+    return Kronecker.kronecker(Ktmp, B)
+end
+
+function _kroneckerkernelmatrix(K, Ktmp, B, ::MOInputIsotopicByOutputs)
```
st--, line 37: The `K` seems like it shouldn't be here:
```suggestion
function _kroneckerkernelmatrix(Ktmp, B, ::MOInputIsotopicByOutputs)
```

src/mokernels/independent.jl
```diff
@@ -23,18 +23,44 @@ struct IndependentMOKernel{Tkernel<:Kernel} <: MOKernel
...
 
+# kernel function should be symmetric
+# would really like (κ::IndependentMOKernel)((x, px)::Tuple{T,Int}, (y, py)::Tuple{T,Int}) where T, but seems to cause autodiff problems
 function (κ::IndependentMOKernel)((x, px)::Tuple{Any,Int}, (y, py)::Tuple{Any,Int})
-    if px == py
-        return κ.kernel(x, y)
-    else
-        return 0.0
-    end
+    return κ.kernel(x, y) * (px == py)
```
st--, line 29: that seems like a good resolution!

src/mokernels/independent.jl
```diff
@@ -23,18 +23,44 @@ struct IndependentMOKernel{Tkernel<:Kernel} <: MOKernel
...
+end
+
+function _kronkernelmatrix(Ktmp, B, ::MOInputIsotopicByFeatures)
```
st--, line 32: It's starting to get a bit confusing with _kronkernelmatrix, _kroneckerkernelmatrix, kernelkronmat... how about at least clarifying in the method name that this is multioutput-specific? e.g.
```suggestion
function _mokernelmatrix_kron(::MOInputIsotopicByFeatures, Kfeatures, Koutputs)
```
(moving the argument on which we actually dispatch to the front)

src/mokernels/moinput.jl
```diff
@@ -194,3 +194,5 @@ function prepare_isotopic_multi_output_data(x::AbstractVector, y::RowVecs)
     length(x) == length(y) || throw(ArgumentError("length(x) not equal to length(y)."))
     return MOInputIsotopicByOutputs(x, size(y.X, 2)), vec(y.X)
 end
+
+MOInputsUnion = Union{MOInputIsotopicByFeatures,MOInputIsotopicByOutputs}
```
st--, line 198:
```suggestion
MOInputsIsotopic = Union{MOInputIsotopicByFeatures,MOInputIsotopicByOutputs}
```
or
```suggestion
IsotopicMOInputs = Union{MOInputIsotopicByFeatures,MOInputIsotopicByOutputs}
```
- happy to keep the `Union` suffix as well - but it's crucial that this makes clear it's about *isotopic* MO inputs

src/matrix/kernelkroneckermat.jl
Crown421: True, too much copy and paste.

src/mokernels/mokernel.jl
```diff
@@ -4,3 +4,16 @@
...
 """
 abstract type MOKernel <: Kernel end
+
+"""
+    matrixkernel(k::MOK, x, y)
+    matrixkernel(k::IndependentMOKernel, x, y(; outputsize))
+
+Convenience function to compute the matrix kernel for two inputs `x` and `y`. The `outputsize` keyword is only required for the `IndependentMOKernel` to indicated the number of outputs. 
+"""
+function matrixkernel(k::MOK, x, y; outputsize) where {MOK<:MOKernel}
```
thomasgudjonwright, line 14: Is there a specific reason for using `MOInputIsotopicByFeatures` here rather than `MOInputIsotopicByOutputs`? The "default" we have right now is the latter, as it has an alias `MOInput`. The only reason I mention this is that `kernelmatrix` will return a different ordering depending on which input type is used, so we should be cautious about the assumptions we enforce on the user.

eg
```
julia> x = ColVecs(rand(2,2));

julia> y = ColVecs(rand(2,2));

julia> x_ByO = kf.MOInputIsotopicByOutputs(x_obs, 2);

julia> y_ByO = kf.MOInputIsotopicByOutputs(y_obs, 2);

julia> x_ByF = kf.MOInputIsotopicByFeatures(x_obs, 2);

julia> y_ByF = kf.MOInputIsotopicByFeatures(y_obs, 2);

julia> H = rand(2,2);

julia> kernelmatrix(LinearMixingModelKernel([SEKernel(), SEKernel()], H), x_ByF, y_ByF)
4×4 Matrix{Float64}:
 0.25171   0.480092  0.222108  0.423631
 0.480092  1.13149   0.423631  0.998427
 0.250325  0.477451  0.264486  0.504461
 0.477451  1.12527   0.504461  1.18893

julia> kernelmatrix(LinearMixingModelKernel([SEKernel(), SEKernel()], H), x_ByO, y_ByO)
4×4 Matrix{Float64}:
 0.25171   0.222108  0.480092  0.423631
 0.250325  0.264486  0.477451  0.504461
 0.480092  0.423631  1.13149   0.998427
 0.477451  0.504461  1.12527   1.18893
```

Crown421: That is true, the ordering is different. I chose `MOInputIsotopicByFeatures`, because it makes sense from a matrix kernel perspective if you compute the kernel matrix as a block matrix of matrix kernels.
This means, coming from a collection of inputs `X` and computing 
```Julia
matrixkernel.(k, X, permutedims(X))
```

Crown421:
Worked in the last comments. Baring a last pass over the documentation, I feel this could be basically ready.

src/mokernels/mokernel.jl
willtebbutt, line 14:
```suggestion
function matrixkernel(k::MOKernel, x, y; outputsize)
```

src/mokernels/mokernel.jl
```diff
@@ -4,3 +4,16 @@
 Abstract type for kernels with multiple outpus.
 """
 abstract type MOKernel <: Kernel end
+
+"""
+    matrixkernel(k::MOK, x, y)
+    matrixkernel(k::IndependentMOKernel, x, y(; outputsize))
+
+Convenience function to compute the matrix kernel for two inputs `x` and `y`. The `outputsize` keyword is only required for the `IndependentMOKernel` to indicated the number of outputs.
```
willtebbutt, line 12: Style: line length

Additionally, it would be helpful to add a doctest here.

src/mokernels/lmm.jl
```diff
@@ -32,6 +34,10 @@ function (κ::LinearMixingModelKernel)((x, px)::Tuple{Any,Int}, (y, py)::Tuple{A
     return sum(κ.H[i, px] * κ.K[i](x, y) * κ.H[i, py] for i in 1:length(κ.K))
 end
 
+function matrixkernel(k::LinearMixingModelKernel, x, y)
```
willtebbutt, line 37: Needs a test

src/mokernels/slfm.jl
```diff
@@ -33,6 +33,10 @@ function (κ::LatentFactorMOKernel)((x, px)::Tuple{Any,Int}, (y, py)::Tuple{Any,
     return cov_f + κ.e((x, px), (y, py))
 end
 
+function matrixkernel(k::LatentFactorMOKernel, x, y)
```
willtebbutt, line 36: Needs a test

willtebbutt (commented):
Agreed. We're nearly there. Just a couple of bits of housekeeping to sort out (and @st-- 's and @thomasgudjonwright 's comments to resolve)

Re the docs, the only new bit of functionality is `matrixkernel` which I think just needs a doctest -- other than that I think I'm happy for the sake of this PR (I think our multi-output docs probably need some attention more generally, but that's not for now).

src/mokernels/mokernel.jl
Crown421: turns out, in this particular case it actually does not matter which type is used within the function, as it is only two single inputs, and the `...Features` and `...Outputs` only affects the ordering of the entire kernel matrix.

src/mokernels/mokernel.jl
Crown421: I have unfortunately never used documenter, will need to read up on this. Style should be fixed by formatter now.

src/mokernels/lmm.jl
Crown421: Done

src/mokernels/slfm.jl
Crown421: Done

Crown421:
Added tests to `lmm` and `slfm`, also put in the `test_interface`. Fixed a test mistake due to post 1.5 kwarg syntax that failed on CI. Apart from documentation issues, this should be fine now. 

As I am almost exclusive working with Multi-Output systems, I am happy to gradually improve the documentation in the near future.

src/mokernels/lmm.jl
Crown421: Still outstanding.

willtebbutt:
> Lastly, can this become a new release?

Yes, definitely. Our policy is that _any_ changes to `/src` warrant a release -- in this case we'll just need to bump the patch version, because there are no breaking changes, which makes life particularly straightforward. So please do increment the patch version in the Project.toml before we merge.

Please tag me when the remaining items are resolved, and I'll approve + merge etc.

src/matrix/kernelkroneckermat.jl
```diff
@@ -25,3 +25,37 @@ end
     k(x,x') = ∏ᵢᴰ k(xᵢ,x'ᵢ)
 """
 @inline iskroncompatible(κ::Kernel) = false # Default return for kernels
+
+@inline ismatrixkroncompatible(κ::MOKernel) = false # Default return for kernels
+@inline ismatrixkroncompatible(κ::IndependentMOKernel) = true
+@inline ismatrixkroncompatible(κ::IntrinsicCoregionMOKernel) = true
```
st--, line 31: Why would this be separate from `iskroncompatible`? If it's not possible to reuse `iskroncompatible`, could you add a docstring explaining why this one is different (and how so)?

src/matrix/kernelkroneckermat.jl
```diff
@@ -25,3 +25,37 @@ end
...
+    return _mo_kernelmatrix_kron(Ktmp, Eye{mtype}(x.out_dim), x)
+end
+
+function kernelkronmat(
+    k::IntrinsicCoregionMOKernel, x::MOI, y::MOI
+) where {MOI<:IsotopicMOInputsUnion}
+    @assert x.out_dim == y.out_dim
+    Ktmp = kernelmatrix(k.kernel, x.x, y.x)
+    return _mo_kernelmatrix_kron(Ktmp, k.B, x)
+end
```
st--, line 56: We could unify these if we had a function that returns the "covariance over outputs" based on the kernel type

src/matrix/kernelkroneckermat.jl
```diff
@@ -25,3 +25,37 @@ end
...
+
+function kernelkronmat(k::MOK, x::MOI) where {MOI<:IsotopicMOInputsUnion,MOK<:MOKernel}
+    @assert ismatrixkroncompatible(κ) "The chosen kernel is not compatible for Kronecker matrices"
+    return kernelkronmat(k, x, x)
```
st--, line 60: This is going to be (somewhat) less efficient than separate implementations that then only call `kernelmatrix(k.kernel, x.x)`

src/mokernels/lmm.jl
```diff
@@ -32,6 +34,10 @@ function (κ::LinearMixingModelKernel)((x, px)::Tuple{Any,Int}, (y, py)::Tuple{A
...
+    return matrixkernel(k, x, y; outputsize=size(k.H, 2))
```
st--, line 38: how does this not turn into an infinite loop ?

Crown421: Because the inner function has/ requires a kwarg

src/matrix/kernelkroneckermat.jl
Crown421: I don't think I understand this?

src/matrix/kernelkroneckermat.jl
Crown421: Good question. I did this because I wanted to avoid confusion with the first two functions, because the MO kernel matrix construction is distinct from the kronecker construction that was there already. 
I realize now that I should have used a different name than `kernelkronmat` for this.

src/matrix/kernelkroneckermat.jl
```diff
@@ -25,3 +25,37 @@ end
...
+
+function kernelkronmat(
+    k::IntrinsicCoregionMOKernel, x::MOI, y::MOI
+) where {MOI<:IsotopicMOInputsUnion}
+    @assert x.out_dim == y.out_dim
+    Ktmp = kernelmatrix(k.kernel, x.x, y.x)
+    return _mo_kernelmatrix_kron(Ktmp, k.B, x)
+end
+
+function kernelkronmat(k::MOK, x::MOI) where {MOI<:IsotopicMOInputsUnion,MOK<:MOKernel}
```
willtebbutt, line 58:
```suggestion
function kernelkronmat(k::MOKernel, x::IsotopicMOInputsUnion)
```
I believe the short-form syntax should be fine here, no?

src/mokernels/lmm.jl
st--: ah, didn't realize dispatch depended on kwargs as well, I thought it was only on the positional args...

st--: this seems to be a bug (https://github.com/JuliaLang/julia/issues/9498) that is only not fixed yet because it would break backwards compatibility - I would prefer not to abuse implementation details that go counter to [what the Julia docs say](https://docs.julialang.org/en/v1/manual/methods/#Note-on-Optional-and-keyword-Arguments).

src/matrix/kernelkroneckermat.jl
st--: Along the lines of
```suggestion
_mo_output_covariance(k::IndependentMOKernel, out_dim) = Eye{Bool}(out_dim)
function _mo_output_covariance(k::IndependentMOKernel, out_dim)
    @assert size(k.B) == (out_dim, out_dim)
    return k.B
end

function kernelkronmat
    k::Union{IndependentMOKernel, IntrinsicCoregionMOKernel}, x::MOI, y::MOI
) where {MOI<:IsotopicMOInputsUnion}
    @assert x.out_dim == y.out_dim
    Kfeatures = kernelmatrix(k.kernel, x.x, y.x)
    Koutputs = _mo_output_covariance(k, x.out_dim)
    return _mo_kernelmatrix_kron(Kfeatures, Koutputs)
end
```

src/matrix/kernelkroneckermat.jl
st--: `kernelkronmat` seems fine to me - it's basically saying "please give me a lazy Kronecker" whether it's because we have outputs-as-inputs or seperable-on-grid-of-features or something else

src/matrix/kernelkroneckermat.jl
st--: that is, do we want separate methods for the sake of optimal efficiency? @willtebbutt

src/mokernels/independent.jl
st--: I'm not quite happy with this being a different pattern from kernelkronmat when they should otherwise be the same, `kron` vs `Kronecker` being the only difference. Could it be a single method that returns either dense kron() or lazy Kronecker? If that would lead to type instability issues, could dispatch on a Val{} maybe? lazy=Val{true} vs lazy=Val{false} (it's just an internal function anyways)

src/matrix/kernelkroneckermat.jl
willtebbutt: It would probably be nice, but I'm not going to insist on it going in in this PR provided it gets sorted in a follow up.

src/mokernels/mokernel.jl
st--:
```suggestion
Convenience function to compute the matrix kernel for two inputs `x` and `y`. The `outputsize` keyword is only required for the `IndependentMOKernel` to indicate the number of outputs. 
```
Also, how about instead just having
```julia
matrixkernel(k::IndependentMOKernel, x, y) = error("You must provide the number of outputs, call `matrixkernel(k, x, y, out_dim)`")
matrixkernel(k::IndependentMOKernel, x, y, out_dim) = ...
```

src/mokernels/mokernel.jl
```diff
@@ -4,3 +4,16 @@
 Abstract type for kernels with multiple outpus.
 """
 abstract type MOKernel <: Kernel end
+
+"""
+    matrixkernel(k::MOK, x, y)
+    matrixkernel(k::IndependentMOKernel, x, y(; outputsize))
```
st--, line 10: Would be nice to be consistent with the MOInput* types and call it `out_dim` instead

src/matrix/kernelkroneckermat.jl
Crown421: I am never quite sure when one or the other is needed, or if they are actually the same.

src/mokernels/intrinsiccoregion.jl
```diff
@@ -34,10 +34,36 @@ function IntrinsicCoregionMOKernel(; kernel::Kernel, B::AbstractMatrix)
...
+function IntrinsicCoregionMOKernel(kernel::Kernel, B::AbstractMatrix)
+    return IntrinsicCoregionMOKernel{typeof(kernel),typeof(B)}(kernel, B)
+end
+
 function (k::IntrinsicCoregionMOKernel)((x, px)::Tuple{Any,Int}, (y, py)::Tuple{Any,Int})
     return k.B[px, py] * k.kernel(x, y)
 end
 
+function matrixkernel(k::IntrinsicCoregionMOKernel, x, y)
+    return matrixkernel(k, x, y; outputsize=size(k.B, 1))
```
st--, line 46: minor suggestion:
```suggestion
    return matrixkernel(k, x, y; outputsize=size(k.B, 2))
```
doesn't make any difference as `B` must be square, but then it's more consistent with the LMC one:)

src/mokernels/mokernel.jl
```diff
@@ -4,3 +4,16 @@
...
+
+Convenience function to compute the matrix kernel for two inputs `x` and `y`. The `outputsize` keyword is only required for the `IndependentMOKernel` to indicated the number of outputs. 
+"""
+function matrixkernel(k::MOKernel, x, y; outputsize)
+    @assert size(x) == size(y)
```
st--, line 15:
```suggestion
```
`x` and `y` could be types on which `size` isn't defined (e.g. a string)

src/mokernels/lmm.jl
st--: Instead of the layers of indirection, how about
```julia
matrixkernel(k::IndependentMOKernel, x, y) = k.kernel(x, y) * I
matrixkernel(k::IntrinsicCoregionMOKernel, x, y) = k.kernel(x, y) * k.B
```
etc.?

src/matrix/kernelkroneckermat.jl
st--: > It would probably be nice, but I'm not going to insist on it going in in this PR provided it gets sorted in a follow up.

@Crown421 if you don't want to do it in this PR, please open a new issue for it instead then: )

src/mokernels/lmm.jl
Crown421: Interesting, I have been doing this for a while, good to know that it will go away at some point.

st--:
General thought/comment: I think we should try to go for smaller PRs, it makes it much easier (and hence faster) to review, respond to reviews, and get it merged in overall - even if it's a bunch of small related (but independent!) changes like here. For all of us to remember next time: )

@Crown421 : please join us on https://join.slack.com/t/turingjl/shared_invite/zt-ubbtg7r8-xtg4WjXEQTG_jvE~estuKw if you like, for future development:)

src/mokernels/independent.jl
st--: Or maybe even neater (because it more explicitly shows what we're evaluating here):
```suggestion
    return κ.kernel(x, y) * I[px, py]
```
I benchmarked it - same speed.

src/mokernels/lmm.jl
Crown421: > matrixkernel(k::IndependentMOKernel, x, y) = k.kernel(x, y) * I

Unfortunately, `I` does not have a size. The current implementation of the `IndependentMOKernel` does not have a output dimension attached, even though I personally would prefer it to be a special case of the the `IntrinsicCoregionMOKernel` where `B = Matrix(I, out_dim, out_dim`.

st--:
Yes exactly, that _is_ why I suggested `I` - but IMO that's the right thing
to do, you can still index into it, use it in any way you like, and then we
don't have to pass in an output dim argument. If you really need a concrete
matrix in your specific use case (not quite sure what that would be, maybe
explain it to us?) You could still have a wrapper function that turns it
into a Matrix.

On Saturday, August 14, 2021, Steffen Ridderbusch ***@***.***>
wrote:
> @Crown421 commented on this pull request.
>
> ________________________________
>
> In src/mokernels/lmm.jl:
>
>> @@ -32,6 +34,10 @@ function (κ::LinearMixingModelKernel)((x,
px)::Tuple{Any,Int}, (y, py)::Tuple{A
>
> return sum(κ.H[i, px] * κ.K[i](x, y) * κ.H[i, py] for i in 1:length(κ.K))
>
> end
>
>
>
> +function matrixkernel(k::LinearMixingModelKernel, x, y)
>
> + return matrixkernel(k, x, y; outputsize=size(k.H, 2))
>
> matrixkernel(k::IndependentMOKernel, x, y) = k.kernel(x, y) * I
>
> Unfortunately, I does not have a size. The current implementation of the
IndependentMOKernel does not have a output dimension attached, even though
I personally would prefer it to be a special case of the the
IntrinsicCoregionMOKernel where B = Matrix(I, out_dim, out_dim.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or unsubscribe.
> Triage notifications on the go with GitHub Mobile for iOS or Android.<
https://ci5.googleusercontent.com/proxy/bFFEnavG2QSKKaSz5xB6bok54rypvA-s3bloN8Mfv3cH0epzxHfXg1X_neomyXBnlkgNs9OAlX-q-pog6LJoiJFPPHfP6PBrVgBHT8uaGLVEFzCQHuuiZBeQhWN46iW0praZaHXX50W84qIcuhpko8P8PI2F84eiCROGZ9AYeBoYdzvO0CQzwRdZ1y9wlpbHLWslktvTJKWW9Bow6hsSXM5k_pVgBh6D6LWkOFGthsQJ9mHuRNRB=s0-d-e1-ft#https://github.com/notifications/beacon/ABL7FD3T4WIRJELXSVAQKVDT4Y5HNA5CNFSM5B27WKQ2YY3PNVWWK3TUL52HS4DFWFIHK3DMKJSXC5LFON2FEZLWNFSXPKTDN5WW2ZLOORPWSZGOFOCC3UA.gif>

src/matrix/kernelkroneckermat.jl
willtebbutt: In the context of method definitions, the only difference is some compiler heuristics around when or when not to specialise, for a very particular set of types: https://docs.julialang.org/en/v1/manual/performance-tips/#Be-aware-of-when-Julia-avoids-specializing . For the most part, these heuristics are there for a good reason (compiler latency), so it's usually only worth overriding them in special cases. In this instance, I don't believe there's any difference whatsoever between the two.

The place where it matters most of the time is inside `struct` declarations.

willtebbutt:
@Crown421 how would you feel about breaking this PR up into the following three PRs:
1. The current PR should just have the functionality relating to accelerating the computation of `kernelmatrix` and `kernelmatrix!` using `kron` and `kron!` + associated tests + docstring tweaks.
2. Another PR should contain lazy stuff using `Kronecker.jl`
3. A third PR should contain the `matrixkernel` stuff.

There's just quite a lot of stuff in this PR now, and it's making it hard to make progress on any piece of it because we need to reach agreement on everything. I know I said I thought it was basically there, but going back over it I realise I've got a few lingering concerns that I didn't notice yesterday.

Crown421:
> you can still index into it, use it in any way you like, and then we don't have to pass in an output dim argument.

That is a good point, in my mind I was expecting `matrixkernel` to always return the explicit matrix that corresponds to the current problem, but in the same way that `I` itself is almost always good enough, I can't really think of a reasonable use case where the explicit matrix is strictly required, and some scaled `Eye` is not sufficient.

Crown421:
> @Crown421 how would you feel about breaking this PR up into the following three PRs:
> 
>     1. The current PR should just have the functionality relating to accelerating the computation of `kernelmatrix` and `kernelmatrix!` using `kron` and `kron!` + associated tests + docstring tweaks.
> 
>     2. Another PR should contain lazy stuff using `Kronecker.jl`
> 
>     3. A third PR should contain the `matrixkernel` stuff.
> 
> 
> There's just quite a lot of stuff in this PR now, and it's making it hard to make progress on any piece of it because we need to reach agreement on everything. I know I said I thought it was basically there, but going back over it I realise I've got a few lingering concerns that I didn't notice yesterday.
> 
> Splitting things up in this way will probably result in us getting this functionality in more quickly than doing it in this single PR.

Generally I am happy with this, I will split things off either later today or tomorrow morning.

test/mokernels/intrinsiccoregion.jl
```diff
@@ -2,25 +2,56 @@
...
+    # kernelmatrix
+    KernelFunctions.TestUtils.test_interface(icoregionkernel, XIF, YIF, ZIF)
+
+    KernelFunctions.TestUtils.test_interface(icoregionkernel, XIO, YIO, ZIO)
 
     KernelFunctions.TestUtils.test_interface(
         icoregionkernel, Vector{Tuple{Float64,Int}}; dim_out=dims.out
     )
 
+    # in-place
```
willtebbutt, line 43: Unless I'm missing the point of the tests, I think this will be tested in `test_interface` anyway? The methods will exist in versions earlier than 1.6, they'll just hit the slow fallback.

Perhaps this block should be removed?

willtebbutt (commented):
Many thanks for reducing the size of this PR, it's now much more easily digestible.

I've left a couple of minor style-related things, but I think they're fairly inconsequential. The main thing I'd like to get sorted is the argument names for `_kronkernelmatrix`, and the function's name itself.

Crown421:
Ok, I have removed the in place tests, and changed a few more small things. 

> I agree with @st-- regarding argument names in particular -- Kfeatures and Koutputs are quite a bit more informative than Ktmp and B. The function name also makes more sense to me.
Re function names, the helper function should not really be visible to a user (which I tried to indicate with the leading `_`, as is done in Distances.jl. I am not sure how to improve the function name further. 
Regarding argument names, I looked at a few papers, and it is not clear to me why `B` could be called `Koutputs`. The paper cited in the docs uses `B` and calls it the "coregionalization matrix", so I would personally prefer to keep it consistent with the cited paper.

willtebbutt:
>  Regarding argument names, I looked at a few papers, and it is not clear to me why B could be called Koutputs. The paper cited in the docs uses B and calls it the "coregionalization matrix", so I would personally prefer to keep it consistent with the cited paper.

The rationale is that `B` represents the "between output" kernel, while `Ktmp` represents the "between feature" kernel (features as opposed to inputs). I can see your point though, so we'll stick with the current argument names.

Could you please remove the `temporary_script.jl` file and bump the patch version. I'll approve once that's done, and merge + tag after that once CI passes.

willtebbutt:
Actually, one other quick thing.

> I am not sure how to improve the function name further.

 I think sticking the word "helper" on the end was a good move. Could you please rename to `_kernelmatrix_kron_helper` though? We go with underscores wherever possible, as mandated by the style guide we follow (somehow `kernelmatrix` slipped through, and now we're stuck with it).

Crown421:
Alright, done all of those things. I am sorry that this was such a long process, I learned  quite a bit though. 
@willtebbutt

Crown421:
Ah, wait, something is still wrong.

willtebbutt:
No problem at all. Thanks for sticking with it and working through everything. The first PR always takes as while -- your follow up PRs will almost certainly progress much faster now that we all know what's going on better.

willtebbutt:
@Crown421 looks like there's some `using`s missing in the tests or something like that.

Crown421:
> @Crown421 looks like there's some `using`s missing in the tests or something like that.

I think I fixed it?

willtebbutt (approved):
LGTM. Will merge + tag.

PR #357, opened 2021-08-11
Update CompatHelper.yml to automatically bump version for main project

st--:


st--:
I *think* it will actually work with a single call to CompatHelper.main, if https://github.com/JuliaRegistries/CompatHelper.jl/pull/365 gets merged. It's also still waiting on CompatHelper v3 getting released (https://github.com/JuliaRegistries/CompatHelper.jl/pull/361).

willtebbutt (approved):
Version 3 is now out, so hopefully this will work nicely!

PR #363, opened 2021-08-28
Matrixkernel convenience functions and related performancec improvements

Crown421:
**Summary**
Following up from #354 , this pull request contains the `matrixkernel` convenience function, some additional tests, a small fix, and performance improvements that came up along the way. 

**Proposed changes**
- Adding `matrixkernel`
- Include changes discussed in #354 
- Fix for helper method in fbm kernel
- Performance improvement for lmm kernelmatrix
- Fix specialized implementation for slfm to work with both `MOInput` types
- Related tests

**What alternatives have you considered?**

PR #364, opened 2021-08-30
Add lazy kronecker product for matrix kernels, if Kronecker.jl is loaded

Crown421:
**Summary**
Following #354 this PR adds lazy Kronecker products for the `IndependentMOKernel` and the `IntrinsicCoregionMOKernel` via an optional dependency on Kronecker.jl. Also includes comments and thoughts from the previous PR. 

**Proposed changes**
- Add lazy `kronecker_kernelmatrix`

The name for `kronecker_kernelmatrix` is perhaps not ideal, maybe an additional `mo` suffix/prefix is needed. I think conflating this with `kernelkronmat` would be confusing.  

I have also considered making the `_mo_output_covariance` function apply also for the regular `kernelmatrix`, but this may cause conflicts with #363 , and would have to be done once both are resolved. 

**Breaking changes**
None.

### DIFF
--- a/src/matrix/kernelkroneckermat.jl
+++ b/src/matrix/kernelkroneckermat.jl
@@ -30,32 +30,47 @@ end
 @inline ismatrixkroncompatible(κ::IndependentMOKernel) = true
 @inline ismatrixkroncompatible(κ::IntrinsicCoregionMOKernel) = true
 
-function _mo_kernelmatrix_kron(::MOInputIsotopicByFeatures, K, B)
-    return Kronecker.kronecker(K, B)
+function _kernelmatrix_kroneckerjl_helper(::MOInputIsotopicByFeatures, Kfeatures, Koutputs)
+    return Kronecker.kronecker(Kfeatures, Koutputs)
 end
 
-function _mo_kernelmatrix_kron(::MOInputIsotopicByOutputs, K, B)
-    return Kronecker.kronecker(B, K)
+function _kernelmatrix_kroneckerjl_helper(::MOInputIsotopicByOutputs, Kfeatures, Koutputs)
+    return Kronecker.kronecker(Koutputs, Kfeatures)
 end
 
-function kernelkronmat(
-    k::IndependentMOKernel, x::MOI, y::MOI
-) where {MOI<:IsotopicMOInputsUnion}
+_mo_output_covariance(k::IndependentMOKernel, out_dim) = Eye{Bool}(out_dim)
+function _mo_output_covariance(k::IntrinsicCoregionMOKernel, out_dim)
+    @assert size(k.B) == (out_dim, out_dim)
+    return k.B
+end
+
+function kronecker_kernelmatrix(
+    k::Union{IndependentMOKernel,IntrinsicCoregionMOKernel},
+    x::IsotopicMOInputsUnion,
+    y::IsotopicMOInputsUnion,
+)
     @assert x.out_dim == y.out_dim
-    Ktmp = kernelmatrix(k.kernel, x.x, y.x)
-    mtype = eltype(Ktmp)
-    return _mo_kernelmatrix_kron(Ktmp, Eye{mtype}(x.out_dim), x)
+    Kfeatures = kernelmatrix(k.kernel, x.x, y.x)
+    Koutputs = _mo_output_covariance(k, x.out_dim)
+    return _kernelmatrix_kroneckerjl_helper(Kfeatures, Koutputs, x)
 end
 
-function kernelkronmat(
-    k::IntrinsicCoregionMOKernel, x::MOI, y::MOI
-) where {MOI<:IsotopicMOInputsUnion}
-    @assert x.out_dim == y.out_dim
-    Ktmp = kernelmatrix(k.kernel, x.x, y.x)
-    return _mo_kernelmatrix_kron(Ktmp, k.B, x)
+function kronecker_kernelmatrix(
+    k::Union{IndependentMOKernel,IntrinsicCoregionMOKernel}, x::IsotopicMOInputsUnion
+)
+    Kfeatures = kernelmatrix(k.kernel, x.x)
+    Koutputs = _mo_output_covariance(k, x.out_dim)
+    return _kernelmatrix_kroneckerjl_helper(Kfeatures, Koutputs, x)
 end
 
-function kernelkronmat(k::MOK, x::MOI) where {MOI<:IsotopicMOInputsUnion,MOK<:MOKernel}
-    @assert ismatrixkroncompatible(κ) "The chosen kernel is not compatible for Kronecker matrices"
-    return kernelkronmat(k, x, x)
+function kronecker_kernelmatrix(
+    k::MOKernel, x::IsotopicMOInputsUnion, y::IsotopicMOInputsUnion
+)
+    return throw(
+        ArgumentError("This kernel does not support a lazy kronecker kernelmatrix.")
+    )
+end
+
+function kronecker_kernelmatrix(k::MOKernel, x::IsotopicMOInputsUnion)
+    return kronecker_kernelmatrix(k, x, x)
 end
--- a/src/matrix/kernelkroneckermat.jl
+++ b/src/matrix/kernelkroneckermat.jl
@@ -26,10 +26,6 @@ end
 """
 @inline iskroncompatible(κ::Kernel) = false # Default return for kernels
 
-@inline ismatrixkroncompatible(κ::MOKernel) = false # Default return for kernels
-@inline ismatrixkroncompatible(κ::IndependentMOKernel) = true
-@inline ismatrixkroncompatible(κ::IntrinsicCoregionMOKernel) = true
-
 function _kernelmatrix_kroneckerjl_helper(::MOInputIsotopicByFeatures, Kfeatures, Koutputs)
     return Kronecker.kronecker(Kfeatures, Koutputs)
 end
--- a/src/matrix/kernelkroneckermat.jl
+++ b/src/matrix/kernelkroneckermat.jl
@@ -25,3 +25,37 @@ end
     k(x,x') = ∏ᵢᴰ k(xᵢ,x'ᵢ)
 """
 @inline iskroncompatible(κ::Kernel) = false # Default return for kernels
+
+@inline ismatrixkroncompatible(κ::MOKernel) = false # Default return for kernels
+@inline ismatrixkroncompatible(κ::IndependentMOKernel) = true
+@inline ismatrixkroncompatible(κ::IntrinsicCoregionMOKernel) = true
+
+function _mo_kernelmatrix_kron(::MOInputIsotopicByFeatures, K, B)
+    return Kronecker.kronecker(K, B)
+end
+
+function _mo_kernelmatrix_kron(::MOInputIsotopicByOutputs, K, B)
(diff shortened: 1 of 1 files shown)


st--:
> I think conflating this with kernelkronmat would be confusing.

Having looked at `kernelkronmat` (and despaired a bit at its lack of documentation), I agree that conflating the two might be confusing (for one, they have quite different signatures and use-cases). `kernelkronmat` is for when I want to evaluate a (1D) kernel on a multi-dimensional grid. It's actually more limited than it could be, e.g. I could get a kronecker matrix also for any product kernel where each component applies to exactly one of the dimensions...

However, for the MO use-case, the calling is _exactly_ the same as for plain `kernelmatrix` - you just get a more efficient object back. Would there be any reason not to just have `kernelmatrix` return a `Kronecker` object, if Kronecker.jl is loaded? This relates to https://github.com/SebastianAment/CovarianceFunctions.jl/issues/2 as well.

Crown421:
> However, for the MO use-case, the calling is _exactly_ the same as for plain `kernelmatrix` - you just get a more efficient object back. Would there be any reason not to just have `kernelmatrix` return a `Kronecker` object, if Kronecker.jl is loaded?

This was my preferred option as well, I just had some issues with getting the function to overwrite correctly/ thought about making it more explicit to the user.

However, I was able to sort out the code issues that I had, and have now made some changes that clean up the code substantially I think.

src/matrix/kernelkroneckermat.jl
```diff
@@ -25,3 +25,11 @@ end
     k(x,x') = ∏ᵢᴰ k(xᵢ,x'ᵢ)
 """
 @inline iskroncompatible(κ::Kernel) = false # Default return for kernels
+
+function _kernelmatrix_kron_helper(::MOInputIsotopicByFeatures, Kfeatures, Koutputs)
+    return Kronecker.kronecker(Kfeatures, Koutputs)
+end
```
devmotion, line 31: I don't think we can (or should) redefine the method if Kronecker is loaded?

Crown421: It does the right thing (when Kronecker is loaded `kernelmatrix` returns a `KroneckerProduct`), but I am not completely happy with it, especially because there is not enough testing. I would like there to be two tests, one with Kronecker.jl loaded, and one without, including a test to verify that the right kind of object is returned, but not sure how to do that.

devmotion: I would be quite surprised if there are no issues with this approach but I don't know :shrug: If you load KernelFunctions without loading Kronecker, the default fallback method will be defined. But as soon as you load Kronecker as well, it will be redefined. It seems this could lead to quite surprising and inconsistent behaviour. And e.g. maybe it is problematic if you want to build a sysimage?

Crown421: My apologies, I was on vacation. I have thought about it, and while I would prefer to replace the `kernelmatrix` method somehow, I can't come up with a way to do this cleanly with the current state of using `Requires.jl`. 
Maybe once the big issue optional dependencies is officially resolved this can be revisited.

src/mokernels/intrinsiccoregion.jl
```diff
@@ -42,21 +42,33 @@ function (k::IntrinsicCoregionMOKernel)((x, px)::Tuple{Any,Int}, (y, py)::Tuple{
...
 
+function _mo_output_covariance(k::IntrinsicCoregionMOKernel, out_dim)
+    @assert size(k.B) == (out_dim, out_dim)
+    return k.B
+end
+
 function kernelmatrix(
-    k::IntrinsicCoregionMOKernel, x::MOI, y::MOI
-) where {MOI<:IsotopicMOInputsUnion}
+    k::Union{IndependentMOKernel,IntrinsicCoregionMOKernel},
```
willtebbutt, line 51: The only thing that concerns me here is related to code organisation -- in particular it feels weird to define functionality for the `IndependentMOKernel` inside the file for the `ICOMOKernel`.

I think I'd actually rather see more code duplication, but have the code for the `IndependentMOKernel` in the correct file.

willtebbutt: I do prefer the style here though -- it's a definite improvement in terms of readability, so I'm keen to see it updated in the other file.

willtebbutt (commented):
I think this broadly looks good -- would just like some code moved around I think.

src/mokernels/intrinsiccoregion.jl
Crown421: I can sometimes be a bit too much focussed on avoiding duplication, maybe. I have made the change.

src/mokernels/independent.jl
```diff
@@ -27,43 +27,20 @@ function (κ::IndependentMOKernel)((x, px)::Tuple{Any,Int}, (y, py)::Tuple{Any,I
...
-    ) where {MOI<:IsotopicMOInputsUnion}
-        @assert x.out_dim == y.out_dim
-        Ktmp = kernelmatrix(k.kernel, x.x, y.x)
-        mtype = eltype(Ktmp)
-        return _kernelmatrix_kron_helper!(
-            K, x, Ktmp, Matrix{mtype}(I, x.out_dim, x.out_dim)
-        )
-    end
-end
+# for specialized kernelmatrix implementation see intrinsiccoregion.jl
```
willtebbutt, line 43:
```suggestion
```
I guess this comment can go now?

src/mokernels/intrinsiccoregion.jl
```diff
@@ -42,21 +42,33 @@ function (k::IntrinsicCoregionMOKernel)((x, px)::Tuple{Any,Int}, (y, py)::Tuple{
...
+    Koutputs = _mo_output_covariance(k, x.out_dim)
+    return _kernelmatrix_kron_helper(x, Kfeatures, Koutputs)
 end
 
 if VERSION >= v"1.6"
     function kernelmatrix!(
-        K::AbstractMatrix, k::IntrinsicCoregionMOKernel, x::MOI, y::MOI
-    ) where {MOI<:IsotopicMOInputsUnion}
+        K::AbstractMatrix,
+        k::Union{IndependentMOKernel,IntrinsicCoregionMOKernel},
```
willtebbutt, line 64: Sorry, I meant the same thing here really re location of code.

willtebbutt (approved):
I'm happy for this to go in now. Just needs a patch bump, then please feel free to merge and tag when CI passes.

src/mokernels/intrinsiccoregion.jl
Crown421: Sorry, not sure I understand?

willtebbutt: This method also applies to both `IndependentMOKernel` and `ICOMOKernel`

Crown421: Right, sorry, done.

devmotion:
The test errors seem real: https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/364/checks?check_run_id=3675219571#step:6:123 (and the same with the stable Julia version)

I assume this is caused by the hardcoded `Eye{Bool}` (before it was `Eye{eltype(Kfeatures)}`) since booleans are not differentiable in ChainRules (and hence also Zygote).

Crown421:
> The test errors seem real: https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/364/checks?check_run_id=3675219571#step:6:123 (and the same with the stable Julia version)
> 
> I assume this is caused by the hardcoded `Eye{Bool}` (before it was `Eye{eltype(Kfeatures)}`) since booleans are not differentiable in ChainRules (and hence also Zygote).

This is very odd, because previously the tests passed, and I didn't think that I changed anything that would affect this.

willtebbutt:
Looking at the error, I think the optimal fix is probably just to make the pullbacks defined for pairwise of `Delta` [around here](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/b8ae35c9d8079e6127b76f52d35d479fff6e7dc3/src/chainrules.jl#L17) accept any type, rather than just `AbstractMatrix`s.

My reasoning is as follows: based on the CI logs that @devmotion linked, it looks like a `ZeroTangent` is _somehow_ making its way into the `pairwise_pullback`, which means that a gradient has (correctly) been dropped somewhere earlier in the reverse pass. Usually I would expect Zygote to pick up on this and to never call the pullback, but it seems like that's not happening here for some reason and we need to handle the `ZeroTangent` manually.

Crown421:
I just had a look at the successful test from a week (or so) ago (https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/runs/3613631221), and while I am not 100% sure I am looking at the right things I see that Zygote has gone from 0.6.21 to 0.6.22 (this happened 13h ago in fact, https://github.com/FluxML/Zygote.jl/releases). 
One closed issues includes work on adjoints for specialized matrices, which may have changed things for this package?

willtebbutt:
>  One closed issues includes work on adjoints for specialized matrices, which may have changed things for this package?

Yeah, that seems like a likely culprit, specfically https://github.com/FluxML/Zygote.jl/pull/1044 . This is 100% a bug fix though, so we've done that classic thing and been relying on incorrect behaviour in KF.

Tbh, the rules in question should probably all be declared no-ops from an AD perspective using `@non_differentiable`, which will also fix the problem (I suspect that they were written before we had `@non_differentiable` available to us) but widening the set of acceptable cotangnets from `AbstractMatrix` to `Any` will solve the problem for now.

Crown421:
> Tbh, the rules in question should probably all be declared no-ops from an AD perspective using `@non_differentiable`, which will also fix the problem (I suspect that they were written before we had `@non_differentiable` available to us) but widening the set of acceptable cotangnets from `AbstractMatrix` to `Any` will solve the problem for now.

After some bad handling of of git on my part, I have now made some changes, I hope I understood correctly what you suggested.

willtebbutt:
Looks like it's nearly there. Just also needs the same kind of modification [here](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/671f960726b0b9826b9e5a81cc3e55d6c4ba0912/src/chainrules.jl#L49) if I'm interpretting the CI logs correctly.

Crown421:
> Looks like it's nearly there. Just also needs the same kind of modification [here](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/671f960726b0b9826b9e5a81cc3e55d6c4ba0912/src/chainrules.jl#L49) if I'm interpretting the CI logs correctly.

Ok, tests are passing now. I should be ready to merge now I think (I don't have permission to merge).

willtebbutt:
Great. @Crown421 I've invited you to join the org. Please just make sure to have read the first bit of the [ColPrac](https://github.com/SciML/ColPrac) 🙂 if you've not already.

DhairyaLGandhi:
This doesn't great for Zygote flexibility wise. It shouldn't be returning ChainRules types since most rules shouldn't be written to handle them. We should convert these to Zygote friendly types where something like this happens.

willtebbutt:
Sorry @DhairyaLGandhi I don't follow. Which bit in particular are you referring to?

DhairyaLGandhi:
Needing to have adjoints be aware of ChainRules types, I mean.

willtebbutt:
Oh, but we're just writing `rrule`s, so we should _only_ have to worry about ChainRules types, no?

PR #368, opened 2021-09-22
Replace cancel action with use of `concurrency`

devmotion:
This PR copies the approach from ChainRulesCore: https://github.com/JuliaDiff/ChainRulesCore.jl/pull/466


devmotion:
It seems to work.

willtebbutt (approved):
Seems fine to me 🤷

PR #369, opened 2021-09-24
Fix input types, improve readability

Crown421:
**Summary**
As a follow up for the recent #354 pull request this PR changes the input signatures to prevent unintended behaviour in specialized `kernelmatrix` and `kronecker_kernelmatrix` implementations.  

**Proposed changes**
Changing signatures to implicitly enforce same input types (Bug Fix)
Adding a doc string to `kronecker_kernelmatrix`

**What alternatives have you considered?**
None

**Breaking changes**
None

PR #370, opened 2021-09-24
Testing (and fixing) handling of AbstractVector{AbstractVector{T}} inputs

Crown421:
**Summary**
While implementing stuff for #363 I noticed issues with the handling of `AbstractVector{AbstractVector{T}}` as inputs for `kernelmatrix`. Specifically, I noticed a missing method for the `fbm`kernel,  and started by adding tests for `AbstractVector{AbstractVector{T}}` to the test interface. This has spawned additional issues. 

**Proposed changes**
- Adding tests for `AbstractVector{AbstractVector{T}}` inputs
- Possibly fixing uncovered issues with, among others, the `delta` metric, `_mod` in `fbm` kernel. 

I understand that `AbstractVector{AbstractVector{T}}` inputs are not recommended, and that `RowVecs` and `ColVecs` should be used instead. However, in the current state those inputs work for some kernels, but not for others. I see two alternatives 
1. Disallow `AbstractVector{AbstractVector{T}}` inputs, i.e. via `kernelmatrix(k, AbstractVector{AbstractVector{T}}, AbstractVector{AbstractVector{T}}) = error("....., see documentation for details")`
2. Go through with this PR, and add a `maxlog=1` warning that vectors of vectors are not recommended. 

**Breaking changes**
Current state none, alternative 1 will probably be breaking.


willtebbutt:
Vey much in favour of ensuring that this works -- if a kernel says its inputs are vector-valued then `kernelmatrix` ought to work with any `AbstractVector{<:AbstractVector}` of inputs.

At what point should we review this PR properly? It looks like it's got some stuff for vector-of-vectors collections of inputs, but also some fixes for kronecker-structured covariance matrices. Is the intention that they be reviewed together, or done separately?

theogf:
> I understand that AbstractVector{AbstractVector{T}} inputs are not recommended

I don't think this is true at all, these inputs should work just as well!

Crown421:
> At what point should we review this PR properly? It looks like it's got some stuff for vector-of-vectors collections of inputs, but also some fixes for kronecker-structured covariance matrices. Is the intention that they be reviewed together, or done separately?

Apologies for this, all but the last commit are actually part of #369 , but I did not pay attention when branching and don't know enough about git to fix it. #369 should be ready to be reviewed. 

For this commit, given that the expression towards fixing the issues, I will add them here "soon". I have solutions for the issues I found so far, but it is of course possible that more comes up.

willtebbutt:
No problem at all. Lets get 369 sort, then turn our attention back to this PR.

Crown421:
> > I understand that AbstractVector{AbstractVector{T}} inputs are not recommended
> 
> I don't think this is true at all, these inputs should work just as well!

The section that I am referring to is "We recommend that collections of vector-valued inputs are stored in an AbstractMatrix{<:Real} when possible," in the [API section](https://juliagaussianprocesses.github.io/KernelFunctions.jl/stable/api/#Vector-Valued-Inputs)

willtebbutt:
@Crown421 is this the PR that you'd like eyes on next?

willtebbutt (commented):
Just a couple of minor queries -- I think this is broadly fine.

src/test_utils.jl
```diff
@@ -133,6 +133,22 @@ function test_interface(
     )
 end
 
+function test_interface(
+    rng::AbstractRNG,
+    k::Kernel,
+    ::Type{<:AbstractVector{AbstractVector{T}}};
```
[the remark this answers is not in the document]
Crown421: I think,  given the random data generated in the interface `Vector{Vector{T}}` probably makes sense. Then we can add an even more general case (with corresponding data) when data.

src/distances/delta.jl
```diff
@@ -1,19 +1,10 @@
...
-        )
-    end
-    return a == b
-end
-
-Distances.result_type(::Delta, Ta::Type, Tb::Type) = Bool
-
+@inline Distances.eval_op(::Delta, a::Real, b::Real) = a == b
+@inline Distances.eval_reduce(::Delta, a, b) = a && b
+@inline Distances.eval_start(::Delta, a, b) = true
```
devmotion, line 6: This will result in much more complicated code, I don't think we should use the default dispatches in Distances and implement these internal functions.

devmotion: Is there even a reason to implement `Distances._evaluate` here? Isn't `(::Delta)(x, y)` sufficient?

Issue #372, opened 2021-09-27
Implementing non-stationary gibbs kernel


Cyberface:
Hi,

I'm new to Julia and Gaussian processes but recently came across a couple of interesting papers about non-stationary spectral mixture kernels [1], [2]. I can see that you have apready implemented a spectral mixture kernel [here](https://juliagaussianprocesses.github.io/KernelFunctions.jl/stable/kernels/#KernelFunctions.spectral_mixture_kernel) (As a side note I couldn't figure out how to get that to work)

I'm interested in trying to implement this but starting off small so just try to implement one part of this first which would be the input-dependent lengthscale kernel called the Gibbs kernel [3] which can be more easily see in [1] just before equation 7. (Just found another reference also here [4]).

I just took a screenshot from a talk that describes this so you can clearly see the difference between the typical RBF type kernel and this variable length-scale kernel.

![kernel](https://user-images.githubusercontent.com/13164315/134936178-789e278b-287c-4454-9ddc-9cda1c1280fa.png)

Basically if you take the length-scale parameter from a Squared-Exponential Kernel and turn this into a function that depends on the input coordinates then you get a non-stationary kernel called the "Gibbs kernel".

The variable length-scale kernel has a learnable function `\ell(x)` which in [1] they parameterise by another GP.

Can anyone help me implementing this? From reading how KernelFunctions works this looks like it could be done easily using some kind of `Transforms`?

Thanks in advance for any help!

[1]: [Non-Stationary Spectral Kernels](https://arxiv.org/abs/1705.08736)
[2]: [Neural Non-Stationary Spectral Kernel](https://arxiv.org/abs/1811.10978)
[3]: [Gibbs 1997](https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.147.1130&rep=rep1&type=pdf)
[4]: [Nonstationary Covariance Functions for
Gaussian Process Regression (NIPS 2004)](https://proceedings.neurips.cc/paper/2003/file/326a8c055c0d04f5b06544665d8bb3ea-Paper.pdf)

willtebbutt:
Welcome :) I'm keen to see this functionality made available too, so thanks for opening the issue.

[...]

PR #374, opened 2021-09-29
Adding Gibbskernel

Cyberface:
**Summary**

This PR begins to implement the non-stationary Gibbs kernel what is being discussed in issue: #372 

This is a modification of the standard Squared-Exponential kernel where the lengthscale is no longer constant but is a function of the input parameters.

In this PR I have simply added a function `GibbsKernel` to evaluate the gibbs kernel as well as added a test to check that when the lengthscale function just returns a constant value of `1.0` then it should be the same as the `SqExponentialKernel()`.

The implementation of the gibbs kernel is probably going to change as I learn more about how your kernels are defined/implemented.

This is my first PR here and I am only just starting to learn julia so I'm sorry for some basic mistakes!
And I appriciate any comments to make this better thanks!

**Proposed changes**

* Adds a new file "src/kernels/gibbskernel.jl"
* include and export this in "src/KernelFunctions.jl"
* add tests and include the test in "test/runtests.jl"

A better implementation would be to use the kernel/transform infrastructure you have and I will work on this after this PR.

**Breaking changes**

None


src/kernels/gibbskernel.jl
```diff
@@ -0,0 +1,37 @@
...
+    "Neural Non-Stationary Spectral Kernel". arXiv:1811.10978, 2018
+"""
+
+#function (k::GibbsKernel)(x, y, ell)
+function GibbsKernel(x::Any, y::Any, ell::Function)
+    fac = ell(x)^2 + ell(y)^2
+    term1 = 2.0 * ell(x) * ell(y) / fac
+    term2 = - (x-y)^2 / fac
+    return term1 * exp(term2)
+end
```
willtebbutt, line 37:
```suggestion
struct GibbsKernel{T} <: Kernel
    ell::T
end

function (k::GibbsKernel)(x, y)
    fac = k.ell(x)^2 + k.ell(y)^2
    term1 = 2 * k.ell(x) * k.ell(y) / fac
    term2 = -(x-y)^2 / fac
    return term1 * exp(term2)
end
```
I think all of the functionality is here. The main change this suggestion makes is to create a `Kernel` called `GibbsKernel` and define a way to evaluate it. This is needed to ensure that the kernel behaves in the same way as our other kernels.

Minor changes -- these are new-to-Julia things that are more general than this PR:
1. remove `::Any` restrictions on `x` and `y`. `Any` is implicitly assumed if no type constraint is specified.
2. `2.0` -> `2`. This will generally interact better with Julia promotion mechanisms. For example, if `k.ell(x) * k.ell(y) / fac` is a `Float32` then probably what we want is for `term1` to be a `Float32`. Multiplying by `2.0` will produce a `Float64`, because `Float64 * Float32 -> Float64`, while multiplying by `2` will give a `Float32` as desired.

In the tests, you'll now be able to write things like
```julia
k_gibbs = GibbsKernel(ell)
k_gibbs(x, y)
```
etc.

test/kernels/gibbskernel.jl
```diff
@@ -0,0 +1,13 @@
+using Test
+using KernelFunctions
+
```
willtebbutt, line 3:
```suggestion
```
You'll get these in this file via the main `runtests.jl` file.

src/kernels/gibbskernel.jl
```diff
@@ -0,0 +1,37 @@
+"""
+
```
willtebbutt, line 2:
```suggestion
```
Don't think this empty line is needed.

src/kernels/gibbskernel.jl
```diff
@@ -0,0 +1,37 @@
...
+    Functions for Gaussian Process Regression". NEURIPS, 2003
+
+[3] - Sami Remes, Markus Heinonen, Samuel Kaski.
+    "Non-Stationary Spectral Kernels". arXiV:1705.08736, 2017
+
+[4] - Sami Remes, Markus Heinonen, Samuel Kaski.
+    "Neural Non-Stationary Spectral Kernel". arXiv:1811.10978, 2018
+"""
+
+#function (k::GibbsKernel)(x, y, ell)
```
willtebbutt, line 31:
```suggestion
```
I think these lines need to go in order for Julia to understand that the docstring is for `GibbsKernel` -- currently, if you type `?GibbsKernel` at the REPL it doesn't produce the docstring.

src/kernels/gibbskernel.jl
```diff
@@ -0,0 +1,37 @@
...
+
+The Gibbs kernel is non-stationary generalisation of the Squared-Exponential
+kernel. The lengthscale parameter ``\\ell`` becomes a function of
+position ``\\ell(x)``.
+
+``\\ell(x) = 1.`` then you recover the standard Squared-Exponential kernel
+with constant lengthscale.
+
+```math
+k(x, y) = \\sqrt{ \\left( \\frac{2 \\ell(x) \\ell(y)}{\\ell(x)^2 + \\ell(y)^2} \\right) } \\quad \\rm{exp} \\left( - \\frac{(x - y)^2}{\\ell(x)^2 + \\ell(y)^2} \\right)
```
willtebbutt, line 15: Style: might it be possible to get this to under 92 characters by spreading it out on to 2 lines?

willtebbutt (commented):
Thanks for opening this! It's a very nice well-sized PR.

The basic functionality all seems is present. The main thing that needs addressing is turning the `GibbsKernel` `function` into a `struct` (see comment below) -- other than that it's just formatting-related things and other minor details that need addressing.

src/kernels/gibbskernel.jl
theogf: Another small thing:
Assuming that `ell` could be expensive, it might worth precomputing them first and then using them.

willtebbutt: Agreed -- I think this should be done in a follow up PR as an optimisation for `kernelmatrix` etc.

devmotion: I think it would be good to allow to specify a distance measure, as in the case of `SqExponentialKernel`. Maybe one could (should?) also implement `(::GibbsKernel)(x, y)` by constructing a corresponding scaled `SqExponentialKernel` with the correct lengthscale and evaluate it to make use of the optimizations for lengthscales and certain distance measures. In any case, we should not use `(x - y)^2` since it will fail for non-scalar inputs.

willtebbutt: Agreed -- could we fix this in a follow-up PR though? Perhaps we just restrict the input types to `Real` in this PR?

devmotion: It would be a breaking change to add a new field for the distance measure (since it implies an additional type parameter and hence can break dispatches).

willtebbutt: Good point. Could you propose a suggestion that extends my one?

devmotion: Something like
```suggest
struct GibbsKernel{T,M} <: Kernel
    l::T
    metric::M
end

GibbsKernel(; l, metric=Euclidean()) = GibbsKernel(l, metric)

function (k::GibbsKernel)(x, y)
    l = k.l
    lx = l(x)
    ly = l(y)
    lengthscale = (lx^2 + ly^2) / 2
    kernel = (lx * ly / lengthscale) * with_lengthscale(SqExponentialKernel(k.metric), lengthscale)
    return kernel(x, y)
end
```
I think though we should use something more descriptive than `l` or `ell` though.

theogf: I don't think that would be correct though!  The square applied on the `lengthscale` terms comes from the `SqEuclidean` norm. For different norms, I believe `lengthscale` would be different.

devmotion: Maybe, I did not study the original issue or the motivation for this particular choice 😄 It is always a squared exponential kernel though, just with a different metric, so the overall mathematical structure is the same. If necessary, we could make the computation of the variance and lengthscale also dependent on the metric.

theogf: All I am saying it that we should probably not allow for other metrics than `SqEuclidean` for now.

devmotion: We use `Euclidean` as default. The squared metric, scaled with 1/2, is used and fixed in the definition of `SqExponentialKernel`. We always have a squared metric (which I referred to as same mathematical structure), and therefore I am less worried about the use of other metrics.

willtebbutt: Following on from Theo's point, a potential issue with using other metrics is that the Gibbs kernel is defined using some kind of integral against a white noise process (I forget the precise details), so it's possible that the right thing to do under metrics other than the Euclidean has some parametric form other than the one above.

devmotion: I'll read up on the details later today and check if it makes sense to allow to use a general metric. We could of course also just restrict the implementation to `(::GibbsKernel{<:Any,<:Euclidean})(x, y)` for now (this would avoid breaking changes in the future) or remove the `metric` field (if we think we don't want/need it and/or don't care about breaking changes).

In any case, my suggestion above was incorrect, it should be of course something like
```suggestion
struct GibbsKernel{T,M} <: Kernel
    l::T
    metric::M
end

GibbsKernel(; l, metric=Euclidean()) = GibbsKernel(l, metric)

function (k::GibbsKernel)(x, y)
    l = k.l
    lx = l(x)
    ly = l(y)
    lengthscale = hypot(lx, ly)
    kernel = (lx * ly / lengthscale^2) * with_lengthscale(SqExponentialKernel(k.metric), lengthscale)
    return kernel(x, y)
end
```

willtebbutt: > remove the metric field (if we think we don't want/need it and/or don't care about breaking changes).

Hmmm honestly I'm inclined to think that this is the case. If we need to make a breaking change at some point to support different metrics, so be it 🤷 .

I do agree that we ought to ensure that we either restrict the argument type to `Real`s, or generalise the implementation to vectors of reals. I'm inclined to think that former is better for the sake of getting this in quickly, so something like
```suggestion
struct GibbsKernel{T} <: Kernel
    ell::T
end

function (k::GibbsKernel)(x::Real, y::Real)
    fac = k.ell(x)^2 + k.ell(y)^2
    term1 = 2 * k.ell(x) * k.ell(y) / fac
    term2 = -(x-y)^2 / fac
    return term1 * exp(term2)
end
```

devmotion: I really think we shouls use `SqExponentialKernel` already in the first implementation since I think we want to use it at least in the vector case at some point - but then it would be another breaking change to switch to it since it uses a slightly different parameterization (factor 1/2 in the exponential).

I agree, probably it's easiest to not include the `metric` for now. Therefore I'd suggest, with a better name for `l`/`ell`:
```suggestion
struct GibbsKernel{T} <: Kernel
    l::T
end

GibbsKernel(; l) = GibbsKernel(l)

function (k::GibbsKernel)(x, y)
    l = k.l
    lx = l(x)
    ly = l(y)
    lengthscale = hypot(lx, ly)
    kernel = (sqrt(2 * lx * ly) / lengthscale) * with_lengthscale(SqExponentialKernel(), lengthscale)
    return kernel(x, y)
end
```
(BTW it seems `term1` is incorrect in all suggestions and the original implementation and does not correspond to the formula in the issue)

src/kernels/gibbskernel.jl
Cyberface: hopefully I have fixed this - I changed `ell` to `l`

src/kernels/gibbskernel.jl
Cyberface: Thanks for such a lively discussion everyone!

@devmotion I just implemented your suggested and thanks for pointing out the error in `term1`!

I changed your `l` to `lengthscale` to hopefully make it obvious what it is and solve the `ell`/`l` issue.

I'm not entirely confident in my unit test either so will have to double check that too. I think it's because of the `1/2` in the `SqExponentialKernel` but I'll double check.

Nice idea about using the `with_lengthscale`.

Cyberface:
I have updated the unit test - it was a little more confusing and complicated that I originally thought! Hopefully you can follow what I've done through the comments.

Let me know what you think, cheers!

test/kernels/gibbskernel.jl
```diff
@@ -0,0 +1,27 @@
+@testset "gibbskernel" begin
+    x = randn(2)
+    y = randn(2)
+
+    # generate random number of standard SqExponentialKernel lengthscale
+    ell = randn(1)[1]
```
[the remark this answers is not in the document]
Cyberface: ah thanks! Implemented this in https://github.com/Cyberface/KernelFunctions.jl/commit/21de7d45d389f30012242f9c314f1faf1f7699f2

And ah I was doing `randn(1)` and didn't think to just do `randn()` very helpful thank you!

test/kernels/gibbskernel.jl
```diff
@@ -0,0 +1,27 @@
...
+    lengthscale = hypot(l_func(x), l_func(y))
+
+    # create a SqExponentialKernel with a lengthscale
+    k = with_lengthscale(SqExponentialKernel(), lengthscale)
+
+    # create a gibbs kernel with our constant lengthscale function
+    k_gibbs = GibbsKernel(l_func)
+
+    # check they are equal
+    @test k_gibbs(x, y) == k(x, y)
```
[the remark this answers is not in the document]
Cyberface: I'm not really sure what you mean, could you elaborate or sketch out an idea of what you would like to see?

Thanks!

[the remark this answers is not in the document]
theogf: Something like 
```
@test k_gibbs(x, y) == sqrt((ell(x) * ell(y)) / (ell(x)^2 + ell(y)^2)) * exp( - norm(x - y)^2 / (ell(x)^2 + ell(y)^2))
```

test/kernels/gibbskernel.jl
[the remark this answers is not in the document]
theogf: I think you just made a typo and wrote `ell = sum(sin, x)` instead of `ell(x) = sum(sin, x)`

test/kernels/gibbskernel.jl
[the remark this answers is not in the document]
willtebbutt: n.b. probably better to use `≈` here rather than `==`.

test/kernels/gibbskernel.jl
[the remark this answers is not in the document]
Cyberface: I think it's correct, I wanted `ell` to just be a number and not a function right?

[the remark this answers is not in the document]
Cyberface: Oh I see what you mean in your other comment. This test is getting a bit confusing :)

test/kernels/gibbskernel.jl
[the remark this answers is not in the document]
Cyberface: OK I see what you mean.  I've added your suggestion but actually this test is failing and I'm not sure why!

I don't think I fully understand the implementation in https://github.com/Cyberface/KernelFunctions.jl/blob/gibbskernel/src/kernels/gibbskernel.jl#L37 so will take a look a bit later again.

src/kernels/gibbskernel.jl
```diff
@@ -0,0 +1,44 @@
...
+end
+
+GibbsKernel(; lengthscale) = GibbsKernel(lengthscale)
+
+function (k::GibbsKernel)(x, y)
+    lengthscale = k.lengthscale
+    lx = lengthscale(x)
+    ly = lengthscale(y)
+    l = hypot(lx, ly)
+    kernel = (sqrt(2 * lx * ly) / l) * with_lengthscale(SqExponentialKernel(), l)
```
theogf, line 42:
```suggestion
    l = hypot(lx, ly) / sqrt(2)
    kernel = (sqrt(lx * ly) / l) * with_lengthscale(SqExponentialKernel(), l)
```
I think this should more correspond to the exact formula

test/kernels/gibbskernel.jl
```diff
@@ -0,0 +1,32 @@
...
+    # in order to compare to the Gibbs kernel we compute the
+    # equivalent lengthscale l(x)^2 + l(y)^2.
+    # See the denominator of the exponential term in the gibbs kernel.
+    lengthscale = hypot(l_func(x), l_func(y))
+
+    # create a SqExponentialKernel with a lengthscale
+    k = with_lengthscale(SqExponentialKernel(), lengthscale)
+
+    # create a gibbs kernel with our constant lengthscale function
+    k_gibbs = GibbsKernel(l_func)
```
theogf, line 24:
```suggestion
    # this is the gibbs lengthscale function.
    ell(x) = exp(sum(sin, 
    # create a gibbs kernel with our specific lengthscale function
    k_gibbs = GibbsKernel(ell)
```
That's more what I meant to use instead of a constant.

src/kernels/gibbskernel.jl
devmotion: Without promotion to Float64:
```suggestion
    l = invsqrt2 * hypot(lx, ly)
    kernel = (sqrt(lx * ly) / l) * with_lengthscale(SqExponentialKernel(), l)
```

Cyberface: Is this correct though because `l` is passed into `with_lengthscale` - doesn't the `SqExponentialKernel ` already have the `1/2` in the exponential looking [here](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/master/src/basekernels/exponential.jl#L23)

test/kernels/gibbskernel.jl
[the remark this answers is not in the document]
Cyberface: Are you sure the using `norm(x-y)` is correct here? Calling this on 

```
x = randn(2)
y = randn(2)
norm(x-y)
```
This returns a single number - shouldn't this be a 2x2 matrix?

Sorry for my confusion!

src/kernels/gibbskernel.jl
Cyberface: shouldn't it be

```
l = hypot(lx, ly)
kernel = (sqrt(2 * lx * ly) / l) * with_lengthscale(SqExponentialKernel(), 1 / l^2)
```

theogf: > Is this correct though because l is passed into with_lengthscale - doesn't the SqExponentialKernel already have the 1/2 in the exponential looking here

Yes but in the latex formula the factor 2 is not there, so this way we cancel it out!

theogf: @devmotion proposition is definitely the right one!

Cyberface: oh yeh.... thanks!

devmotion: In general, the factor 2 in the exponential does not matter but it is a bit closer to the formulas in the issue 🤷‍♂️

theogf: I guess the 2 disappears because we already have l(x) + l(y) ?

test/kernels/gibbskernel.jl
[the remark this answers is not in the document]
theogf: No no, this the L2 norm! So it's `sqrt(sum((x_i - y_i)^2))`, this is what we would expect I think

src/kernels/gibbskernel.jl
Cyberface: Yeah the 2 would be there when l(x) = l(y) = constant.

I think I finally understand most of the things here but I am confused by `with_lengthscale`. Can you explain why we shouldn't square the argument to with_lengthscale? Looking at the with_lengthscale code it seems to just multiply by `1/l`.

test/kernels/gibbskernel.jl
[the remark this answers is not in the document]
Cyberface: OK I get you now - I thought this was supposed to give the covariance matrix but that is the `kernelmatrix` function!

This is just the covariance between two points in n-dimensions.

Hopefully I got this correct :)

Cyberface:
OK I think this PR is basically done?

src/kernels/gibbskernel.jl
```diff
@@ -0,0 +1,44 @@
...
+[2] - Christopher J. Paciorek and Mark J. Schervish. "Nonstationary Covariance
+    Functions for Gaussian Process Regression". NEURIPS, 2003
+
+[3] - Sami Remes, Markus Heinonen, Samuel Kaski.
+    "Non-Stationary Spectral Kernels". arXiV:1705.08736, 2017
+
+[4] - Sami Remes, Markus Heinonen, Samuel Kaski.
+    "Neural Non-Stationary Spectral Kernel". arXiv:1811.10978, 2018
+"""
+
```
willtebbutt, line 30:
```suggestion
```
Needs to have no space between struct and docstring in order for the REPL and docs to pick up on it.

willtebbutt (approved):
I'm happy -- once @theogf is happy, we can merge :)

src/kernels/gibbskernel.jl
```diff
@@ -0,0 +1,43 @@
+"""
+    GibbsKernel(x, y)
+
+# Definition
+
+The Gibbs kernel is non-stationary generalisation of the Squared-Exponential
+kernel. The lengthscale parameter ``l`` becomes a function of
+position ``l(x)``.
+
+``l(x) = l`` then you recover the standard Squared-Exponential kernel
```
theogf, line 10:
```suggestion
For a constant function``l(x) = c``, one recovers the standard Squared-Exponential kernel with lengthscale `0.5 c`
```

theogf (commented):
Just a comment for the docstring. And could you also add the kernel in the docs (just add the kernel name in the list in docs/kernels.md)?

src/kernels/gibbskernel.jl
Cyberface: great thanks and that answers my other question about the square root / squaring about with_lengthscale

Cyberface:
> Just a comment for the docstring. And could you also add the kernel in the docs (just add the kernel name in the list in docs/kernels.md)?

Yeah sure thing. Shall I add a new section in `base kernels` called 'non-stationary kernels' ? Or shall I put it in 'Exponential Kernels' ?

willtebbutt:
I'd just put it in alphabetical order in the base kernels list. Feels like the appropriate place to me (the base kernels aren't necessarily stationary)

Cyberface:
great thanks - done!

src/kernels/gibbskernel.jl
devmotion: > I guess the 2 disappears because we already have l(x) + l(y) ?

I'm not sure what 2 you refer to (and there is no l(x) + l(y) term anywhere 🤔). The 2 in the square root outside of the squared exponential kernel disappears because we already scale the square root of the denominator by 1/sqrt(2).

> why we shouldn't square the argument to `with_lengthscale`

It will be squared by the squared exponential kernel. A squared exponential kernel with lengthscale l is of the form `k(x, y) = exp(- (x - y)^2/(2*l^2))`.

Cyberface: So the `l(x) + l(y)` term I was refering to is the denominator in the exponential i.e.
`exp(-(x - y)^2 / (l(x)^2 + l(y)^2)`

And when `l(x) = l(y) = l = constant` you get `exp(-(x - y)^2 / (2l^2)` ?

Also thanks for explaining to me about the `with_lengthscale`

src/kernels/gibbskernel.jl
devmotion: The lengthscale is just `c`, it seems: In this case we get `exp(-(x-y)^2/(2*c^2))`, both with our implementation and the formula it is based on which corresponds to a lengthscale of `c`.

src/kernels/gibbskernel.jl
```diff
@@ -0,0 +1,43 @@
+"""
+    GibbsKernel(x, y)
+
+# Definition
+
+The Gibbs kernel is non-stationary generalisation of the Squared-Exponential
+kernel. The lengthscale parameter ``l`` becomes a function of
+position ``l(x)``.
+
+For a constant function``l(x) = c``, one recovers the standard Squared-Exponential kernel with lengthscale `0.5 c`
```
devmotion, line 10: See my comment above:
```suggestion
For a constant function``l(x) = c``, one recovers the standard squared exponential kernel with lengthscale `c`.
```

devmotion (changes requested):

test/kernels/gibbskernel.jl
```diff
@@ -0,0 +1,12 @@
+@testset "gibbskernel" begin
+    x = randn()
+    y = randn()
+
+    # this is the gibbs lengthscale function.
+    ell(x) = exp(sum(sin, x))
+    # create a gibbs kernel with our specific lengthscale function
+    k_gibbs = GibbsKernel(ell)
+
+    @test k_gibbs(x, y) ≈ sqrt((2 * ell(x) * ell(y)) / (ell(x)^2 + ell(y)^2)) * exp(- norm(x - y)^2 / (ell(x)^2 + ell(y)^2))
```
devmotion, line 10: No need for `norm` here:
```suggestion
    @test k_gibbs(x, y) ≈ sqrt((2 * ell(x) * ell(y)) / (ell(x)^2 + ell(y)^2)) * exp(-(x - y)^2 / (ell(x)^2 + ell(y)^2))
```

Cyberface: Ah ok - i suppose if we used something like `x=rand(2)` or something then we should use `norm` ?

src/kernels/gibbskernel.jl
Cyberface: ah ok - thanks for thinking about this and figuring it out, it is a bit confusing :)

test/kernels/gibbskernel.jl
devmotion: Yes, if `x` and `y` would be vectors it would be needed.

src/kernels/gibbskernel.jl
theogf: But no that's not true :) That's why we needed to divide by `sqrt(2)`

src/kernels/gibbskernel.jl
devmotion: Exactly, if they are constant you get `exp(-(x-y)^2 / (2 * l^2))`. Very explicitly, in general the implementation yields
```math
k(x, y) = sqrt(l(x) * l(y)) / (sqrt(l(x)^2 + l(y)^2) / sqrt(2)) * exp(- (x - y)^2 / (2 * (sqrt(l(x)^2 + l(y)^2) / sqrt(2))^2))
= sqrt(2 * l(x) * l(y) / (l(x)^2 + l(y)^2)) * exp(- (x - y)^2 / (l(x)^2 + l(y)^2))
```
which is exactly the formula in the docstring (and the other issue).

Cyberface:
The PR says there is a request to change something but I'm not sure what it's referring to.

devmotion:
> The PR says there is a request to change something but I'm not sure what it's referring to.

It was my request to fix the docstring.

test/kernels/gibbskernel.jl
```diff
@@ -0,0 +1,14 @@
...
+    # this is the gibbs lengthscale function.
+    ell(x) = exp(sum(sin, x))
+    # create a gibbs kernel with our specific lengthscale function
+    k_gibbs = GibbsKernel(ell)
+
+    @test k_gibbs(x, y) ≈
+          sqrt((2 * ell(x) * ell(y)) / (ell(x)^2 + ell(y)^2)) *
+          exp(-(x - y)^2 / (ell(x)^2 + ell(y)^2))
+
+end
```
devmotion, line 14: Seems JuliaFormatter is not completely happy:
```suggestion
end
```

src/kernels/gibbskernel.jl
```diff
@@ -0,0 +1,43 @@
+"""
+    GibbsKernel(x, y)
```
devmotion, line 2:
```suggestion
    GibbsKernel(; lengthscale)
```

src/kernels/gibbskernel.jl
```diff
@@ -0,0 +1,43 @@
...
+
+# Definition
+
+The Gibbs kernel is non-stationary generalisation of the Squared-Exponential
```
devmotion, line 6:
```suggestion
The Gibbs kernel is non-stationary generalisation of the squared exponential
```

src/kernels/gibbskernel.jl
```diff
@@ -0,0 +1,43 @@
...
+kernel. The lengthscale parameter ``l`` becomes a function of
+position ``l(x)``.
+
+For a constant function``l(x) = c``, one recovers the standard squared exponential kernel with lengthscale `c`.
+with constant lengthscale ``l``.
```
devmotion, line 11:
```suggestion
For a constant function``l(x) = c``, one recovers the standard squared exponential kernel
with lengthscale `c`.
```

src/kernels/gibbskernel.jl
```diff
@@ -0,0 +1,43 @@
...
+
+```math
+k(x, y) = \\sqrt{ \\left(\\frac{2 l(x) l(y)}{l(x)^2 + l(y)^2} \\right) }
```
devmotion, line 14:
```suggestion
k(x, y; l) = \\sqrt{ \\left(\\frac{2 l(x) l(y)}{l(x)^2 + l(y)^2} \\right) }
```

src/kernels/gibbskernel.jl
```diff
@@ -0,0 +1,43 @@
...
+\\quad \\rm{exp} \\left( - \\frac{(x - y)^2}{l(x)^2 + l(y)^2} \\right)
+```
+
```
devmotion, line 17:
```suggestion

# References

```

src/kernels/gibbskernel.jl
```diff
@@ -0,0 +1,43 @@
...
+    PhD thesis, 1997
+
+[2] - Christopher J. Paciorek and Mark J. Schervish. "Nonstationary Covariance
+    Functions for Gaussian Process Regression". NEURIPS, 2003
+
+[3] - Sami Remes, Markus Heinonen, Samuel Kaski.
+    "Non-Stationary Spectral Kernels". arXiV:1705.08736, 2017
+
+[4] - Sami Remes, Markus Heinonen, Samuel Kaski.
+    "Neural Non-Stationary Spectral Kernel". arXiv:1811.10978, 2018
```
devmotion, line 28:
```suggestion
Mark N. Gibbs. "Bayesian Gaussian Processes for Regression and Classication." PhD thesis, 1997

Christopher J. Paciorek and Mark J. Schervish. "Nonstationary Covariance Functions
for Gaussian Process Regression". NeurIPS, 2003

Sami Remes, Markus Heinonen, Samuel Kaski. "Non-Stationary Spectral Kernels". arXiV:1705.08736, 2017

Sami Remes, Markus Heinonen, Samuel Kaski. "Neural Non-Stationary Spectral Kernel". arXiv:1811.10978, 2018
```

devmotion (commented):
Some minor improvements to the docstring.

devmotion (approved):

theogf:
Great! All we need now is a version bump. In `Project.toml` can you change `version = "0.10.20"` to `version = "0.10.21"`

devmotion:
Thanks a lot @Cyberface!

Cyberface:
that's really great! Thanks so much for all your help with this! Next I'll work on getting the kernelmatrix working and eventually getting it so that the gibbs kernel length scale can be modelled with a separate GP and fit/optimise that!

PR #375, opened 2021-10-01
add gibbskernel example to prior example script

Cyberface:
**Summary**

Adding an example how to use the Gibbs Kernel in the prior example.
The Gibbs kernel has a lengthscale *function* as it's parameter.
I just chose a simple function for demonstration purposes i.e. `l(x) = exp(sin(x))`.

**Proposed changes**

* Adding an example to `examples/gaussian-process-priors/script.jl`

Could make a separate example and will probably do that when I have a working version of the Gibbs kernel that also learns the lengthscale function.

**Breaking changes**

None

### DIFF
--- a/examples/gaussian-process-priors/script.jl
+++ b/examples/gaussian-process-priors/script.jl
@@ -120,6 +120,7 @@ kernels = [
     LinearKernel(),
     compose(PeriodicKernel(), ScaleTransform(0.2)),
     NeuralNetworkKernel(),
+    GibbsKernel(x -> exp(sin.(x))),
 ]
 plot(
     [visualize(k) for k in kernels]...;


examples/gaussian-process-priors/script.jl
```diff
@@ -120,6 +120,7 @@ kernels = [
     LinearKernel(),
     compose(PeriodicKernel(), ScaleTransform(0.2)),
     NeuralNetworkKernel(),
+    GibbsKernel(x -> exp(sin.(x))),
```
devmotion, line 123: Usually we prefer the keyword argument constructor since it is a bit clearer. You can also simplify the lengthscale function a bit (broadcasting is only useful if `x` are vectors - in which case `exp(sin.(x))` would fail though):
```suggestion
    GibbsKernel(; lengthscale=exp ∘ sin),
```

theogf: I think we should generally handle vectors so something like `exp(sum(sin, x))` maybe?

devmotion: It seemed the example only deals with scalar inputs but maybe I didn't check it carefully enough. If vectors occur as inputs, then I agree something like `x -> exp(sum(sin, x))` or `x -> sum(exp ∘ sin, x)` should be used.

Cyberface: Thanks both for the helpful information and suggestions. I went with `x -> sum(exp ∘ sin, x)` and checked locally that it works in a notebook so hopefully the CI will be happy.

PR #376, opened 2021-10-01
Implement Zygote hack

willtebbutt:
**Summary**

Zygote has type-stability issues associated with getfield and getproperty. Hopefully we'll get a solution at some point (see e.g. [this PR](https://github.com/FluxML/Zygote.jl/pull/909/files)), but in the mean time, avoiding the`getproperty` call here occassionally removes a type-stability problem.

**Proposed changes**

Write `Distances_pairwise` instead of `Distances.pairwise`.

Ideally we wouldn't have our own version of this function, so I thought about trying to solve that properly. Seemed like this proposed solution is easiest, doesn't break anything, and works now.

**Breaking changes**

None

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -80,8 +80,6 @@ Base.vcat(a::ColVecs, b::ColVecs) = ColVecs(hcat(a.X, b.X))
 
 dim(x::ColVecs) = size(x.X, 1)
 
-const Distances_pairwise = Distances.pairwise
-
 pairwise(d::PreMetric, x::ColVecs) = Distances_pairwise(d, x.X; dims=2)
 pairwise(d::PreMetric, x::ColVecs, y::ColVecs) = Distances_pairwise(d, x.X, y.X; dims=2)
 function pairwise(d::PreMetric, x::AbstractVector, y::ColVecs)
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -80,13 +80,15 @@ Base.vcat(a::ColVecs, b::ColVecs) = ColVecs(hcat(a.X, b.X))
 
 dim(x::ColVecs) = size(x.X, 1)
 
-pairwise(d::PreMetric, x::ColVecs) = Distances.pairwise(d, x.X; dims=2)
-pairwise(d::PreMetric, x::ColVecs, y::ColVecs) = Distances.pairwise(d, x.X, y.X; dims=2)
+const Distances_pairwise = Distances.pairwise
+
+pairwise(d::PreMetric, x::ColVecs) = Distances_pairwise(d, x.X; dims=2)
+pairwise(d::PreMetric, x::ColVecs, y::ColVecs) = Distances_pairwise(d, x.X, y.X; dims=2)
 function pairwise(d::PreMetric, x::AbstractVector, y::ColVecs)
-    return Distances.pairwise(d, reduce(hcat, x), y.X; dims=2)
+    return Distances_pairwise(d, reduce(hcat, x), y.X; dims=2)
 end
 function pairwise(d::PreMetric, x::ColVecs, y::AbstractVector)
-    return Distances.pairwise(d, x.X, reduce(hcat, y); dims=2)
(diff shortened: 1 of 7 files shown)


theogf:
Can't we just export pairwise directly?

willtebbutt:
> Can't we just export pairwise directly?

Sorry, I don't follow. What would that do?

src/utils.jl
```diff
@@ -80,13 +80,15 @@ Base.vcat(a::ColVecs, b::ColVecs) = ColVecs(hcat(a.X, b.X))
 
 dim(x::ColVecs) = size(x.X, 1)
 
-pairwise(d::PreMetric, x::ColVecs) = Distances.pairwise(d, x.X; dims=2)
-pairwise(d::PreMetric, x::ColVecs, y::ColVecs) = Distances.pairwise(d, x.X, y.X; dims=2)
+const Distances_pairwise = Distances.pairwise
+
```
devmotion, line 84: I guess this should be removed since it is already defined in src/KernelFunctions.jl?
```suggestion
```

willtebbutt: Sorry, yes, that was an oversight.

devmotion (approved):
A bit annoying but we already have some Zygote workarounds in the package :man_shrugging: Seems OK to me, just update the version number before merging it. If it is possible to test this change (e.g. inference of a Zygote pullback?) it would be good to add one.

willtebbutt:
I've added some extra tests, and also improved type-stability for `getproperty(::ColVecs)`, which was needed in order to test the original chance in a practical setting. On the bright side, `pairwise(::SqEuclidean(), ::ColVecs)` is now type-stable under Zygote 🤷

test/test_utils.jl
```diff
@@ -31,6 +31,17 @@ end
 
 # AD utilities
 
+# Type to work around some performance issues that can happen on the reverse-pass of Zygote.
+using Zygote: @adjoint, accum, AContext
+
```
willtebbutt, line 36:
```suggestion
```

test/utils.jl
```diff
@@ -51,6 +51,21 @@
...
             @test back(ones(size(X)))[1].X == ones(size(X))
         end
+
+        @testset "Zygote type-inference" begin
+            ctx = NoContext()
+            x = ColVecs(randn(2, 4))
+            y = ColVecs(randn(2, 3))
+
+            # Why on earth I need to write KernelFunctions.pairwise, rather than just
+            # pairwise, is beyond me. Probably something to do with globals or something.
```
devmotion, line 61: `pairwise` here refers to `Distances.pairwise`, so it is not the same function as `KernelFunctions.pairwise`.

devmotion (commented):
It became even uglier 😂 I'm still fine with it in general. Maybe my comment explains some of the weirdness?

test/utils.jl
willtebbutt: Ohhhhhh okay. I'll tweak the comment.

willtebbutt, line 61:
```suggestion
            # Ensure KernelFunctions.pairwise rather than Distances.pairwise is used.
```

willtebbutt:
Doesn't appear to work on 1.3, so have disabled the test on that version.

devmotion:
IMO it is fine if Zygote's type inference fails with older Julia versions.

src/zygoterules.jl
```diff
@@ -5,3 +5,9 @@ end
 ZygoteRules.@adjoint function Base.map(t::Transform, X::RowVecs)
     return ZygoteRules.pullback(_map, t, X)
 end
+
+function ZygoteRules._pullback(
+    cx::AContext, ::typeof(literal_getproperty), x::ColVecs, ::Val{f}
```
devmotion, line 10: I assume we need the same for `RowVecs`?

willtebbutt: Indeed. I've got [this PR](https://github.com/FluxML/Zygote.jl/pull/1091) going through at the minute though, so I'll start adding stuff once that's in.

Issue #299, opened 2021-06-10
Remove hard-coded `Vector` fields and add vectors constructors


theogf:
Right now kernels like `LinearKernel` have two problems: they can only take `Real` arguments (no way to pass `kernel.c`) as an argument and also they do not allow for different `AbstractVector` types to be stored.
The interested use case is when using GPUs, one cannot do `kernelmatrix` on `CuArrays` without this.

```julia
struct LinearKernel{Tc<:Real} <: SimpleKernel
    c::Vector{Tc}

    function LinearKernel(c::Real)
        @check_args(LinearKernel, c, c >= zero(c), "c ≥ 0")
        return new{typeof(c)}([c])
    end
end
```
should be
```julia
struct LinearKernel{Tc<:Real,Vc<:AbstractVector{<:Tc}} <: SimpleKernel
    c::V
    function LinearKernel(c::V) where {T<:Real, V<:AbstractVector{T}}
        @check_args(LinearKernel, first(c), first(c) >= zero(c), "c ≥ 0")
        return new{T,V}(c)
    end
    function LinearKernel(c::Real)
        @check_args(LinearKernel, c, c >= zero(c), "c ≥ 0")
        C = [c]
        return new{eltype(C),typeof(C)}(C)
    end
end
```

devmotion:
Or we could just not make it a vector at all but a scalar :stuck_out_tongue:

devmotion:
If we stick with vectors (e.g. because of Flux/Zygote), in the example I guess `LinearKernel(c::Real)` could be a simple outer constructor `LinearKernel(c::Real) = LinearKernel([c])`.

theogf:
But then how do I do my inplace updates! :laughing:. Also nobody like scalars! Especially GPUs!

devmotion:
One could eg use ParameterHandling and perform in-place updates of the resulting vector 😉 But then you would still have to go back and reconstruct the kernel.

willtebbutt:
Did we decide that `Ref`s were a bad idea? i.e. we shouldn't be using `Ref(5.0)` doesn't play nicely with Functors.jl?

theogf:
`Ref` is not captured by `Flux.params` AFAICT

devmotion:
I still think one could consider dropping support for Flux.params and switching to scalar parameters instead. Both Ref and Vector seem a bit unnatural.

willtebbutt:
[...]

PR #396, opened 2021-11-02
Add `InverseMultiquadricKernel`

devmotion:
This PR adds the inverse multiquadric kernel
```math
k(x, x'; alpha, c) = (c + d(x, x')^2)^(- alpha)
```
with parameters `alpha, c > 0`.

In recent years, in particular in the ML community (possibly due to https://arxiv.org/abs/1711.01558) this kernel or some special cases (e.g. with `alpha = 1/2` and `c = 1`: http://proceedings.mlr.press/v119/delbridge20a/delbridge20a.pdf) are called "inverse multiquadratic kernel". As far as I know, however, the traditional name is "inverse multiquadric kernel" (see e.g. Schölkopf and Smola's "Learning with Kernels").

### DIFF
--- a/src/basekernels/rational.jl
+++ b/src/basekernels/rational.jl
@@ -139,3 +139,71 @@ function Base.show(io::IO, κ::GammaRationalKernel)
         ")",
     )
 end
+
+@doc raw"""
+    InverseMultiQuadricKernel(; α::Real=1.0, c::Real=1.0, metric=Euclidean())
+
+Inverse multiquadric kernel with respect to the `metric` with parameters `α` and `c`.
+
+# Definition
+
+For inputs ``x, x'`` and metric ``d(\cdot, \cdot)``, the inverse multiquadric kernel with
+parameters ``\alpha, c > 0`` is defined as
+```math
+k(x, x'; \alpha, c) = \big(c + d(x, x')^2\big)^{-\alpha}.
+```
+By default, ``d`` is the Euclidean metric ``d(x, x') = \|x - x'\|_2``.
+
+For ``\alpha = c = 1``, the [`GammaRationalKernel`](@ref) with parameters ``\alpha = 1``
+and ``\gamma = 2`` is recovered.
+
+For ``\alpha = 1/2`` and ``c = 1``, the [`RationalQuadraticKernel`](@ref) with parameter
+``\alpha = 1/2`` is recovered.
+
+# References
+
+Micchelli, C.A. (1986). Interpolation of scattered data: Distance matrices and conditionally
+positive definite functions. Constructive Approximation 2, 11-22.
+"""
+struct InverseMultiQuadricKernel{Tα<:Real,Tc<:Real,M} <: SimpleKernel
+    α::Vector{Tα}
+    c::Vector{Tc}
+    metric::M
+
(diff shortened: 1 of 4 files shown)


src/basekernels/rational.jl
```diff
@@ -139,3 +139,71 @@ function Base.show(io::IO, κ::GammaRationalKernel)
         ")",
     )
 end
+
+@doc raw"""
+    InverseMultiQuadricKernel(; α::Real=1.0, c::Real=1.0, metric=Euclidean())
+
+Inverse multiquadric kernel with respect to the `metric` with parameters `α` and `c`.
```
theogf, line 146: I don't think this sentence is needed, it does not bring anything

src/basekernels/rational.jl
```diff
@@ -139,3 +139,71 @@ function Base.show(io::IO, κ::GammaRationalKernel)
...
+
+# Definition
+
+For inputs ``x, x'`` and metric ``d(\cdot, \cdot)``, the inverse multiquadric kernel with
+parameters ``\alpha, c > 0`` is defined as
+```math
+k(x, x'; \alpha, c) = \big(c + d(x, x')^2\big)^{-\alpha}.
```
theogf, line 153: Looking at this, it looks extremely similar to the polynomial kernel (except for -alpha < 0). Could this eventually be unified?

src/basekernels/rational.jl
```diff
@@ -139,3 +139,71 @@ function Base.show(io::IO, κ::GammaRationalKernel)
...
+For ``\alpha = 1/2`` and ``c = 1``, the [`RationalQuadraticKernel`](@ref) with parameter
+``\alpha = 1/2`` is recovered.
+
+# References
+
+Micchelli, C.A. (1986). Interpolation of scattered data: Distance matrices and conditionally
+positive definite functions. Constructive Approximation 2, 11-22.
+"""
+struct InverseMultiQuadricKernel{Tα<:Real,Tc<:Real,M} <: SimpleKernel
+    α::Vector{Tα}
```
theogf, line 169: Although there will be the solution from #299 coming, we should avoid hardcoded `Vector`s no?

src/basekernels/rational.jl
devmotion: Yes, I guess one has to know at least what the inverse multiquadric kernel is. However, the same problem exists in all other docstrings. I added this sentence only to be consistent with them.

src/basekernels/rational.jl
devmotion: This the current way in which all kernels are implemented🤷‍♂️ IMO we should be consistent and use `Vector` for now (the current test setup does not work with scalar parameters) - and hopefully replace all of them with scalars soon.

src/basekernels/rational.jl
devmotion: IMO it is more similar to the rational quadratic kernel, therefore I put it in this file.

devmotion: As the tests show it is the same as scaling an RQ kernel with rescaled inputs. The construction is a bit annoying though, so the question is more: should it be a separate kernel or a function that constructs the corresponding RQ kernel?

PR #398, opened 2021-11-04
Add Benchmarking CI

theogf:
**Summary**

Following #387 (and #2) We need some benchmarking to see the effect of our changes!
Using [BenchmarkCI.jl](https://github.com/tkf/BenchmarkCI.jl) we should be able to trigger a Benchmark CI when the PR is tagged with `performance critical`.
What BenchmarkCI does is to run tests both on `master` and the current `commit` and produces a report.
This PR is also to try things and see how they run.

This PR also introduce a very short benchmark suite to test things are working. This can be extended in a future PR.

None

**Breaking changes**

Nope


theogf:
Aaaaand I did not branch from master once again....

PR #400, opened 2021-11-04
Allows to trigger on labelling and clean-up the process

theogf:
Allows to trigger benchmarking on labelling, will cancel previous running jobs and will always reuse the same comment to upload the benchmarks

The in place workflow is from https://github.com/tkf/BenchmarkCI.jl/issues/91
Note that the proposed method in the posted issue is not relevant anymore and the workflow used corresponds to what is currently in master for JET.jl


theogf:
[benchmark]

theogf:
[benchmark]

theogf:
[benchmark]

devmotion:
> when writing [benchmark] in a comment, once labeled the PR will also run at every change

So the main motivation for the PR is to be able to rerun the benchmarks automatically whenever a new commit is pushed? Couldn't this be achieved by checking the labels as well? I also wonder if it's good to rerun benchmarks automatically. I would have thought a natural workflow would be to only trigger it manually, eg before/when a PR is reviewed or merged. Basically similar to Nanosoldier in base.

theogf:
This is what this PR is supposed to achieve, both continuous CI via the label, and ponctual trigger via comment. The coma before "once labeled" should be a full stop

theogf:
[benchmark]

theogf:
[benchmark]

theogf:
That's again one of these things that will only work once merged in `master`...

.github/workflows/benchmark.yml
```diff
@@ -2,23 +2,58 @@ name: Run benchmarks
...
+              judgement = BenchmarkCI._loadjudge(BenchmarkCI.DEFAULT_WORKSPACE)
+              title = "Kernel Benchmark Result"
+              ciresult = BenchmarkCI.CIResult(; judgement, title)
+              BenchmarkCI.printcommentmd(stdout::IO, ciresult)
+          end
+          ')
+          body="${body//'%'/'%25'}"
+          body="${body//$'\n'/'%0A'}"
+          body="${body//$'\r'/'%0D'}"
+          echo $body > ./benchmark-result.artifact
```
devmotion, line 50: Maybe simpler if you use a Julia script?
```suggestion
          using BenchmarkCI
          judgement = BenchmarkCI._loadjudge(BenchmarkCI.DEFAULT_WORKSPACE)
          title = "Kernel Benchmark Result"
          ciresult = BenchmarkCI.CIResult(; judgement, title)
          comment = sprint() do io
              return BenchmarkCI.printcommentmd(io, ciresult)
          end
          comment = replace(comment, "%" => "%25", "\\n" => "%0A", "\\r" => "%0D")
          write("benchmark-result.artifact", comment)
        shell: julia --color=yes {0}
```
Instead of the manual replacements maybe you could also use `Markdown.htmlesc`.

.github/workflows/benchmark.yml
```diff
@@ -2,23 +2,58 @@ name: Run benchmarks
...
+      # record the pull request number
+      - name: record pull request number
+        run: echo ${{ github.event.pull_request.number }} > ./pull-request-number.artifact
+
```
devmotion, line 54:
```suggestion
```

.github/workflows/benchmark-comment.yml
```diff
@@ -0,0 +1,64 @@
...
+          echo ::set-output name=body::$(cat ./pull-request-number.artifact)
+      # check if the previous comment exists
+      - name: find comment
+        uses: peter-evans/find-comment@v1
+        id: fc
+        with:
+          issue-number: ${{ steps.output-pull-request-number.outputs.body }}
+          comment-author: 'github-actions[bot]'
+          body-includes: Kernel Benchmark Result
+
```
devmotion, line 51:
```suggestion
```

.github/workflows/benchmark-comment.yml
```diff
@@ -0,0 +1,64 @@
...
+jobs:
+  comment:
+    runs-on: ubuntu-latest
+    if: >
+      ${{ github.event.workflow_run.event == 'pull_request' &&
+          github.event.workflow_run.conclusion == 'success' }}
+    steps:
+      - uses: actions/checkout@v2
+
```
devmotion, line 28:
```suggestion
```

devmotion (commented):
The general setup seems reasonable but of course it's difficult to review properly if we can't run the action. Maybe we just have to iterate and fix issues as they appear.

devmotion (approved):
OK, let's see if it works :slightly_smiling_face:

Issue #389, opened 2021-10-23
`PeriodicKernel` does not work with AD (see issue for work-around)


david-vicente:
Hi.
I noticed that the PeriodicKernel is failling AD tests and I would like to implement the Mauna Loa GPR example. I tried looking for discussion regarding this issue in past Issues but there isn't much. Is the problem an instance of "it is doable but no one has had the time yet" or is it pending because of stuff that depends on AD packages?
Thanks!

st--:
Oh, I was independently trying to build the MaunaLoa example (I put the work-in-progress here: https://github.com/JuliaGaussianProcesses/EcosystemIntegrationTesting/blob/main/api_testing/mauna_loa_example/script.jl as a Pluto.jl notebook) and ran into the same bug.

Here's code to reproduce the bug:
```julia
using AbstractGPs
using ParameterHandling
function build_gp(th)
    return GP(th.s^2 * with_lengthscale(PeriodicKernel(; r=[th.l/2]), th.p))
end
th = (; s = positive(exp(0.0)), l=positive(exp(1.0)), p=positive(exp(0.0)))
thflat, unflatten = ParameterHandling.value_flatten(th)
x = rand(5)
y = rand(5)
function loss(th)
    f=build_gp(th)
    return -logpdf(f(x, 0.01), y)
end
loss_packed = loss ∘ unflatten
using Zygote
Zygote.gradient(loss_packed, thflat)
```

st--:
@david-vicente the following two lines give equivalent kernels (in 1D):
```julia
k1 = with_lengthscale(PeriodicKernel(;r=[wiggle_scale / 2]), period)
k2 = with_lengthscale(SqExponentialKernel(), wiggle_scale) ∘ PeriodicTransform(1/period)
```
and we might want to provide `with_period(kernel, period) = kernel ∘ PeriodicTransform(1/period)`.
AutoDiff works fine for the PeriodicTransform. Using the second form, you can also swap out the base kernel, e.g. for a Matern32Kernel:)

(Note that `wiggle_scale` is relative to a period, not relative to the actual input scale!)

david-vicente:
Thank you @st-- , I did look to the `PeriodicTransform` at the time, but it wasn't obvious how to use it to achieve equivalent kernels from the documentation. Maybe we should add
```
k1 = with_lengthscale(PeriodicKernel(; r=[wiggle_scale / 2]), period)
[...]

PR #401, opened 2021-11-06
Add `with_period` convenience function and a example to `PeriodicTransform` docstring

david-vicente:
This PR follows the suggestion [in this comment](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/389#issuecomment-961956874
)
As it is  `with_period` only works with scalars, which contrasts with `with_lengthscale` that uses `ARDTransform` to deal with vector inputs. Any ideas on how should we deal with this?

### DIFF
--- a/src/transform/periodic_transform.jl
+++ b/src/transform/periodic_transform.jl
@@ -14,6 +14,21 @@ julia> f = rand(); t = PeriodicTransform(f); x = rand();
 julia> t(x) == [sinpi(2 * f * x), cospi(2 * f * x)]
 true
 ```
+
+For 1 dimensional inputs it is possible to create a kernel equivalent
+to the `PeriodicKernel` using `PeriodicTransform` and a
+`SqExponentialKernel`.
+
+```jldoctest
+julia> wiggle_scale = 0.5; period = π/2; x = rand(); y = rand();
+
+julia> k1 = with_lengthscale(PeriodicKernel(; r=[wiggle_scale / 2]), period);
+
+julia> k2 = with_lengthscale(SqExponentialKernel(), wiggle_scale) ∘ PeriodicTransform(1/period)
+
+julia> k1(x,y) ≈ k2(x,y)
+true
+```
 """
 struct PeriodicTransform{Tf<:AbstractVector{<:Real}} <: Transform
     f::Tf
--- a/src/transform/with_period.jl
+++ b/src/transform/with_period.jl
@@ -1,0 +1,21 @@
+"""
+    with_period(kernel::Kernel, period::Real)
+
+Construct a transformed kernel with `period`.
+
+# Examples
+
+```jldoctest
+julia> kernel = with_period(SqExponentialKernel(), π/2);
+
+julia> x = rand();
+
+julia> y = rand();
+
+julia> kernel(x, y) ≈ (SqExponentialKernel() ∘ PeriodicTransform(2/π))(x, y)
+true
+```
+"""
+function with_period(kernel::Kernel, period::Real)
+    return kernel ∘ PeriodicTransform(inv(period))
+end


devmotion (commented):
One of the motivations of `with_lengthscale` is that it provides a unified API for `ScaleTransform` and `ARDTransform` (although one can save allocations by using `ARDTransform` directly). This is not the case here since it will always just add a `PeriodTransform`. Usually, users are expected to use the transforms, eg there's also no `with_linear` or `with_function` as one can just use `LinearTransform` and `FunctionTransform`. Therefore maybe we should not add an additional function to the API but rather
- tell users (how) to use `PeriodTransform` directly
- add a convenience constructor `PeriodKernel(r::Real)`

david-vicente:
> One of the motivations of `with_lengthscale` is that it provides a unified API for `ScaleTransform` and `ARDTransform` (although one can save allocations by using `ARDTransform` directly).

I thought that `with_lengthscale` was added because the convention when dealing with kernels is dividing the lengscale and `ScaleTransform` is a multiplication. The same happens with the period/frequency here. But I agree that adding a lot of different ways to achieve the same things ends up polluting the API.

st--:
My motivation for introducing `with_lengthscale` was also to not have to think in terms of inverse lengthscale.
And when constructing a periodic kernel, I'd rather just write `with_period(kernel, period)` than `kernel \circ<tab> PeriodicTransform(1/period)`.

> add a convenience constructor `PeriodKernel(r::Real)`

I'm not quite sure what you mean by that, and how it's different from `with_period`? how would you pass in the base kernel?

What should the meaning of periodicity be in more than one dimension? Coordinate-wise (i.e., equivalent to a product of kernels operating on one of the dimensions each)?

devmotion:
So it seems the main motivation is to introduce a different parametrization of `PeriodicTransform`? Then we could just define
```julia
function PeriodicTransform(; period=nothing, frequency=period === nothing ? 1.0 : inv(period))
    return PeriodicTransform{typeof(frequency)}(frequency)
end
```
I am mainly questioning if it is worth adding a new function `with_period` just to avoid having to write `inv` explicitly. Maybe in general we could add more keyword argument based constructors of transforms, similar to the kernels, to make it clearer what the individual arguments mean and to allow different parameterizations.

st--:
Hmm, having looked at the code again, I'm wondering about what the right way for defining periodic kernels actually is. `PeriodicTransform` only works on 1D inputs, and maps each Real to two values. (So the base kernel needs to handle 2D inputs correctly.) For distance-based kernels, the combination of Euclidean metric & PeriodicTransform can be done in closed form - this is effectively what the `Sinus` "distance" does, so _in principle_ we could achieve the same by constructing e.g. `SqExponentialKernel(; metric=Sinus())` **except** the `Sinus` metric doesn't do the `sqrt()`. Which is why `PeriodicKernel` is basically same as SqExponential, but without the square. And then it's kinda confusing that the period is set with the lengthscale transform, and the lengthscale is set as an argument to the period...

st--:
That was quite rambly. So in bullet points:
- it'd be great if we could construct all sorts of periodic kernels (based on any kernel supporting an Euclidean metric?) - `PeriodicTransform` does that (I think?)
- it'd be great if we could avoid the intermediate ℝ → ℝ ² conversion required by `PeriodicTransform` - `PeriodicKernel` does this (but only for SqExponential base kernel)
- it'd be great if we could directly specify the period and lengthscale in ways that make sense when just reading code (`PeriodicTransform(; period=...)` is good, `with_lengthscale(PeriodicKernel(), period)` not so good)

devmotion:
> PeriodicTransform only works on 1D inputs, and maps each Real to two values.

This is the main reason why I still think that PeriodicTransform should map to complex numbers with `cis` or `cispi` (only available in more recent Julia versions). Then the "outer" dimensionality and structure would not be changed.

st--:
> This is the main reason why I still think that PeriodicTransform should map to complex numbers with `cis` or `cispi` (only available in more recent Julia versions). Then the "outer" dimensionality and structure would not be changed.

Then the base kernel would have to handle complex numbers, how would that affect speed e.g. when trying to run it on the GPU? It seems like it'd still be more extra effort than just computing the "equivalent distance" for the base kernel, no?

PR #397, opened 2021-11-03
[WIP] Use scalar parameters with ParameterHandling

devmotion:
This PR implements https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/299#issuecomment-943761371. See the comment and the additional explanations in the issue for more details.

This PR is still WIP and tests will fail (I assume).

PR #403, opened 2021-11-09
use only() instead of first()

st--:
…for 1-"vectors" that were for the benefit of Flux.

https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/397#issuecomment-958873094 says this only works for Julia 1.4+, but we already depend on the Compat package, which backports `only`.

This has the advantage that it a) makes clearer which parameters are actually scalar and b) until #397 is ready to be merged, adds additional error-checking.

### DIFF
--- a/src/basekernels/constant.jl
+++ b/src/basekernels/constant.jl
@@ -73,8 +73,8 @@ end
 
 @functor ConstantKernel
 
-kappa(κ::ConstantKernel, x::Real) = first(κ.c) * one(x)
+kappa(κ::ConstantKernel, x::Real) = only(κ.c) * one(x)
 
 metric(::ConstantKernel) = Delta()
 
-Base.show(io::IO, κ::ConstantKernel) = print(io, "Constant Kernel (c = ", first(κ.c), ")")
+Base.show(io::IO, κ::ConstantKernel) = print(io, "Constant Kernel (c = ", only(κ.c), ")")
--- a/src/basekernels/exponential.jl
+++ b/src/basekernels/exponential.jl
@@ -137,7 +137,7 @@ end
 
 @functor GammaExponentialKernel
 
-kappa(κ::GammaExponentialKernel, d::Real) = exp(-d^first(κ.γ))
+kappa(κ::GammaExponentialKernel, d::Real) = exp(-d^only(κ.γ))
 
 metric(k::GammaExponentialKernel) = k.metric
 
@@ -145,6 +145,6 @@ iskroncompatible(::GammaExponentialKernel) = true
 
 function Base.show(io::IO, κ::GammaExponentialKernel)
     return print(
-        io, "Gamma Exponential Kernel (γ = ", first(κ.γ), ", metric = ", κ.metric, ")"
+        io, "Gamma Exponential Kernel (γ = ", only(κ.γ), ", metric = ", κ.metric, ")"
     )
 end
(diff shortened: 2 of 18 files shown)


devmotion:
https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/397 would remove basically all these occurrences of `first`. I am bit worried that this PR introduces many merge conflicts.

theogf:
I would also be in favor of moving #397 forward, this looks like a lot of unnecessary overlapping

st--:
I prepared this because it seemed like #397 might take rather longer to be ready for merge as it's still WIP.
I'm happy to fix all the merge conflicts for you (by making a subbranch off dw/scalar_parameterhandling and cleaning up the merge of either this branch or master in there), if you like?

st--:
In fact I've already fixed your merge conflicts here: https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/new/st/scalar_parameterhandling_clean_merge

st--:
Just to clarify, I'm still in favour of moving #397 forward. I just think this would be a neat improvement in the meantime, without impeding #397.

devmotion:
Even if the PR takes some weeks until it is merged (hopefully it doesn't, there's not much to implement and fix, mainly some discussions about the design), I don't think it's useful to introduce these merge conflicts. This PR doesn't seem to fix any bugs or pressing issues, so I think it's not worth the additional time and effort that someone has to spend.

devmotion:
> In fact I've already fixed your merge conflicts here:

I appreciate this but I would like to keep the git history in the PR clean. I know you spent time on this but I really don't see the need for this PR given that the other one will be merged reasonably soon and it does not fix any issues (also most/all vectors are constructed internally).

st--:
>  I don't think it's useful to introduce these merge conflicts. This PR doesn't seem to fix any bugs or pressing issues, so I think it's not worth the additional time and effort that someone has to spend.

Well, it's something that's been bugging me for a long time, and I would consider it a bug that the following runs, instead of erroring:
```julia
julia> k = with_lengthscale(SqExponentialKernel(), [2, 3])
Squared Exponential Kernel (metric = Distances.Euclidean(0.0))
	- ARD Transform (dims: 2)

julia> k(4., 2.)
0.6065306597126334
```
whereas it errors on this branch.

I agree it's not really worth spending lots of time arguing back and forth, but I don't think there's any time involved in fixing any merge conflicts. I've already done so in a branch, and I'm happy to update the branch to make sure master merges cleanly into your future branch without any effort on your side! You can just do `git checkout --ours --` anyways.

> I appreciate this but I would like to keep the git history in the PR clean. 

Could you help me gain a better understanding of a) what would be "unclean" by merging master into your branch and b) what value you get out of avoiding that merge ?

st--:
@devmotion @willtebbutt  bump

devmotion:
I still don't think it's necessary and that the disadvantages (such as merge conflicts - BTW not only in the PR discussed above but also eg in the `show` PR) outweigh potential improvements which won't be needed anyway in the not-to-distant future.

st--:
Could you make "not-too-distant future" a bit more precise ? And I'd still like to understand where you see the disadvantages - re merge conflicts, as I said before, I don't think it's a big hassle _and_ I already volunteered to sort them all out for you...

theogf (approved):

Issue #338, opened 2021-07-13
Things to change before making a breaking release


thomasgudjonwright:
1) Renaming of multi-output related items, see  
https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/328 and  
https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/332

2) Change the multi output input type definitions to not have abstract types as fields. Change to something like:
```
struct MOInputIsotopicByFeatures{S,T<:AbstractVector{S}, U<:Integer} <: AbstractVector{Tuple{S,U}}
    x::T
    out_dim::U
end
```

Add to the list as you see fit :)

willtebbutt:
As @thomasgudjonwright pointed out, the docstrings for `MOInputsBy...` need to be added to the docs: https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/353#discussion_r687580200

PR #405, opened 2021-11-10
**BREAKING** Type Stability for MOInputs

willtebbutt:
**Summary**

Make input collection types for isotopic multi-output things type stable.

**Proposed changes**

Change the `Integer` type constraint in `MOInputIsotopicByFeatures` and `MOInputIsotopicByOutputs` with`Int`s.

Could have parametrised the type to allow any `Integer`, but I think this is probably fine.

**Breaking changes**

If anyone is currently using anything other than an `Int` to specify the number of outputs, they'll need to change it to an `Int`. I can't imagine that this happens very often, so effects should be minimal.


st--:
Seems fine to me.
For the doc issue, you need to update the Manifests of doc/ and examples/*/.
For the test issue, see https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/409

theogf (approved):

willtebbutt:
@theogf @st-- there are a number of other things that we've wanted to address in the next breaking release for a while: https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/338

I don't have time to push these through right now, so I'm going to hold off on merging this until we've got them in place either in this PR or another, so that we don't have to block patch releases on master.

theogf:
Should be generally have a branch `breaking_master` where we can merge these breaking PRs?

willtebbutt:
Closed in favour of #465 , which is non-breaking

PR #406, opened 2021-11-10
first go at extending the GibbsKernel to work with other base kernels

Cyberface:
**Summary**
@st-- in issue https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/372 suggested modifying the GibbsKernel to take as an argument a base kernel function.

This is a first cut at trying to implement this to get the conversation going. I'm sure I haven't implemented it correctly and I have changed any docstrings yet.

**Proposed changes**

* adding a new function argument `base_kernel` which is by default the `SqExponentialKernel`. 

**Breaking changes**

### DIFF
--- a/src/kernels/gibbskernel.jl
+++ b/src/kernels/gibbskernel.jl
@@ -36,11 +36,11 @@ end
 
 GibbsKernel(; lengthscale) = GibbsKernel(lengthscale)
 
-function (k::GibbsKernel)(x, y)
+function (k::GibbsKernel)(x, y, base_kernel::Kernel=SqExponentialKernel)
     lengthscale = k.lengthscale
     lx = lengthscale(x)
     ly = lengthscale(y)
     l = invsqrt2 * hypot(lx, ly)
-    kernel = (sqrt(lx * ly) / l) * with_lengthscale(SqExponentialKernel(), l)
+    kernel = (sqrt(lx * ly) / l) * with_lengthscale(base_kernel(), l)
     return kernel(x, y)
 end


devmotion:
We can't pass additional arguments to `kernel(x, y)`, so implementation-wise the base kernel would have to be a field of `kernel`.

More generally, I wonder though if this generalization is mathematically. IIRC we had a discussion in the GibbsKernel PR in whixh @theogf and/or @willtebbutt were worried that allowing eg different metrics would already be problematic. I wasn't sure and didn't spend more thought on it but if this would be the case, wouldn't it be even more problematic to allow general kernels (which also allow you to pass a Gaussian kernel wirh a different metric)?

st--:
As discussed in reference [[4]](https://proceedings.neurips.cc/paper/2003/file/326a8c055c0d04f5b06544665d8bb3ea-Paper.pdf) of #372 and [Paciorek's PhD thesis](https://www.stat.berkeley.edu/~paciorek/diss/paciorek-thesis.pdf), I believe "stationary" is sufficient criterion (i.e. kernels that are implemented by a `kappa` of a scalar argument), assuming that the base kernel is positive definite in any dimension (I don't fully understand why the second part wouldn't just be trivial, but I'm assuming it's got something to do with the metric being valid in higher dimensions?) - in any case it seems that we'd have to somehow define "this kernel is stationary" - @willtebbutt suggested a trait, i.e. we need definitions `isstationary(::Kernel) = nothing` and `isstationary(::SqExponentialKernel) = true` etc. Then it should be possible to have this for arbitrary base kernels.

st--:
> We can't pass additional arguments to `kernel(x, y)`, so implementation-wise the base kernel would have to be a field of `kernel`.

:+1:

PR #408, opened 2021-11-18
Fix typo in valid_inputs error

theogf:
This is just typo fix in an error message, it just pains me so much every time I see it

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -178,7 +178,7 @@ function validate_inputs(x, y)
     if dim(x) != dim(y) # Passes by default if `dim` is not defined
         throw(
             DimensionMismatch(
-                "Dimensionality of x ($(dim(x))) not equal to that of y ($(dim(y)))"
+                "dimensionality of x ($(dim(x))) is not equal to that of y ($(dim(y)))"
             ),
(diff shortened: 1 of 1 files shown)


willtebbutt (approved):

theogf:
yay a random error from Zygote!

st--:
See https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/409

src/utils.jl
```diff
@@ -178,7 +178,7 @@ function validate_inputs(x, y)
     if dim(x) != dim(y) # Passes by default if `dim` is not defined
         throw(
             DimensionMismatch(
-                "Dimensionality of x ($(dim(x))) not equality to that of y ($(dim(y)))"
+                "Dimensionality of x ($(dim(x))) not equal to that of y ($(dim(y)))"
```
devmotion, line 181: :+1: Since we are about to change it, maybe we could make it a proper sentence and follow the Julia docs and start with a lower-case letter?
```suggestion
                "dimensionality of x ($(dim(x))) is not equal to that of y ($(dim(y)))"
```

PR #409, opened 2021-11-19
Fix for Zygote 0.6.30 breaking our tests

st--:
Some change in Zygote 0.6.30 seems to break our tests.


willtebbutt (approved):
Assuming that tests pass, please bump the patch and merge this.

edit: I guess we don't need a patch bump as it's just part of the tests. Please just merge when tests pass.

theogf:
I would put my money on https://github.com/FluxML/Zygote.jl/pull/785

devmotion:
It's nice to fix the tests but unfortunately it won't help users with newer Zygote versions. Did anyone check if the type inference problem causes a performance regression?

devmotion:
Generally, I think it would better to mark the failing test as broken since then 1) we don't test the package with an outdated Zygote version (the latest version might fix or cause some other issues without us noticing it) and 2) we notice if the problem is fixed upstream without following the Zygote development.

devmotion:
Can we use something like `@test_throws ErrorException @inferred(...)`?

theogf:
Can we get this fast-forwarded? It blocks the rest of the PRs

test/test_utils.jl
```diff
@@ -100,7 +100,7 @@ function check_zygote_type_stability(f, args...; ctx=Zygote.Context())
     @inferred f(args...)
     @inferred Zygote._pullback(ctx, f, args...)
     out, pb = Zygote._pullback(ctx, f, args...)
-    @inferred pb(out)
+    @test_broken @inferred pb(out)
```
devmotion, line 103: I just checked and my suggestion seems to work:
```julia
julia> using Test

julia> f(x) = rand() < 0.5 ? zero(x) : 1

julia> @inferred f(1.0)
ERROR: return type Float64 does not match inferred return type Union{Float64, Int64}
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:33
 [2] top-level scope
   @ REPL[3]:1

julia> @test_throws ErrorException @inferred f(1.0)
Test Passed
      Thrown: ErrorException

julia> g(x) = rand() < 0.5 ? 0 : 1

julia> @test_throws ErrorException @inferred g(1.0)
Test Failed at REPL[6]:1
  Expression: #= REPL[6]:1 =# @inferred g(1.0)
    Expected: ErrorException
  No exception thrown
ERROR: There was an error during testing
```
Whereas `@test_broken` is not very helpful (as you noticed):
```julia
julia> @test_broken @inferred f(1.0)
Test Broken
  Expression: #= REPL[8]:1 =# @inferred f(1.0)

julia> @test_broken @inferred g(1.0)
Test Broken
  Expression: #= REPL[9]:1 =# @inferred g(1.0)
```
So I suggest we just change this here to
```suggestion
    @test_throws ErrorException @inferred pb(out)
```
and close https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/410 :slightly_smiling_face:

theogf (approved):

st--:
@devmotion thanks, I didn't think of that :facepalm: and only just saw the message. Good it's sorted now!

st--:
Would've just added a comment to the `@test_throws` to say "this is because test is broken, not because we actually want to ensure that this throws an error" - though hopefully we'll remember it anyways as soon as Zygote fixes it ;)

PR #414, opened 2021-12-09
Zygote AD failure workarounds & test cleanup

st--:
Zygote seems to have changed again and we now need to revert #409 

Edit: there is now a new AD failure in the select transform (#415), so temporarily disabled that test case.

In the process of trying (and failing) to figure out what's going on, I noticed that there was a lot of code duplication in test/test_utils which I've removed here (left comments to explain what I changed).

### DIFF
--- a/test/test_utils.jl
+++ b/test/test_utils.jl
@@ -280,6 +280,9 @@ function test_AD(AD::Symbol, kernelfunction, args=nothing, dims=[3, 3])
                     compare_gradient(AD, args) do p
                         testfunction(kernelfunction(p), A, dim)
                     end
+                    compare_gradient(AD, args) do p
+                        testfunction(kernelfunction(p), A, B, dim)
+                    end
                 end
 
                 compare_gradient(AD, A) do a
@@ -294,6 +297,9 @@ function test_AD(AD::Symbol, kernelfunction, args=nothing, dims=[3, 3])
                 if args !== nothing
                     compare_gradient(AD, args) do p
                         testdiagfunction(kernelfunction(p), A, dim)
+                    end
+                    compare_gradient(AD, args) do p
+                        testdiagfunction(kernelfunction(p), A, B, dim)
                     end
                 end
             end
--- a/test/test_utils.jl
+++ b/test/test_utils.jl
@@ -224,65 +224,77 @@ end
 
 function test_AD(AD::Symbol, kernelfunction, args=nothing, dims=[3, 3])
     @testset "$(AD)" begin
-        # Test kappa function
         k = if args === nothing
             kernelfunction()
         else
             kernelfunction(args)
         end
         rng = MersenneTwister(42)
+
         if k isa SimpleKernel
-            for d in log.([eps(), rand(rng)])
-                compare_gradient(AD, [d]) do x
-                    kappa(k, exp(x[1]))
+            @testset "kappa function" begin
+                for d in log.([eps(), rand(rng)])
+                    compare_gradient(AD, [d]) do x
+                        kappa(k, exp(x[1]))
+                    end
                 end
             end
         end
-        # Testing kernel evaluations
-        x = rand(rng, dims[1])
-        y = rand(rng, dims[1])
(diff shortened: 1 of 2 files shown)


test/test_utils.jl
```diff
@@ -224,65 +224,77 @@ end
...
+                compare_gradient(AD, B) do b
+                    testdiagfunction(k, A, b, dim)
+                end
+                if args !== nothing
+                    compare_gradient(AD, args) do p
+                        testdiagfunction(kernelfunction(p), A, dim)
+                    end
                 end
             end
         end
```
st--, line 300: @rossviljoen doing the changes above I spotted that you overloaded test_AD() for MOKernel - this worries me: it seems like all the "fallback" AD tests above are not run. It also seems to suggest that we can't substitute an MOKernel where we need an arbitrary Kernel (which would I believe violate Liskov substitution principle). Could you comment pls? thanks:)

st--: Ah, apologies! Now I'm not even sure anymore why I assumed it was you :pensive:  @willtebbutt / @thomasgudjonwright ?

willtebbutt: Hmm don't think it was me. A lot of this code was @theogf I think?

theogf: I did the original compare gradient but someone else extended it to MOKernel

thomasgudjonwright: Wasn't me! Could always look back at the commit history :)

st--: Should've checked git blame in the first place... it was https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/263 (@david-vicente).

This to me seems like a case where overloading the same method isn't really the right thing to do (because it suggests they do the same thing - it's a kernel AD check in either case - but they don't actually seem to check the same things)... now that i've got y'all in a thread, what is your opinion on these ?

st--: Separated out into #416

test/test_utils.jl
```diff
@@ -114,70 +118,6 @@ function test_ADs(
     end
 end
 
-function test_FiniteDiff(kernelfunction, args=nothing, dims=[3, 3])
```
st--, line 117: This was doing exactly the same as `test_AD` (except for `testdiagfunction(kernelfunction(p), A, dim)` vs `testdiagfunction(kernelfunction(p), A, B, dim)`, which I've now added to `test_AD`). The only difference was calling `compare_gradient` vs the `@test_warn`, so I simply added a `compare_gradient(f, ::Val{:FiniteDiff}, args)` so we can avoid all this code duplication.

test/test_utils.jl
```diff
@@ -224,68 +164,69 @@ end
...
-            end
-            compare_gradient(AD, A) do a
-                testdiagfunction(k, a, B, dim)
-            end
-            compare_gradient(AD, B) do b
-                testdiagfunction(k, A, b, dim)
-            end
-            if args !== nothing
-                compare_gradient(AD, args) do p
-                    testdiagfunction(kernelfunction(p), A, dim)
```
st--, line 285: this was doing exactly the same for `testdiagfunction` as the code above for `testfunction`, so I've unified it with a for loop over the two functions.

test/test_utils.jl
```diff
@@ -224,68 +164,69 @@ end
...
+                    compare_gradient(AD, B) do b
+                        _testfn(k, A, b, dim)
+                    end
+                    if !(args === nothing)
+                        @testset "hyperparameters" begin
+                            compare_gradient(AD, args) do p
+                                _testfn(kernelfunction(p), A, dim)
+                            end
+                            compare_gradient(AD, args) do p
+                                _testfn(kernelfunction(p), A, B, dim)
```
st--, line 222: added missing case

test/test_utils.jl
willtebbutt: Very nice.

willtebbutt (approved):
I'm happy for this to go in. Thanks for sorting this out @st--

PR #419, opened 2021-12-23
St/dw/scalar parameterhandling/merge

st--:
Merges master back into #397

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "KernelFunctions"
 uuid = "ec8451be-7e33-11e9-00cf-bbf324bd1392"
-version = "0.10.26"
+version = "0.10.27"
 
 [deps]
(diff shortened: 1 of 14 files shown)


devmotion (commented):
The branch is outdated, I have additional changes and fixes locally. So unfortunately it's not helpful ro merge this PR currently. I'm sorry but exactly such problems I wanted to avoid, I still think it was a bad idea to make these changes.

st--:
> The branch is outdated, I have additional changes and fixes locally. So unfortunately it's not helpful ro merge this PR currently. I'm sorry but exactly such problems I wanted to avoid, I still think it was a bad idea to make these changes.

Well, it wasn't any work; let me know once you've pushed your local changes, I'm happy to re-do it. I still think it's not a big problem and that it was better to get the other PR in as it is.

PR #420, opened 2021-12-23
Stop testing on 1.3

willtebbutt:
IIRC we agreed to stop testing on 1.3 once 1.6 became the LTS.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
       matrix:
         version:
           - '1'
-          - '1.3'
+          - '1.6'
           - 'nightly'
         os:
           - ubuntu-latest


devmotion:
Just my two cents: I think dropping tests but not upper bounding the Julia version will lead to silent breakage (had to fix too many of these in other packages...). Regarding compat bounds, I think there should be a clear motivation for dropping old Julia versions, and personally I don't think it's particularly useful to drop < 1.6 just because 1.6 is the new LTS *if* the package works fine with older Julia versions. However, e.g. I don't think we should spend time on tracking down and fixing bugs that are specific to older Julia versions or on making PRs compatible with these versions if it requires significant effort.

st--:
Whilst we proclaim compatibility with 1.3, we should keep running tests on 1.3. I think at this point we shouldn't do anything extra _to maintain 1.3 compatibility_, but as long as it's at no extra cost, we might as well keep it going. The change I'd like to see is more on the process level: once we've got a PR that fails on 1.3 (but works on 1.6+), _then_ we should remove 1.3 compatibility (and drop the test).

willtebbutt:
Sounds good to me.

PR #422, opened 2022-01-11
fix figure & cleanup

st--:
@theogf apologies I had made a mistake with Plots... and then I also cleaned up the code a bit more


st--:
I've taken the opportunity to bump the LIBSVM dependency, deprecating #413

examples/support-vector-machine/script.jl
```diff
@@ -17,16 +17,15 @@ Random.seed!(1234);
...
-class1x = cos.(range(0, π; length=nout))
-class1y = sin.(range(0, π; length=nout))
-class2x = 1 .- cos.(range(0, π; length=nin))
-class2y = 1 .- sin.(range(0, π; length=nin)) .- 0.5
-X = hcat(vcat(class1x, class2x), vcat(class1y, class2y))
-X .+= 0.1randn(size(X))
+angle1 = range(0, π; length=n1)
+angle2 = range(0, π; length=n2)
+X1 = [cos.(angle1) sin.(angle1)] .+ 0.1randn(n1, 2)
+X2 = [1 .- cos.(angle2) 1 .- sin.(angle2) .- 0.5] .+ 0.1randn(n2, 2)
```
devmotion, line 27: `0.1randn(n1, 2)` creates two arrays that can be avoided:
```suggestion
X1 = [cos.(angle1) sin.(angle1)] .+ 0.1 .* randn.()
X2 = [1 .- cos.(angle2) 1 .- sin.(angle2) .- 0.5] .+ 0.1 .* randn.()
```
Alternatively, with one allocation:
```suggestion
X1 = [cos.(angle1) sin.(angle1)] .+ 0.1 .* randn(n1, 2)
X2 = [1 .- cos.(angle2) 1 .- sin.(angle2) .- 0.5] .+ 0.1 .* randn(n2, 2)
```

examples/support-vector-machine/script.jl
```diff
@@ -49,7 +48,7 @@ model = svmtrain(kernelmatrix(k, x_train), y_train; kernel=LIBSVM.Kernel.Precomp
 y_pred, _ = svmpredict(model, kernelmatrix(k, x_train, x_test));
 
 # Visualize prediction on a grid:
-plot(; lim=extrema(test_range), aspect_ratio=1)
+plot(; xlim=extrema(test_range), ylim=extrema(test_range), aspect_ratio=1)
```
devmotion, line 51: I guess you want
```suggestion
plot(; widen=false, aspect_ratio=1)
```
?

st--: No, that doesn't fix it.

devmotion: Ah, I guess you might have to specify the keyword argument when you actually plot data since otherwise subsequent calls might update the limits. I missed that there's nothing plotted in this call here.

examples/support-vector-machine/script.jl
st--: In this case, the number of allocations makes no practical difference... but not having to pass the shape is neat! I didn't know `randn.()` works like that, I would have thought it only shifts everything by the same amount.

theogf: Damn! This is nice! I never new about  `x .+ randn.()` is it working because of broadcast fusion?

examples/support-vector-machine/script.jl
st--: No, widen isn't actually needed. Looks like updating Plots.jl seems to have fixed it though :sweat_smile:

st--:
Can I have an approve to merge while even the nightly build is passing? :joy:

theogf (approved):

PR #425, opened 2022-01-13
fix MaternKernel AD, but remove differentiation wrt \nu

st--:
~to keep track of whether it's working again~

~Now changed so that it only tests Forward/Reverse diff AD, which are working fine at this point.

Not sure why Zygote AD test is failing, or how to fix it :|~

So..... turns out the Zygote issue was the accessing of the kernel object's field. :exploding_head: 

Something like this works fine:
```julia
using KernelFunctions, Zygote
@inline function matern_kappa_hardcoded_nu(d::Real)
       result = KernelFunctions._matern(1.234, d)
       return ifelse(iszero(d), one(result), result)
end
Zygote.gradient(matern_kappa_hardcoded_nu, 2.345)
```

So the solution was to simply wrap the `k.\nu` in a `ignore_derivatives`... which is fine because the docstring explicitly denies this anyways.


devmotion:
It's unlikely that this will be fixed anytime soon since eg for ChainRules derivatives wrt nu are not implemented in SpecialFunctions because they are more involved and would require additional dependencies.

theogf:
The problem is not differentiating through \nu no? The problem is to differentiate through the Bessel function given the distance input.
It's quite clear that the derivatives given \nu are too ugly and complicated to deal with and we should make it clear that the kernel is not differentiable by \nu

devmotion:
That's what's tested and commented on in the test though, isn't it?

theogf:
That's true. That's why I am proposing that we just ban differentiation through \nu and stop testing for it.

devmotion:
Sounds good to me 🙂

theogf:
> Would you want to do that explicitly in code? Or just add a comment to the docstring?

I think adding a comment to the docstring is enough

src/basekernels/matern.jl
```diff
@@ -17,6 +17,10 @@ By default, ``d`` is the Euclidean metric ``d(x, x') = \\|x - x'\\|_2``.
 A Gaussian process with a Matérn kernel is ``\\lceil \\nu \\rceil - 1``-times
 differentiable in the mean-square sense.
 
+!!! note
+
+    Differentiation with respect to the ν parameter is not currently supported.
```
devmotion, line 22: I think we should be a bit clearer since we don't plan to change this:
```suggestion
    Differentiation with respect to the order ν is not supported.
```

st--: Well, if someone wanted to contribute all the derivatives I'd be happy to include it

st--:
```suggestion
    Differentiation with respect to the order ν is not currently supported.
```

theogf: I would more expect request like: when do you plan to support this differentation

devmotion: To a large extent it's also an upstream issue, SpecialFunctions and DiffRules implement the derivatives of the bessel functions but not wrt to the order.

devmotion:
Might be an upstream issue, it seems Zygote tries to convert ChainRulesCore.NotImplemented which by design does not support most operations.

src/matrix/kernelkroneckermat.jl
```diff
@@ -14,7 +14,7 @@ where `D` is given by `dims`.
 
 !!! warning
 
-    Require `Kronecker.jl` and for `iskroncompatible(κ)` to return `true`.
+    Requires `Kronecker.jl` and for `iskroncompatible(κ)` to return `true`.
```
devmotion, line 17: Unrelated change it seems?
```suggestion
    Require `Kronecker.jl` and for `iskroncompatible(κ)` to return `true`.
```

test/basekernels/matern.jl
```diff
@@ -4,25 +4,23 @@
     v1 = rand(rng, 3)
     v2 = rand(rng, 3)
     @testset "MaternKernel" begin
-        ν = 2.0
+        ν = 2.1
```
devmotion, line 7: Why not 2.0?

st--: if we're testing the general case, I'd prefer to test it at a general number (just how if you want to test some matrix functions it's better to test it for a 3x5 matrix not a 1x1 for which it might accidentally work...)

src/matrix/kernelkroneckermat.jl
st--: would you like this in a separate PR ?

src/basekernels/matern.jl
```diff
@@ -33,8 +37,9 @@ MaternKernel(; nu::Real=1.5, ν::Real=nu, metric=Euclidean()) = MaternKernel(ν,
 
 @functor MaternKernel
 
-@inline function kappa(κ::MaternKernel, d::Real)
-    result = _matern(only(κ.ν), d)
+@inline function kappa(k::MaternKernel, d::Real)
+    nu = ChainRulesCore.@ignore_derivatives only(k.ν)  # work-around for Zygote AD
```
willtebbutt, line 41: I wonder whether this is quite the right way to handle this?
In particular, what happens if someone _does_ try to compute gradients w.r.t. `k.ν`? Do they silently get the wrong answer?

I wonder whether it would make more sense to try and return a `NotImplemented` tangent, see [here](https://github.com/JuliaDiff/ChainRulesCore.jl/blob/c979ab7fd9bea979fdb659e1d5a5c78ed4db9b44/src/tangent_types/notimplemented.jl#L21), which should error if the user ever tries to do anything with the tangent.

st--: Yes, this change would mean that there should be a zero gradient wrt k.\nu. @theogf suggested previously that a note in the docstring would be enough.

I agree in principle returning an error would be better, but I wouldn't be sure off the top of my head how to actually implement that. Could we leave that for an issue/future PR?

st--: NB- I don't know _why_ the field access confuses Zygote. It's not got anything to do with the gradient not being implemented!?

willtebbutt: I'm really not sure that I'm comfortable with a silently-incorrect gradient -- if I encountered this as a user, I would be furious 😂 . I'll take a look at how to do this on my machine now and provide a suggestion.

st--: if you can fix it right away, that also works:)

st--: Personally I'd still prefer a docstring-explained missing gradient wrt one hyperparameter over no gradients working :smile:

willtebbutt: Haha fair enough. If I can't find a fix, we should merge it.

willtebbutt, line 41:
```suggestion
# Work-around for Zygote -- `NotImplemented` doesn't appear to play nicely with whatever
# rule currently exists for `only`.
_get_ν(k::MaternKernel) = only(k.ν)
function ChainRulesCore.rrule(::typeof(_get_ν), k::T) where {T<:MaternKernel}
    function _get_ν_pullback(Δ)
        dν = @not_implemented("Derivatives w.r.t. ν are not implemented.")
        return Tangent{T}(ν=dν, metric=NoTangent())
    end
    return _get_ν(k), _get_ν_pullback
end

@inline function kappa(k::MaternKernel, d::Real)
    nu = _get_ν(k)
```

I think this will do it, if you `using ChainRulesCore` somewhere.

src/basekernels/matern.jl
```diff
@@ -33,8 +37,20 @@ MaternKernel(; nu::Real=1.5, ν::Real=nu, metric=Euclidean()) = MaternKernel(ν,
...
+    function _get_ν_pullback(Δ)
+        dν = ChainRulesCore.@not_implemented("Derivatives w.r.t. ν are not implemented.")
+        return Tangent{T}(ν=dν, metric=NoTangent())
+    end
+    return _get_ν(k), _get_ν_pullback
+end
+
+@inline function kappa(k::MaternKernel, d::Real)
+    nu = _get_ν(k)
+    result = _matern(nu, d)
```
st--, line 53:
```suggestion
    result = _matern(_get_ν(k), d)
```

src/zygoterules.jl
```diff
@@ -1,13 +1,13 @@
...
-ZygoteRules.@adjoint function Base.map(t::Transform, X::RowVecs)
-    return ZygoteRules.pullback(_map, t, X)
-end
-
-function ZygoteRules._pullback(
-    cx::AContext, ::typeof(literal_getproperty), x::ColVecs, ::Val{f}
-) where {f}
-    return ZygoteRules._pullback(cx, literal_getfield, x, Val{f}())
-end
+#ZygoteRules.@adjoint function Base.map(t::Transform, X::ColVecs)
```
willtebbutt, line 1: Why are these commented out?

st--: accidental commit :sweat_smile: I'll revert it, I'm just curious now to see which tests actually require these so I'll let them run before pushing

src/basekernels/matern.jl
```diff
@@ -33,8 +37,10 @@ MaternKernel(; nu::Real=1.5, ν::Real=nu, metric=Euclidean()) = MaternKernel(ν,
 
 @functor MaternKernel
 
-@inline function kappa(κ::MaternKernel, d::Real)
-    result = _matern(only(κ.ν), d)
+@inline _get_ν(k::MaternKernel) = ChainRulesCore.@ignore_derivatives only(k.ν)  # work-around for Zygote AD
```
devmotion, line 40: Feel free to ignore my comment but one should not use `@ignore_derivatives` here:
``suggestion
# work-around for Zygote AD
@inline _get_ν(k::MaternKernel) = only(k.ν)
ChainRulesCore.@non_differentiable _get_ν(::MaternKernel)
```

willtebbutt (approved):
I'm basically happy with this, please feel free to merge once the commented-out code has been un-commented :)

edit: also, patch bump

src/basekernels/matern.jl
st--: Hahah, I actually just thought of that and already did it :blush:

PR #427, opened 2022-01-21
make nystrom work with AbstractVector

st--:
Resolves #426

PR #429, opened 2022-01-27
Deprecate implicit `obsdim` keyword argument

devmotion:
Lately, due to the discussions also in AbstractGPs, I thought a bit about our support of matrices and the `obsdim` keyword argument.

I realized that I like it that one can still call `kernelmatrix` etc. with matrices, even though internally we only work with `RowVecs` and `ColVecs` (fortunately :pray:). I think this might make it a bit easier for new users as they don't have to learn about `ColVecs` and `RowVecs` right away - and even to me sometimes it feels quite nice, e.g., if other parts and packages of my script don't work with, or don't even know about, `ColVecs`/`RowVecs` and the hence the most natural form of the data is a raw matrix with a clearly defined convention (rows or columns).

However, I also realized that I really don't like it that there is an implicit default `obsdim` since this makes code more difficult to read and reason about and could lead to misunderstandings and bugs. And any default choice is somewhat arbitrary and probably heavily biased by research area and personal preferences.

I think one can get both - matrices as arguments and no default convention - by deprecating the default implicit `obsdim` keyword argument and requiring users to state `obsdim` explicitly. A prominent example of another package that did this is Distances.jl.

The idea grew on me, and hence I put together this PR. It has to be updated once #427 is merged.

### DIFF
--- a/src/approximations/nystrom.jl
+++ b/src/approximations/nystrom.jl
@@ -9,7 +9,7 @@ function sampleindex(X::AbstractVector, r::Real)
     return S
 end
 
-function nystrom_sample(k::Kernel, X::AbstractVector, S::Vector{<:Integer})
+function nystrom_sample(k::Kernel, X::AbstractVector, S::AbstractVector{<:Integer})
     C = kernelmatrix(k, X[S], X)
     Cs = C[:, S]
     return (C, Cs)
--- a/src/approximations/nystrom.jl
+++ b/src/approximations/nystrom.jl
@@ -9,9 +9,7 @@ function sampleindex(X::AbstractVector, r::Real)
     return S
 end
 
-function nystrom_sample(
-    k::Kernel, X::AbstractVector, S::Vector{<:Integer}
-)
+function nystrom_sample(k::Kernel, X::AbstractVector, S::Vector{<:Integer})
     C = kernelmatrix(k, X[S], X)
     Cs = C[:, S]
     return (C, Cs)
@@ -84,7 +82,12 @@ If `obsdim=2`, equivalent to `nystrom(k, ColVecs(X), S)`.
 
 See also: [`ColVecs`](@ref), [`RowVecs`](@ref)
 """
-function nystrom(k::Kernel, X::AbstractMatrix, S::AbstractVector{<:Integer}; obsdim::Union{Int,Nothing}=defaultobs)
+function nystrom(
+    k::Kernel,
+    X::AbstractMatrix,
+    S::AbstractVector{<:Integer};
+    obsdim::Union{Int,Nothing}=defaultobs,
+)
     return nystrom(k, vec_of_vecs(X; obsdim=obsdim), S)
 end
 
--- a/src/approximations/nystrom.jl
+++ b/src/approximations/nystrom.jl
@@ -1,21 +1,18 @@
 # Following the algorithm by William and Seeger, 2001
 # Cs is equivalent to X_mm and C to X_mn
 
-function sampleindex(X::AbstractMatrix, r::Real; obsdim::Integer=defaultobs)
+function sampleindex(X::AbstractVector, r::Real)
     0 < r <= 1 || throw(ArgumentError("Sample rate `r` must be in range (0,1]"))
-    n = size(X, obsdim)
+    n = length(X)
     m = ceil(Int, n * r)
     S = StatsBase.sample(1:n, m; replace=false, ordered=true)
     return S
 end
 
 function nystrom_sample(
-    k::Kernel, X::AbstractMatrix, S::Vector{<:Integer}; obsdim::Integer=defaultobs
+    k::Kernel, X::AbstractVector, S::Vector{<:Integer}
 )
-    obsdim ∈ [1, 2] ||
-        throw(ArgumentError("`obsdim` should be 1 or 2 (see docs of kernelmatrix))"))
-    Xₘ = obsdim == 1 ? X[S, :] : X[:, S]
-    C = kernelmatrix(k, Xₘ, X; obsdim=obsdim)
+    C = kernelmatrix(k, X[S], X)
     Cs = C[:, S]
     return (C, Cs)
 end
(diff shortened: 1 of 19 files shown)


theogf:
So the idea behind the deprecation is that `defaultobs` is now `nothing` and that it will trigger an warning but still work like before and we then make a new breaking release without this default?

devmotion:
Yes, exactly, currently it is just changing `defaultobsdim` to `nothing`, widening the type annotation, and adding a deprecation warning if `obsdim=nothing`. In an upcoming breaking release we would then remove `defaultobsdim` and change `obsdim::Union{Int,Nothing}=defaultobsdim` to just `obsdim::Int`, without default value.

theogf:
Can you maybe comment or introduce what will be the behavior when the user will not provide obsdim (once the deprecations are removed)

devmotion:
The default behaviour would be:
```julia
julia> f(x; obsdim::Int) = size(x, obsdim)

julia> f(rand(3, 4))
ERROR: UndefKeywordError: keyword argument obsdim not assigned
Stacktrace:
 [1] f(x::Matrix{Float64})
   @ Main ./REPL[6]:1
 [2] top-level scope
   @ REPL[7]:1

julia> f(rand(3, 4); obsdim=nothing)
ERROR: TypeError: in keyword argument obsdim, expected Int64, got a value of type Nothing
Stacktrace:
 [1] top-level scope
   @ REPL[8]:1

julia> f(rand(3, 4); obsdim=1)
3

julia> f(rand(3, 4); obsdim=2)
4
```

If we are not happy with these messages, we could continue using `nothing` but throw an error in `vec_of_vecs` where in this PR a deprecation warning is shown.

theogf:
Yes that's exactly what we discussed last time with @st--, @willtebbutt and @rossviljoen . We think that it should throw an error but with a extremely verbose explicative message

src/utils.jl
```diff
@@ -21,9 +21,23 @@ macro check_args(K, param, cond, desc=string(cond))
...
 end
 
-function vec_of_vecs(X::AbstractMatrix; obsdim::Int=2)
-    @assert obsdim ∈ (1, 2) "obsdim should be 1 or 2, see docs of kernelmatrix"
-    if obsdim == 1
+function deprecated_obsdim(obsdim::Union{Nothing,Integer})
+    if obsdim === nothing
+        Base.depwarn(
+            "implicit `obsdim=2` argument is deprecated and now has to be passed " *
+            "explicitly to specify that columns are observation",
```
st--, line 28:
```suggestion
            "explicitly to specify that each column corresponds to one observation",
```

src/utils.jl
```diff
@@ -21,9 +21,23 @@ macro check_args(K, param, cond, desc=string(cond))
...
+        )
+        return 2
+    else
+        return obsdim
+    end
+end
+
+function vec_of_vecs(X::AbstractMatrix; obsdim::Union{Nothing,Int}=nothing)
+    _obsdim = deprecated_obsdim(obsdim)
+    @assert _obsdim ∈ (1, 2) "obsdim should be 1 or 2, see docs of kernelmatrix"
```
st--, line 39: what was your reason to here use an `@assert` instead of an explicit if & throw(ArgumentError())?

test/basekernels/matern.jl
```diff
@@ -45,7 +45,7 @@
         k = Matern52Kernel()
         @test kappa(k, x) ≈ (1 + sqrt(5) * x + 5 / 3 * x^2)exp(-sqrt(5) * x)
         @test k(v1, v2) ≈
-              (
+            (
```
st--, line 48: did JuliaFormatter change behaviour ?

st--:
Looks great, thanks! Happy to have this merged into master.

src/utils.jl
devmotion: That I didn't want to change the existing code :smile: I think it should throw an error but this seemed unrelated to the changes here.

test/basekernels/matern.jl
devmotion: Apparently 🤷‍♂️ I didn't do anything special and JuliaFormatter seems to be happy.

src/utils.jl
devmotion: Since I prefer `ArgumentError` as well I changed it nevertheless.

src/approximations/nystrom.jl
```diff
@@ -9,9 +9,9 @@ function sampleindex(X::AbstractVector, r::Real)
     return S
 end
 
-@deprecate sampleindex(X::AbstractMatrix, r::Real; obsdim::Integer=defaultobs) sampleindex(
-    vec_of_vecs(X; obsdim=obsdim), r
-) false
+@deprecate sampleindex(
+    X::AbstractMatrix, r::Real; obsdim::Union{Integer,Nothing}=defaultobs
+) sampleindex(vec_of_vecs(X; obsdim=obsdim === nothing ? nothing : Int(obsdim)), r) false
```
st--, line 14: What's the benefit of the ternary ? : over just
```suggestion
) sampleindex(vec_of_vecs(X; obsdim=obsdim), r) false
```
?

src/matrix/kernelpdmat.jl
```diff
@@ -3,15 +3,15 @@ using .PDMats: PDMat
 export kernelpdmat
 
 """
-    kernelpdmat(k::Kernel, X::AbstractMatrix; obsdim::Int=2)
+    kernelpdmat(k::Kernel, X::AbstractMatrix; obsdim)
```
st--, line 6: I really liked the explicit explanation of `obsdim`, how about adding that the same way here, too?

src/utils.jl
```diff
@@ -21,12 +21,27 @@ macro check_args(K, param, cond, desc=string(cond))
...
+function deprecated_obsdim(obsdim::Union{Int,Nothing})
+    return if obsdim === nothing
+        Base.depwarn(
+            "implicit `obsdim=2` argument is deprecated and now has to be passed " *
+            "explicitly to specify that each column corresponds to one observation",
+            :vec_of_vecs,
+        )
+        2
     else
+        obsdim
```
st--, line 33: I found the previous version rather more readable / more easily understandable:
```suggestion
    if obsdim === nothing
        Base.depwarn(
            "implicit `obsdim=2` argument is deprecated and now has to be passed " *
            "explicitly to specify that each column corresponds to one observation",
            :vec_of_vecs,
        )
        return 2
    else
        return obsdim
```

src/utils.jl
```diff
@@ -21,12 +21,27 @@ macro check_args(K, param, cond, desc=string(cond))
...
+    end
+end
+
+function vec_of_vecs(X::AbstractMatrix; obsdim::Union{Int,Nothing}=nothing)
+    _obsdim = deprecated_obsdim(obsdim)
+    return if _obsdim == 1
+        RowVecs(X)
+    elseif _obsdim == 2
         ColVecs(X)
```
st--, line 42: Likewise, I find
```suggestion
    if _obsdim == 1
        return RowVecs(X)
    elseif _obsdim == 2
        return ColVecs(X)
```
much easier to parse

src/approximations/nystrom.jl
st--: is it because here it was `Integer` whereas `vec_of_vecs` requires more specific `Int`?

devmotion: Yes, exactly. The current deprecation is not correct since it does not handle non-Int Integers. It doesn't matter much since we're allowed to remove it any time but since I had to change it anyway I fixed it.

src/utils.jl
devmotion: I have the opposite view 😄 I prefer the version in this PR much more since I think it's cleaner and easier to parse. I read it as a short form of
```suggestion
    _obsdim = if obsdim === nothing
        Base.depwarn(
            "implicit `obsdim=2` argument is deprecated and now has to be passed " *
            "explicitly to specify that each column corresponds to one observation",
            :vec_of_vecs,
        )
        2
    else
        obsdim
    end
    return _obsdim
```
which makes it more obvious that the evaluation result of the `if` expression is returned. I guess I prefer this view and use of `if` statements as something that can be evaluated and returns a specific value due to my FP/Haskell background.

src/utils.jl
devmotion: Same here, I prefer the version in this PR.

src/matrix/kernelpdmat.jl
devmotion: Done.

src/utils.jl
theogf: It's not a major thing but for such parts of the code I would be on the side of @st--

theogf (approved):

src/utils.jl
st--: I'd still prefer your suggestion (with explicit `return _obsdim`) over the `return if ...`.

PR #430, opened 2022-01-28
Updated Train- kernel-parameters example

Crown421:
**Summary**
Picking up #317, and adding more to it. The commits might a bit of a mess, due to some unfortunate experiments with the github desktop app.

**Proposed changes**


Crown421:
I didn't build the documentation locally, is there a way to preview from here?

devmotion:
There's a preview for PRs that are not from forks, the link to the deployed docs will show up among the Github actions (see e.g. https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/429). However, currently there's no link since the documentation is not built successfully.

devmotion:
The problem is that BenchmarkTools is not installed: https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/runs/4983109277?check_suite_focus=true#step:5:31 Is it actually necessary to perform benchmarks?

Crown421:
> The problem is that BenchmarkTools is not installed: https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/runs/4983109277?check_suite_focus=true#step:5:31 Is it actually necessary to perform benchmarks?

Ah, that makes alot of sense. Not strictly necessary, but there are interesting things to see.

Crown421:
> The problem is that BenchmarkTools is not installed: https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/runs/4983109277?check_suite_focus=true#step:5:31 Is it actually necessary to perform benchmarks?

I have added BenchmarkTools to the commit, so that you can see. 
If the consensus is the benchmarks are not super exciting, they can be removed again.

examples/train-kernel-parameters/Project.toml
```diff
@@ -0,0 +1,20 @@
...
+Flux = "587475ba-b771-5e3f-ad9e-33799f191a9c"
+ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
+KernelFunctions = "ec8451be-7e33-11e9-00cf-bbf324bd1392"
+LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+Literate = "98b081ad-f1c9-55d3-8b20-4c87d4299306"
+ParameterHandling = "2412ca09-6db7-441c-8e3a-88d5709968c5"
+Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
+Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
+
+[compat]
```
st--, line 13: It's currently missing entries for BenchmarkTools, ParameterHandling, and Zygote, could you add those?

Crown421: yep, sorry about that.

Crown421:
I think this is done now. I can probably tweak this further, but I should not.

Crown421:
@st-- I think you would probably be best placed to approve this?

examples/train-kernel-parameters/script.jl
```diff
@@ -0,0 +1,235 @@
+# # Train Kernel Parameters
+
+# In this example we show a few ways to perform regression on a kernel from KernelFunctions.jl.
```
st--, line 3:
```suggestion
# Here we show a few ways to train (optimize) the kernel (hyper)parameters at the example of kernel-based regression using KernelFunctions.jl.
```

examples/train-kernel-parameters/script.jl
```diff
@@ -0,0 +1,235 @@
...
+
+# We load KernelFunctions and some other packages
+
+using KernelFunctions
+using LinearAlgebra
+using Distributions
+using Plots;
```
st--, line 10:
```suggestion
using Plots
```

examples/train-kernel-parameters/script.jl
```diff
@@ -0,0 +1,235 @@
...
+using Flux: Optimise
+using Zygote
+using Random: seed!
+seed!(42);
+
+# ## Data Generation
+# We generated data in 1 dimension
+
+xmin = -3;
+xmax = 3; # Bounds of the data
```
st--, line 22:
```suggestion
xmin, xmax = -3, 3  # Bounds of the data
```

examples/train-kernel-parameters/script.jl
```diff
@@ -0,0 +1,235 @@
...
+N = 50 # Number of samples
+x_train = rand(Uniform(xmin, xmax), N) # We sample 100 random samples
```
st--, line 24:
```suggestion
x_train = rand(Uniform(xmin, xmax), N) # sample the inputs
```

examples/train-kernel-parameters/script.jl
```diff
@@ -0,0 +1,235 @@
...
+σ = 0.1
+y_train = sinc.(x_train) + randn(N) * σ # We create a function and add some noise
```
st--, line 26:
```suggestion
y_train = sinc.(x_train) + randn(N) * σ  # evaluate a function and add some noise
```

examples/train-kernel-parameters/script.jl
```diff
@@ -0,0 +1,235 @@
...
+x_test = range(xmin - 0.1, xmax + 0.1; length=300)
+nothing #hide
+
+# Plot the data
+
+scatter(x_train, y_train; lab="data")
+plot!(x_test, sinc; lab="true function")
```
st--, line 33:
```suggestion
scatter(x_train, y_train; label="data")
plot!(x_test, sinc; label="true function")
```

examples/train-kernel-parameters/script.jl
```diff
@@ -0,0 +1,235 @@
...
+
+# ## Base Approach
```
st--, line 35:
```suggestion
# ## Manual Approach
```

examples/train-kernel-parameters/script.jl
```diff
@@ -0,0 +1,235 @@
...
+# The first option is to rebuild the parametrized kernel from a vector of parameters 
+# in each evaluation of the cost fuction. This is similar to the approach taken in 
+# [Stheno.jl](https://github.com/JuliaGaussianProcesses/Stheno.jl).
+
+# To train the kernel parameters via ForwardDiff.jl
```
st--, line 40:
```suggestion
# To train the kernel parameters via ForwardDiff.jl,
```

examples/train-kernel-parameters/script.jl
```diff
@@ -0,0 +1,234 @@
...
+
+scatter(x_train, y_train; label="data")
+plot!(x_test, sinc; label="true function")
+
+# ## Manual Approach
+# The first option is to rebuild the parametrized kernel from a vector of parameters 
+# in each evaluation of the cost fuction. This is similar to the approach taken in 
+# [Stheno.jl](https://github.com/JuliaGaussianProcesses/Stheno.jl).
+
+# To train the kernel parameters via ForwardDiff.jl,
```
st--, line 39: You don't actually use ForwardDiff though. And Zygote seems to be able to differentiate through more complex types, too, not necessarily requiring it to be a single flat vector of parameters?

examples/train-kernel-parameters/script.jl
```diff
@@ -0,0 +1,234 @@
...
+end
+nothing #hide
+
+# We look how the prediction looks like
+# with starting parameters [1.0, 1.0, 1.0, 1.0] we get :
+
+ŷ = f(x_test, x_train, y_train, log.(ones(4)))
+scatter(x_train, y_train; lab="data")
+plot!(x_test, sinc; lab="true function")
+plot!(x_test, ŷ; lab="prediction")
```
st--, line 66:
```suggestion
scatter(x_train, y_train; label="data")
plot!(x_test, sinc; label="true function")
plot!(x_test, ŷ; label="prediction")
```

examples/train-kernel-parameters/script.jl
```diff
@@ -0,0 +1,234 @@
...
+
+# We define the loss based on the L2 norm both
+# for the loss and the regularization
```
st--, line 69: do you mean
```suggestion
# for the data fit term and the regularization:
```
?

examples/train-kernel-parameters/script.jl
```diff
@@ -0,0 +1,234 @@
...
+
+function loss(θ)
+    ŷ = f(x_train, x_train, y_train, θ)
+    return sum(abs2, y_train - ŷ) + exp(θ[4]) * norm(ŷ)
```
st--, line 73: Two questions:
a) why not
```suggestion
    return norm(y_train - ŷ) + exp(θ[4]) * norm(ŷ)
```
b) can you explain where the second term comes from? Just thinking about "parameter regularisation" might make one think it ought to be `norm(\theta)`.

examples/train-kernel-parameters/script.jl
```diff
@@ -0,0 +1,234 @@
...
+opt = Optimise.ADAGrad(0.5)
+nothing #hide
+
+# The loss with our starting point:
+
+loss(θ)
+
+# Computational cost for one step
+
+@benchmark let θt = θ[:], optt = Optimise.ADAGrad(0.5)
```
st--, line 89: why do you introduce a new `optt`? (I'm assuming it's due to the internal optimizer state that you don't want to affect for the actual optimisation loop, but this might not be clear to the casual reader!)

As it's within a `let` block, you could just call it `opt` and it wouldn't get outside the scope of the `let` block.
It might then also be cleaner if you move the `opt = ...` definition from line 80 to the section below (line 94 following) where it's actually used.

examples/train-kernel-parameters/script.jl
```diff
@@ -0,0 +1,234 @@
...
+        x_train, y_train; lab="data", title="i = $(i), Loss = $(round(loss(θ), digits = 4))"
+    )
+    plot!(x_test, sinc; lab="true function")
+    plot!(x_test, f(x_test, x_train, y_train, θ); lab="Prediction", lw=3.0)
+    frame(anim)
+end
+gif(anim, "train-kernel-param.gif"; show_msg=false, fps=15);
+nothing; #hide
+
+# ![](train-kernel-param.gif)
```
st--, line 110: what's the benefit of this vs. having the output from gif() inline ?

examples/train-kernel-parameters/script.jl
```diff
@@ -0,0 +1,234 @@
...
+
+function loss(θ)
+    ŷ = f(x_train, x_train, y_train, θ)
+    return sum(abs2, y_train - ŷ) + exp(θ[4]) * norm(ŷ)
+end
+nothing #hide
+
+# ### Training
+# Setting an initial value and initializing the optimizer: 
+θ = log.([1.1, 0.1, 0.01, 0.001]) # Initial vector
```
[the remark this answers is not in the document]
st--: NB- what do you think of commenting explicitly on this being a mutable vector that will be changed in-place by the optimiser ?

examples/train-kernel-parameters/script.jl
```diff
@@ -0,0 +1,234 @@
...
+nothing; #hide
+
+# ![](train-kernel-param.gif)
+
+# Final loss
+loss(θ)
+
+# ## Using ParameterHandling.jl
+# Alternatively, we can use the [ParameterHandling.jl](https://github.com/invenia/ParameterHandling.jl) package 
+# to handle the requirement that all kernel parameters should be positive.
```
st--, line 117: might be nice to also comment on it allowing arbitrary nested namedtuples for easily accessing parameters without worrying about where in the flat param vector they are, their constraints, etc...

examples/train-kernel-parameters/script.jl
Crown421: This part is from your original example, I am honestly not sure.

examples/train-kernel-parameters/script.jl
```diff
@@ -0,0 +1,234 @@
...
+# to handle the requirement that all kernel parameters should be positive. 
+
+using ParameterHandling
+
+raw_initial_θ = (
+    k1=positive(1.1), k2=positive(0.1), k3=positive(0.01), noise_var=positive(0.001)
+)
+
+flat_θ, unflatten = ParameterHandling.value_flatten(raw_initial_θ)
+nothing #hide
```
st--, line 126: Do you actually need these? I think Literate.jl only splits up code cells at non-code markdown (comments), so all of this would run as a single cell and its return values wouldn't get printed anyways?

examples/train-kernel-parameters/script.jl
```diff
@@ -0,0 +1,234 @@
...
+function loss(θ)
+    ŷ = f(x_train, x_train, y_train, θ)
+    return sum(abs2, y_train - ŷ) + θ.noise_var * norm(ŷ)
+end
+nothing #hide
+
+initial_θ = ParameterHandling.value(raw_initial_θ)
+nothing #hide
+
+# The loss with our starting point :
```
st--, line 149:
```suggestion
# The loss at the initial parameter values:
```

examples/train-kernel-parameters/script.jl
Crown421: I tried a few things, but there seemed to be some leaking. This was my way of fixing that leakage. 
I can take another look.

examples/train-kernel-parameters/script.jl
```diff
@@ -0,0 +1,234 @@
...
+    Optimise.update!(opt, flat_θ, grads)
+end
+nothing #hide
+
+# Final loss
+
+(loss ∘ unflatten)(flat_θ)
+
+# ## Flux.destructure
+# If don't want to write an explicit function to construct the kernel, we can alternatively use the `Flux.destructure` function.
```
st--, line 176:
```suggestion
# If we don't want to write an explicit function to construct the kernel, we can alternatively use the `Flux.destructure` function. 
```

examples/train-kernel-parameters/script.jl
```diff
@@ -0,0 +1,234 @@
...
+
+function loss(θ)
+    ŷ = f(x_train, x_train, y_train, exp.(θ))
+    return sum(abs2, y_train - ŷ) + exp(θ[4]) * norm(ŷ)
+end
+nothing #hide
+
+# ### Training the model
+
+# The loss with our starting point :
```
st--, line 209:
```suggestion
# The loss at our initial parameter values:
```

examples/train-kernel-parameters/script.jl
Crown421: Locally it didn't display correctly.

examples/train-kernel-parameters/script.jl
```diff
@@ -0,0 +1,234 @@
...
+# ## Flux.destructure
+# If don't want to write an explicit function to construct the kernel, we can alternatively use the `Flux.destructure` function. 
+# Again, we need to ensure that the parameters are positive. Note that the `exp` function is now part of the loss function, instead of part of the kernel construction. 
+# We could also use ParameterHandling.jl here, similar to the example above. 
+
+θ = [1.1, 0.1, 0.01, 0.001]
+
+kernel = (θ[1] * SqExponentialKernel() + θ[2] * Matern32Kernel()) ∘ ScaleTransform(θ[3])
+
+p, kernelc = Flux.destructure(kernel);
```
st--, line 184: what's the `p`? what is it needed for? what does it do?

examples/train-kernel-parameters/script.jl
Crown421: You are probably right, and this is some artifact from me moving things around.

examples/train-kernel-parameters/script.jl
st--: It's a nit-pick, but I keep thinking of "lab"="laboratory", my brain really doesn't see how that would map to "label", and it's barely longer, so I'd prefer to spell it out: )

examples/train-kernel-parameters/script.jl
```diff
@@ -0,0 +1,234 @@
...
+nothing #hide
+
+# Final loss
+
+(loss ∘ unflatten)(flat_θ)
+
+# ## Flux.destructure
+# If don't want to write an explicit function to construct the kernel, we can alternatively use the `Flux.destructure` function. 
+# Again, we need to ensure that the parameters are positive. Note that the `exp` function is now part of the loss function, instead of part of the kernel construction. 
+# We could also use ParameterHandling.jl here, similar to the example above.
```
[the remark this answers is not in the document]
Crown421: My impression was that one could do a `kernelc \circ flatten` (or unflatten) construction.

examples/train-kernel-parameters/script.jl
st--: You definitely need to copy the \theta so you don't change the initial values. But
```julia
@benchmark let
    θ = log.(ones(4))
    opt = Optimise.ADAGrad(0.5)
    ...
end
```
should work

examples/train-kernel-parameters/script.jl
Crown421: Probably not needed anymore. Initially I wanted to show that `kernelc(p) == kernel`, but due to a mutability issue, this does not evaluate as `true`, despite them being `==` in almost all respects.

examples/train-kernel-parameters/script.jl
```diff
@@ -0,0 +1,234 @@
...
+# Again, we need to ensure that the parameters are positive. Note that the `exp` function is now part of the loss function, instead of part of the kernel construction. 
+# We could also use ParameterHandling.jl here, similar to the example above. 
+
+θ = [1.1, 0.1, 0.01, 0.001]
+
+kernel = (θ[1] * SqExponentialKernel() + θ[2] * Matern32Kernel()) ∘ ScaleTransform(θ[3])
+
+p, kernelc = Flux.destructure(kernel);
+
+# This returns the `trainable` parameters of the kernel and a function to reconstruct the kernel.
```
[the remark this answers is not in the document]
st--: Flux.trainable isn't exported (and barely documented - the only reference I could find was in https://github.com/FluxML/Flux.jl/blob/4a3483efd8e13437b3d86371723c977fb61c2793/docs/src/models/advanced.md, but that doesn't seem relevant here), so as it is it seems to me to be rather confusing. Maybe
```suggestion
# This returns the trainable `params` of the kernel and a function to reconstruct the kernel.
```
?
or just
```suggestion
# This returns the trainable parameters of the kernel and a function to reconstruct the kernel from these parameters.
```

examples/train-kernel-parameters/script.jl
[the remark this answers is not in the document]
st--: if you think it's something one would actually want to do in practice, it'd be great to expand on it a bit more, but otherwise I would prefer removing the reference so as to not confuse people (like myself reading it now :sweat_smile: )

examples/train-kernel-parameters/script.jl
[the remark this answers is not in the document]
st--: it's probably a reasonable setting to end up in the right local optimum - maybe simpler to just change the very first `ones(4)` to also be [1.1, 0.1, 0.01, 0.001] instead?:)

examples/train-kernel-parameters/script.jl
st--: doesn't really matter, I just think it should be consistent between comments and code:)

st--:
@Crown421 let me know once you've addressed my remaining comments (either through changes or by replying what you doesn't make sense to you / you disagree with:)), and I'll finish off the review!

Crown421:
@st--  I think I addressed all comments now. Apologies for the delay, I also managed to catch covid since we spoke.

examples/train-kernel-parameters/script.jl
```diff
@@ -0,0 +1,234 @@
...
+# the kernel parameters and normalization constant
+
+function f(x, x_train, y_train, θ)
+    k = kernelc(θ[1:3])
+    return kernelmatrix(k, x, x_train) * ((kernelmatrix(k, x_train) + (θ[4]) * I) \ y_train)
+end
+nothing #hide
+
+# We define the loss based on the L2 norm both
+# for the loss and the regularization
```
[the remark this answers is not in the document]
st--: @Crown421 given that there's no such explanation for the other loss definitions (and it's not really about that, anyways), how about just deleting it?
```suggestion
```

st-- (approved):
This looks like a very helpful addition, thanks for contributing!

there's just one remaining comment on the loss comment:) let me know what you want to do with that, then we can merge it

examples/train-kernel-parameters/script.jl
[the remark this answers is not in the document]
Crown421: Ah yes, I think I simplified it in some places, but in a way this is a terrible example, since there is so much duplication going on. Will just delete.

PR #434, opened 2022-02-01
Add more kernels to the benchmark and AD benchmarks

theogf:
Adds a pair more kernels and benchmarks for AD
Right now the issue is with ForwardDiff where I am not sure how to turn the parameters into a single `Array`, if we want to use `ParameterHandling` we first need https://github.com/invenia/ParameterHandling.jl/pull/39 to be merged.


willtebbutt:
Shall we just stick with Zygote for now? I'm not going to have a chance to look at the above ParameterHandling PR for a while, and it needs some proper thought to ensure that it doesn't mess up anything else.

The performance of ForwardDiff is, in general in my (limited) experience, more robust and closely linked to primal performance than Zygote anyway, so I'm less concerned about it (I might be horribly wrong though).

PR #436, opened 2022-02-14
Use the Aqua workflow to test things further

theogf:
This adds tests for the [Aqua.jl](https://juliatesting.github.io/Aqua.jl/dev/) workflow. It tests methods ambiguities, exported non-existing functions, lost parameters and more.

### DIFF
--- a/test/Project.toml
+++ b/test/Project.toml
@@ -1,4 +1,5 @@
 [deps]
+Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
 AxisArrays = "39de3d68-74b9-583c-8d2d-e117c070f3a9"
 Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
 Distances = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
@@ -17,6 +18,7 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
 
 [compat]
+Aqua = "0.5"
 AxisArrays = "0.4.3"
 Compat = "3"
 Distances = "0.10"
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -63,8 +63,10 @@ include("test_utils.jl")
     if GROUP == "" || GROUP == "Aqua"
         @testset "Aqua" begin
             Aqua.test_all(KernelFunctions; ambiguities=false)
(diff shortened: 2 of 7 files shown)


test/runtests.jl
```diff
@@ -59,6 +60,13 @@ const GROUP = get(ENV, "GROUP", "")
 include("test_utils.jl")
 
 @testset "KernelFunctions" begin
+    if GROUP == "" || GROUP == "Aqua"
+        @testset "Aqua" begin
+            Aqua.test_all(KernelFunctions; ambiguities=false)
+            Aqua.test_ambiguities(KernelFunctions; recursive=false)
```
st--, line 66: Should it be
```suggestion
            Aqua.test_ambiguities([KernelFunctions, Base, Core]; recursive=false)
```
to be consistent with what `test_all` would do? (I'm not quite clear why you don't want to do it recursively, though.)

test/Project.toml
```diff
@@ -22,6 +23,7 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
 
 [compat]
+Aqua = "0.5"
```
devmotion, line 26:
```suggestion
Aqua = "0.8"
```

test/runtests.jl
```diff
@@ -64,6 +66,15 @@ const GROUP = get(ENV, "GROUP", "")
 include("test_utils.jl")
 
 @testset "KernelFunctions" begin
+    if GROUP == "" || GROUP == "Aqua"
+        @testset "Aqua" begin
+            Aqua.test_all(KernelFunctions; ambiguities=false)
+            @testset "Method Ambiguity" begin
+                Aqua.test_ambiguities([KernelFunctions, Base, Core]; recursive=false)
+                # Avoid checking extra package ambiguity issues
+            end
```
devmotion, line 75: Based on https://github.com/JuliaTesting/Aqua.jl/issues/77 (and in particular Jameson's comments) I think this should be changed to
```suggestion
            # Ref https://github.com/JuliaTesting/Aqua.jl/issues/77
            Aqua.test_ambiguities(KernelFunctions; recursive=false)
```

theogf:
Well the type piracy is explicit in a comment:
https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/2a6e5e9a7296c1f4624a83e3929fc0daf40d5eb3/src/chainrules.jl#L3

PR #439, opened 2022-02-23
Add optimized implementations for `ZeroKernel` and `ConstantKernel`

devmotion:
This PR adds optimized implementations of `kernelmatrix`, `kernelmatrix!`, `kernelmatrix_diag` and `kernelmatrix_diag!` for `ZeroKernel` and `ConstantKernel`, similar to #438.

### DIFF
--- a/src/basekernels/constant.jl
+++ b/src/basekernels/constant.jl
@@ -15,9 +15,40 @@ See also: [`ConstantKernel`](@ref)
 """
 struct ZeroKernel <: SimpleKernel end
 
-kappa(κ::ZeroKernel, d::T) where {T<:Real} = zero(T)
-
+# SimpleKernel interface
+kappa(::ZeroKernel, ::Real) = false
 metric(::ZeroKernel) = Delta()
+
+# Optimizations
+(::ZeroKernel)(x, y) = false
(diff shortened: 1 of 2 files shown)


src/basekernels/constant.jl
```diff
@@ -73,14 +106,41 @@ end
 
 @functor ConstantKernel
 
-kappa(κ::ConstantKernel, x::Real) = only(κ.c) * one(x)
-
+# SimpleKernel interface
+kappa(κ::ConstantKernel, ::Real) = only(κ.c)
```
theogf, line 110: So this is part of a larger discussion but why change the output type?
What should be the general rule? For example, if I have a F64 lengthscale, even if my input it F32 I'll get a F64 in the end.

devmotion: This PR doesn't change anything - currently `x` is only `true` or `false` since the `Delta` metric is used. So the current implementation just performs the computation `only(k.c) * true`.

Additionally, the `kernelmatrix` as implemented in a previous PR by @willtebbutt does also not take into account the type of the inputs.

theogf (approved):

PR #440, opened 2022-03-10
Minor doc fix

theogf:
Right now it looks like this 
image
This should fix it I think


willtebbutt (approved):
Yup, LGTM in the docs.

PR #444, opened 2022-04-07
add zero(::ColVecs) / zero(::RowVecs) definitions

st--:
Moved here from the AD tests in https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/pull/313

### DIFF
--- a/test/utils.jl
+++ b/test/utils.jl
@@ -28,6 +28,7 @@
         DX[2] = v
         @test DX[2] == v
         @test X[:, 2] == v
+        @test all(zero(DX) .== Ref(zeros(D)))
 
         Y = randn(rng, D, N + 1)
         DY = ColVecs(Y)
@@ -85,6 +86,7 @@
         DX[2] = w
         @test DX[2] == w
         @test X[2, :] == w
+        @test all(zero(DX) .== Ref(zeros(N)))
 
         Y = randn(rng, D + 1, N)
         DY = RowVecs(Y)
--- a/test/utils.jl
+++ b/test/utils.jl
@@ -1,10 +1,3 @@
-function test_zero(DX::Union{ColVecs,RowVecs})
-    zero_DX = zero(DX)
-    @test all(iszero, zero_DX)
-    @test zero_DX isa typeof(DX)
-    @test size(zero_DX.X) == size(DX.X)
-end
-
 @testset "utils" begin
     using KernelFunctions: vec_of_vecs, ColVecs, RowVecs
     rng, N, D = MersenneTwister(123456), 10, 4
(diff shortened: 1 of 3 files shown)


devmotion (commented):
Can you add some tests? Otherwise looks good to me.

theogf:
Look nice but maybe add some unit tests?

test/utils.jl
```diff
@@ -28,6 +28,7 @@
         DX[2] = v
         @test DX[2] == v
         @test X[:, 2] == v
+        @test all(zero(DX) .== Ref(zeros(D)))
```
devmotion, line 31: IMO these are not good and appropriate tests here. We would like to ensure that
- the function returns a `ColVecs`/`RowVecs` object of the same type as the input
- the size of the returned object is as expected
- all elements are zeros

willtebbutt: Would
```suggestion
        @test all(iszero, zero(DX))
        @test zero(DX) isa ColVecs
        @test size(zero(DX)) == size(DX)
```
do @devmotion ?

devmotion: Yeah that would be better. Additionally, one could check the size of the elements. So maybe
```suggestion
        zero_DX = zero(DX)
        @test all(iszero, zero_DX)
        @test zero_DX isa ColVecs
        @test size(zero_DX.X) == size(DX.X)
```

willtebbutt (approved):

test/utils.jl
```diff
@@ -1,3 +1,10 @@
+function test_zero(DX::Union{ColVecs,RowVecs})
+    zero_DX = zero(DX)
+    @test all(iszero, zero_DX)
+    @test zero_DX isa typeof(DX)
+    @test size(zero_DX.X) == size(DX.X)
+end
+
```
devmotion, line 7: Can you move this to https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/master/test/test_utils.jl? Seems like a test helper and is not part of the test set.

devmotion (changes requested):

devmotion:
Is this PR so important that you can't even wait for me to recheck whether I'm happy with the changes and officially approve? :smile: :see_no_evil: :eyes: I really don't think this is how it should work.

devmotion:
This is just ridiculous - all suggestions here were reasonable, responsive, and ended up in the code, in some way or another. I would have approved the PR anyway today, the main point is just that I don't see how merging PRs before reviewers were able to recheck is good behaviour. Well, I'm out here and giving up on you. I've tried to discuss and fix these collaborations long enough and have better things to do if you don't plan to work on it.

st--:
> I've tried to discuss and fix these collaborations long enough and have better things to do if you don't plan to work on it.

Yes, and the same applies in reverse - I've tried to discuss & improve these collaborations too! I'm just really at a loss how to get across how upset I get and how much I struggle with the way you give feedback on PRs. I'd be happy to discuss what we could both do differently, if it's not just "how do _you_ plan to work on it" but also how do _you_ plan to work on it. In the meantime, maybe better if you don't look at my PRs and I don't look at your PRs, and either can be reviewed by other people involved in the packages (@theogf / @willtebbutt / etc. as appropriate). Can we do that?

devmotion:
I only try to provide valuable suggestions, in the same way I get and I would like to get suggestions on my PRs. But since clearly this is not desired, leads to endless unproductive discussions, and I don't see how this will change in the near future, I won't review any of your PRs anymore and hope others, possibly with more time and less frustration, provide you with the same suggestions and comments. In general, I will refrain from any further interaction with you, it's too time consuming and unproductive. As a final possibly not well received comment, maybe you should start considering that the problem is on your side (as well) since reviews, discussions, etc. work and worked fine with basically all other people on Github for me.

PR #447, opened 2022-04-13
Speed up tests

st--:
The `basekernels` tests are currently by far the slowest, taking almost twice the time of the next-slowest test. A quick profiling suggests that the large kernelmatrix computations (by default on the order of 1000x1000) are at least somewhat to blame. The validity of the tests should not depend on how large they are, so this PR significantly reduces that.

### DIFF
--- a/src/test_utils.jl
+++ b/src/test_utils.jl
@@ -94,7 +94,7 @@ function test_interface(
     rng::AbstractRNG, k::Kernel, ::Type{Vector{T}}; kwargs...
 ) where {T<:Real}
     return test_interface(
-        k, randn(rng, T, 1001), randn(rng, T, 1001), randn(rng, T, 1000); kwargs...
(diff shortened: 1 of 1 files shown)


src/test_utils.jl
```diff
@@ -94,18 +94,18 @@ function test_interface(
...
 ) where {T<:Real}
     return test_interface(
-        k, randn(rng, T, 1001), randn(rng, T, 1001), randn(rng, T, 1000); kwargs...
+        k, randn(rng, T, 11), randn(rng, T, 11), randn(rng, T, 13); kwargs...
     )
 end
 
 function test_interface(
-    rng::AbstractRNG, k::MOKernel, ::Type{Vector{Tuple{T,Int}}}; dim_out=1, kwargs...
+    rng::AbstractRNG, k::MOKernel, ::Type{Vector{Tuple{T,Int}}}; dim_out=3, kwargs...
```
st--, line 102: it seemed a bit flawed to only test a multi-output kernel with a single output.

Issue #450, opened 2022-04-13
MaternKernel AD


willtebbutt:
https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/425 makes it possible to AD through the `MaternKernel` by dropping the derivative w.r.t. `\nu`. We tried ensuring that it returns a `NotImplemented`, but Zygote doesn't appear to handle it properly, returning a `nothing`, rather than `NotImplemented`.

devmotion:
Isn't that the convention in Zygote and expected? In Zygote every undefined gradient is `nothing`, so it seems that it would quite breaking if in some cases this convention would be violated.

To me it doesn't seem like a Zygote issue and rather `nothing` seems to be the correct result (regardless of whether one thinks that this `nothing` convention is a good thing or not - personally I think it's not :smile:).

devmotion:
Maybe the only problem is our `test_AD` function here?

willtebbutt:
I'm okay (although disappointed) that this is what Zygote is doing. I think it's good to keep a record of this behaviour, so that we can a) point disgruntled users towards it, and b) point Zygote devs towards it if we feel so inclined.

devmotion:
Ideally, the convention in Zygote would be changed but I don't think this will happen anytime soon (and possibly never).

But it's still not clear to me why Zygote being surprising here should affect if we return `NotImplented` or `NoTangent`. Other CR ADs hopefully handle NotImplemented correctly, and in these cases the choice matters and the implementation in #425 seems wrong.

devmotion:
I'll take back my statement above. Apparently Zygote returns `NotImplemented`: https://github.com/FluxML/Zygote.jl/issues/1204#issuecomment-1098460916

Possibly the issue in #425 was that the pullback was incorrect (missing `NoTangent` for the function itself) - but IIRC Zygote just uses `nothing` for missing derivatives.

PR #452, opened 2022-04-13
Remove `_get_ν`

devmotion:
Fix the bug introduced in #425 (see https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/450).

Removes the workaround.

### DIFF
--- a/src/basekernels/matern.jl
+++ b/src/basekernels/matern.jl
@@ -37,21 +37,8 @@ MaternKernel(; nu::Real=1.5, ν::Real=nu, metric=Euclidean()) = MaternKernel(ν,
 
 @functor MaternKernel
 
-# workaround for Zygote
-# unclear why it's needed but it is fine since it's stated officially that we don't support differentiation with respect to ν
-@inline _get_ν(k::MaternKernel) = only(k.ν)
-function ChainRulesCore.rrule(::typeof(_get_ν), k::T) where {T<:MaternKernel}
-    function _get_ν_pullback(Δ)
-        dν = ChainRulesCore.@not_implemented(
-            "derivatives of `MaternKernel` w.r.t. order `ν` are not implemented."
-        )
-        return NoTangent(), Tangent{T}(; ν=dν, metric=NoTangent())
-    end
-    return _get_ν(k), _get_ν_pullback
-end
-
 @inline function kappa(k::MaternKernel, d::Real)
-    result = _matern(_get_ν(k), d)
+    result = _matern(only(k.ν), d)
     return ifelse(iszero(d), one(result), result)
 end
 
--- a/src/basekernels/matern.jl
+++ b/src/basekernels/matern.jl
@@ -42,8 +42,10 @@ MaternKernel(; nu::Real=1.5, ν::Real=nu, metric=Euclidean()) = MaternKernel(ν,
 @inline _get_ν(k::MaternKernel) = only(k.ν)
 function ChainRulesCore.rrule(::typeof(_get_ν), k::T) where {T<:MaternKernel}
     function _get_ν_pullback(Δ)
-        dν = ChainRulesCore.@not_implemented("derivatives of `MaternKernel` w.r.t. order `ν` are not implemented.")
-        return Tangent{T}(ν=dν, metric=NoTangent())
+        dν = ChainRulesCore.@not_implemented(
+            "derivatives of `MaternKernel` w.r.t. order `ν` are not implemented."
+        )
+        return Tangent{T}(; ν=dν, metric=NoTangent())
     end
     return _get_ν(k), _get_ν_pullback
 end
--- a/src/basekernels/matern.jl
+++ b/src/basekernels/matern.jl
@@ -37,8 +37,16 @@ MaternKernel(; nu::Real=1.5, ν::Real=nu, metric=Euclidean()) = MaternKernel(ν,
 
 @functor MaternKernel
 
+# workaround for Zygote
+# unclear why it's needed but it is fine since it's stated officially that we don't support differentiation with respect to ν
 @inline _get_ν(k::MaternKernel) = only(k.ν)
-ChainRulesCore.@non_differentiable _get_ν(k)  # work-around; should be "NotImplemented" rather than NoTangent
+function ChainRulesCore.rrule(::typeof(_get_ν), k::T) where {T<:MaternKernel}
+    function _get_ν_pullback(Δ)
+        dν = ChainRulesCore.@not_implemented("derivatives of `MaternKernel` w.r.t. order `ν` are not implemented.")
+        return Tangent{T}(ν=dν, metric=NoTangent())
+    end
+    return _get_ν(k), _get_ν_pullback
+end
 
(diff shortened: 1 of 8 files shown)


theogf:
Is it possible to test this?

devmotion:
> So the workaround is tested by virtue of the Zygote AD test on MaternKernel now also passing.

The main problem is that tests passed successfully for https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/452/commits/08437ccc44fa134b86eb447f59d0bdfe651d6acd even though the pullback is clearly incorrect. So I was wondering if the pullback was evaluated at all.

src/basekernels/matern.jl
```diff
@@ -37,8 +37,18 @@ MaternKernel(; nu::Real=1.5, ν::Real=nu, metric=Euclidean()) = MaternKernel(ν,
 
 @functor MaternKernel
 
+# workaround for Zygote
+# unclear why it's needed but it is fine since it's stated officially that we don't support differentiation with respect to ν
```
st--, line 41: Without this workaround, the Zygote AD test fails. You can see this in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/425/commits/30385acdee781a48d5960fe190b87c2e6c43aad0 and the corresponding build https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/runs/6002685848?check_suite_focus=true

st--:
> The main problem is that tests passed successfully for [08437cc](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/commit/08437ccc44fa134b86eb447f59d0bdfe651d6acd) even though the pullback is clearly incorrect.

What was incorrect about the pullback?

> So I was wondering if the pullback was evaluated at all.

It was, I only added it because without Zygote errored, see https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/452#discussion_r849550385

devmotion:
> What was incorrect about the pullback?

It returned the wrong number of outputs (was missing `NoTangent()` derivative for the function itself).

> Without this workaround, the Zygote AD test fails. You can see this in [30385ac](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/commit/30385acdee781a48d5960fe190b87c2e6c43aad0) and [the corresponding build](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/runs/6002685848?check_suite_focus=true)

The problem is just that https://github.com/FluxML/Zygote.jl/blob/95e10218f0187a839ac0539fefcbf39c727bf068/src/lib/array.jl#L30 does not support `NotImplemented`. This should have been fixed in Zygote, introducing workarounds in KernelFunctions is the wrong approach.

st--:
> This should have been fixed in Zygote, introducing workarounds in KernelFunctions is the wrong approach.

Yeah, I completely agree, but unfortunately it _hasn't_ been fixed in Zygote, and I think "workaround + AD is working" is still preferable over "no workaround + AD is broken".

devmotion:
Instead of making a PR to KernelFunctions introducing a workaround you could have made one for Zygote that fixes the issue :thinking:

st--:
> Instead of making a PR to KernelFunctions introducing a workaround you could have made one for Zygote that fixes the issue :thinking:

I managed to figure out how to fix the issue with a workaround in KernelFunctions. That doesn't mean I would immeditely know how to identify & fix the underlying issue upstream. If I _did_ know how to fix it there, I would have, as I have in various other cases. You seem to have an amazing overview as well as deep understanding of a lot of different packages. That's unfortunately not true of everyone, and I'm assuming you're not intending to but you give the impression of being rather disparaging of other people who're not at your level.

devmotion:
IMO usually it is very helpful to open at least Github issues in cases where it is not immediately clear how to fix the issue. Workarounds can be acceptable if they are fixed upstream eventually, but I assume it is unlikely that they are fixed if nobody makes the Zygote developers aware of it.

Generally, I would have given you this advice in your PR but as desired I don't review your PRs anymore.

st--:
@devmotion I've pared it down to a minimal reproducible example: https://github.com/FluxML/Zygote.jl/issues/1204 but I can't comment on anything else, if you've got further thoughts on what the issue is / what's needed to fix it, please comment there.

devmotion:
Test errors seem unrelated? I didn't change anything related to them, I think.

willtebbutt:
> Test errors seem unrelated? I didn't change anything related to them, I think.

Agreed. Maybe we need to relax a tolerance somewhere?

st--:
> > Test errors seem unrelated? I didn't change anything related to them, I think.
> 
> Agreed. Maybe we need to relax a tolerance somewhere?

Yes, I think it's due to the changes in #447; the tests are stochastic (if you just re-run the failed tests, they might pass the next time - I've seen that in #454 as well which has zero code changes). In #447 I reduced the size of the kernel matrix evaluations; the isapprox() call (set with atol=1e-9, rtol=1e-9) is on the arrays as a whole, which does not look at individual elements, but instead checks whether the norm of the difference is within the tolerances, relative to the max of the norm of the individual matrices (see [discussion](https://discourse.julialang.org/t/unit-test-equality-for-arrays/8138/15)). For the random inputs we're using in the tests, the norm of the kernel matrix is linear in the dimensions. So by reducing the size from ~1000 to ~10 we're reducing the norm by a factor ~100, so yes that suggests to me we should increase the rtol correspondingly.

devmotion:
Shouldn't the scale be irrelevant if we use `rtol`? I think it's problematic that we don't use `rtol` in most of the tests but only in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/8e805ef25d745ab391e4ad424e690507e02cd042/src/test_utils.jl#L75-L76. Probably because otherwise tests failed - which IMO, however, would not indicate a problem with `rtol` per se but that the default already was already too low even before the recent changes. I completely agree with https://discourse.julialang.org/t/unit-test-equality-for-arrays/8138/12?u=devmotion that in general you want `rtol` since it is completely unclear what a meaningful `atol` is in general. Specifying a small `atol` as well mainly helps with comparisons that involve exact 0s which would fail otherwise.

st--:
> I think it's problematic that we don't use `rtol` in most of the tests

Yes, I agree. Probably an oversight in the original test code of not realizing that by only setting atol you end up setting rtol=0...

devmotion:
Apparently https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/455 was not sufficient.

devmotion:
Bump.

devmotion:
Bump.

theogf:
Could we keep the global atol and rtol variables?

devmotion:
I don't mind. It just seemed unnecessary to have global variables that are used only in a single place, and clearer (ie easier to spot) if the values are just used as default keyword arguments in the function definition.

devmotion:
@theogf Did you see my last comment? Would you like it to be reverted?

theogf:
I think keeping the global const is a bit more elegant and easily findable since it's in the same file. But again I do not mind, I'll just accept the PR and do what you want 😅

theogf (approved):

PR #455, opened 2022-04-14
looser numeric tolerance for TestUtils.test_interface

st--:
- fixes the docstring
- adds `rtol` specification where `atol` is also specified


src/TestUtils.jl
```diff
@@ -16,18 +16,18 @@ using Test
...
-    test_interface([rng::AbstractRNG], k::Kernel, T::Type{<:AbstractVector}; atol=__ATOL)
+    test_interface([rng::AbstractRNG], k::Kernel, T::Type{<:Real}; atol=__ATOL, rtol=__RTOL)
 
-`test_interface` offers certain types of test data generation to make running these tests
-require less code for common input types. For example, `Vector{<:Real}`, `ColVecs{<:Real}`,
-and `RowVecs{<:Real}` are supported. For other input vector types, please provide the data
-manually. 
+`test_interface` offers automated test data generation for kernels whose inputs are reals.
+This will run the tests for `Vector{<:Real}`, `Vector{Vector{<:Real}}`, `ColVecs{<:Real}`,
+and `RowVecs{<:Real}`. For other input vector types, please provide the data manually.
```
willtebbutt, line 30:
```suggestion
This will run the tests for `Vector{T}`, `Vector{Vector{T}}`, `ColVecs{T}`,
and `RowVecs{T}`. For other input vector types, please provide the data manually.
```
I think this is probably more accurate.

willtebbutt (approved):
I'm happy for this to be merged, subject to resolving my comment, CI passing, and a patch bump being made.

st--:
Looks like this hasn't resolved the test issue :disappointed: should we just bump the default rtol then ?

devmotion:
Yeah that's what I expected unfortunately. According to the logs in the other PR for the failing tests `rtol` was already specified. I think the default values for the `atol = 0` case would make sense, ie. `rtol = sqrt(eps(T))`. (Probably it would also make sense to use a `T`-dependent `atol` but could be done separately if needed at some point.)

src/TestUtils.jl
```diff
@@ -1,7 +1,7 @@
 module TestUtils
 
-const __ATOL = 1e-9
-const __RTOL = 1e-9
+const __ATOL = sqrt(eps(Float64))
```
devmotion, line 3: Is this change required? I would assume changing `rtol` to the default value for `atol = 0` (i.e., `sqrt(eps(T))`) is sufficient. `atol` should only be needed when checking approximate equality involving an exact zero which I assume doesn't happen and I guess is fine with lower tolerances such as 1e-9.

st--: :shrug: maybe not? I didn't think too much about it; we didn't carefully assess the tolerances to check whether atol=1e-9 is neeeded, or 1e-10 is enough... and sqrt(eps(Float64)) is ~2e-8 so not much of a change and it seemed nicely symmetric. Anyways, happy to change it back if (atol, rtol) = (1e-9, 2e-8) is better. What do you think @willtebbutt ?

willtebbutt: To my mind the point of the tests is to pick up on really substantial issues with a kernel implementation (i.e. substantial asymmetry in the kernel matrix, large negative eigenvalues), rather than to test that the numerics, as that presumably has to be a kernel-specific thing. Consequently, I'm happy with the looser `atol` that @st-- proposes.

@st-- is also correct that `1e-9` was chosen for no particular reason other than it seemed to satisfy the above in the cases we tested.

st--: ok, I'm going to add that as a comment though to make it easier to understand in the future

src/TestUtils.jl
```diff
@@ -16,18 +16,18 @@ using Test
         x1::AbstractVector,
         x2::AbstractVector;
         atol=__ATOL,
+        rtol=__RTOL,
     )
 
 Run various consistency checks on `k` at the inputs `x0`, `x1`, and `x2`.
 `x0` and `x1` should be of the same length with different values, while `x0` and `x2` should
 be of different lengths.
```
st--, line 25:
```suggestion

These tests are intended to pick up on really substantial issues with a kernel implementation (e.g. substantial asymmetry
in the kernel matrix, large negative eigenvalues), rather than to test the numerics in detail, which can be kernel-specific.
The default value of `__ATOL` and `__RTOL` is `sqrt(eps(Float64)) ≈ 1.5e-8`, which satisfied this intention in the cases
tested within KernelFunctions.jl itself.
```

st--, line 25:
```suggestion
be of different lengths.

These tests are intended to pick up on really substantial issues with a kernel implementation (e.g. substantial asymmetry
in the kernel matrix, large negative eigenvalues), rather than to test the numerics in detail, which can be kernel-specific.
The default value of `__ATOL` and `__RTOL` is `sqrt(eps(Float64)) ≈ 1.5e-8`, which satisfied this intention in the cases
tested within KernelFunctions.jl itself.
```

st--: doesn't seem to like these suggestions

theogf (approved):

Issue #458, opened 2022-06-23
Type stability of `KernelSum`


simsurace:
```julia
using KernelFunctions
using Test
k = RBFKernel() + RBFKernel() * ExponentialKernel()
@inferred k(0.1, 0.2) # ERROR
@time k(0.1, 0.2)  # 0.000009 seconds (5 allocations: 176 bytes)
```
but e.g.
```julia
using KernelFunctions
using Test
k = RBFKernel() + RBFKernel() * LinearKernel()
@inferred k(0.1, 0.2) # works
```
Similar results hold for `kernelmatrix` and `kernelmatrix_diag`.
I haven't figured out why certain kernels show the problem while others don't.
This may be a variant of https://github.com/JuliaLang/julia/issues/45748 and could be a Julia issue.
The similarity to the referenced Julia issue is that calling `sum` before affects type inference:
```
# restart Julia
using KernelFunctions
using Test
k = RBFKernel() + RBFKernel() * ExponentialKernel()
sum(f -> f(0.1, 0.2), k.kernels)
@inferred k(0.1, 0.2) # works
@time k(0.1, 0.2) # 0.000008 seconds (1 allocation: 16 bytes)
```
Despite this probably being a Julia issue, we may want to patch it anyways, as it can adversely impact AD for certain GP models.
I'll open a PR.

devmotion:
> Despite this probably being a Julia issue, we may want to patch it anyways

What's a possible workaround? Can we just change
https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/ce7923f8252d183768f247256079a2c8279f7b1b/src/kernels/kernelsum.jl#L46
to
```julia
function (κ::KernelSum)(x, y)
    return sum(κ.kernels) do k
        return k(x, y)
    end
end
```
etc.?

simsurace:
I'll double-check, but I think I tried that and it did not solve the issue. I think I'll use the `init` keyword, as the Julia issue seems to be related to the inability to figure that out automatically.

devmotion:
Maybe one has to add a `let` block to avoid the closure type inference bug?

simsurace:
Hmm, seemed to be a false alarm. The issue is that many things seem to work if the function is redefined after module import, but not in a new Julia session immediately after precompilation.

willtebbutt:
@simsurace can we close this now?

simsurace:
[...]

PR #459, opened 2022-06-23
Resolve type stability of evaluation of `KernelSum`

simsurace:
**Summary**
This is an attempt to patch #458. Since this is probably some deeper Julia type inference issue, this solution is likely to be temporary.

**Proposed changes**
For now, I just added tests that expose the type instability. For those kernels, and only or those, the issue can be fixed by including an evaluation of `sum` in the module initialization function.

Re-writing
https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/ce7923f8252d183768f247256079a2c8279f7b1b/src/kernels/kernelsum.jl#L46
either by using `do` or `let` blocks, or passing an `init` argument to `sum`.
None of those made the tests pass, but they work if the function is redefined in the REPL, after the module has been loaded.

**Breaking changes**
None.


src/KernelFunctions.jl
```diff
@@ -133,6 +133,20 @@ function __init__()
...
+        RBFKernel() * (LinearKernel() + ExponentialKernel())
+    )
+        x = rand(4)
+        y = rand(4)
+        sum(f -> f(0.1, 0.2), k.kernels)
+        sum(f -> kernelmatrix(f, x), k. kernels)
+        sum(f -> kernelmatrix(f, x, y), k. kernels)
+        sum(f -> kernelmatrix_diag(f, x), k. kernels)
+        sum(f -> kernelmatrix_diag(f, x, y), k. kernels)
+    end
```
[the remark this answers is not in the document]
simsurace: Yes, that's only a hack to have a CI run that passes the tests, not a generic solution.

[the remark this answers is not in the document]
simsurace: I guess we can wait a bit to see if other ideas come up, and then mark the tests as broken (although I'm not sure how to do that with `@inferred`) and remove the hack in `__init__`.

simsurace:
Hmm, that'd then be similar to the hack I used in https://github.com/JuliaGaussianProcesses/GPLikelihoods.jl/pull/90, right?

willtebbutt:
Indeed!

simsurace:
I'd rather not use that here because the allocations might increase too much. A loopy version is probably also not so great for AD.

willtebbutt:
> I'd rather not use that here because the allocations might increase too much. A loopy version is probably also not so great for AD.

Agreed -- I mean, `map` ought to be completely fine for AD. It would only increase the allocations when outside of the context of reverse-mode AD when the total amount of memory allocated is less of an issue anyway. Just seems weird that there isn't a decent type-stable version of `mapreduce` for tuples anywhere in the ecosystem.

simsurace:
Anyway, this seems to work on nightly, presumably due to https://github.com/JuliaLang/julia/pull/45789, but unfortunately that wasn't backported to 1.8.

willtebbutt:
Ahh I see. Well I'm happy to wait for 1.9 if it will just fix this -- @simsurace is this causing you unacceptable performance issues, or is it something that you could live with for a few months?

simsurace:
Hmm I just did another minimal example on nightly and it didn't pass the test. I guess things might be more complicated after all.
I don't really know how big of a performance issue this is. But hard-coding my kernel in terms of a custom type, where the sum is also hard-coded gave a moderate improvement and made the loss function fully type-stable.

willtebbutt:
Okay. I'll have a think about this. It might be that there's a straightforward, albeit slightly more verbose, way to implement this that we're not seeing..

simsurace:
The following fails on a nightly build on GitHub:
```
using Test

struct FunctionSum{Tf}
    functions::Tf
end

(F::FunctionSum)(x) = sum(f -> f(x), F.functions)

F = FunctionSum((x -> sqrt(x), FunctionSum((x -> x^2, x -> x^3))))
@inferred F(0.1)
```

To the extent that this is similar (JET.jl shows type inference problems that look about the same to the above kernel examples), this is still a Julia issue. Could someone here verify this error?

simsurace:
> Okay. I'll have a think about this. It might be that there's a straightforward, albeit slightly more verbose, way to implement this that we're not seeing..

On a related note, I was thinking about a totally different implementation that would optimize away e.g. multiple uses of the same distance matrix and similar things, using some kind of computational graph. Or is this supposed to be something that the Julia compiler can do on its own? I don't know whether it would help type stability though as long as those kinds of lazy constructions have type inference issues.

willtebbutt:
Aha! This seems to work nicely:
```julia
_sum(f::Tf, x::Tuple) where {Tf} = f(x[1]) + _sum(f, Base.tail(x))

_sum(f::Tf, x::Tuple{Tx}) where {Tf, Tx} = f(x[1])

function kernelmatrix(κ::KernelSum, x::AbstractVector)
    return _sum(Base.Fix2(kernelmatrix, x), κ.kernels)
end

function kernelmatrix(κ::KernelSum, x::AbstractVector, y::AbstractVector)
    return _sum(k -> kernelmatrix(k, x, y), κ.kernels)
end

function kernelmatrix_diag(κ::KernelSum, x::AbstractVector)
    return _sum(Base.Fix2(kernelmatrix_diag, x), κ.kernels)
end

function kernelmatrix_diag(κ::KernelSum, x::AbstractVector, y::AbstractVector)
    return _sum(k -> kernelmatrix_diag(k, x, y), κ.kernels)
end
```

I've basically just implemented a specialised recursive implementation of `sum(f, x)` for the case in which `x` is a `Tuple`.

willtebbutt:
Seems to be fine (according to the unit tests). Generally speaking small tuples + recursion are going to be fine with Zygote I believe.

test/kernels/kernelsum.jl
```diff
@@ -18,4 +18,23 @@
...
+            y = rand(10)
+            @inferred kernelmatrix(k, x)
+            @inferred kernelmatrix(k, x, y)
+            @inferred kernelmatrix_diag(k, x)
+            @inferred kernelmatrix_diag(k, x, y)
+        end
+        @testset for k in (
+            RBFKernel() + RBFKernel() * LinearKernel(),
+            RBFKernel() + RBFKernel() * ExponentialKernel(),
+            RBFKernel() * (LinearKernel() + ExponentialKernel())
```
simsurace, line 35:
```suggestion
            RBFKernel() * (LinearKernel() + ExponentialKernel()),
```

src/kernels/kernelsum.jl
```diff
@@ -43,22 +43,25 @@ end
 
 Base.length(k::KernelSum) = length(k.kernels)
 
-(κ::KernelSum)(x, y) = sum(k(x, y) for k in κ.kernels)
+_sum(f::Tf, x::Tuple) where {Tf} = f(x[1]) + _sum(f, Base.tail(x))
+_sum(f::Tf, x::Tuple{Tx}) where {Tf, Tx} = f(x[1])
```
simsurace, line 47:
```suggestion
_sum(f::Tf, x::Tuple{Tx}) where {Tf,Tx} = f(x[1])
```

test/kernels/kernelsum.jl
```diff
@@ -18,4 +18,23 @@
     end
 
     test_params(k1 + k2, (k1, k2))
+
+    @testset "Type stability" begin
```
willtebbutt, line 22:
```suggestion
    # Regression tests for https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/458
    @testset "Type stability" begin
```
just so that we know where things have come from

willtebbutt (approved):
I'm happy with this subject to CI passing and a bump of the patch version. Once they're both sorted, I'm happy to merge :)

simsurace:
I just did some benchmarks, looks fine to me.
```julia
using BenchmarkTools
using KernelFunctions
using Test
using Zygote

k = RBFKernel() + RBFKernel() * ExponentialKernel()

@inferred k(0., 1.)

x = rand(100)
@btime kernelmatrix($k, $x)
# 207.625 μs (33 allocations: 631.05 KiB)
# 212.833 μs (28 allocations: 630.91 KiB)
@btime Zygote.pullback(kernelmatrix, $k, $x)
# 331.417 μs (402 allocations: 2.54 MiB)
# 335.875 μs (235 allocations: 2.54 MiB)
out, pb = Zygote.pullback(kernelmatrix, k, x)
@btime $pb($out)
# 270.625 μs (545 allocations: 1.09 MiB)
# 199.208 μs (325 allocations: 1.09 MiB)




k = sum(rand() * RBFKernel() ∘ ScaleTransform(rand()) for _ in 1:20)

@inferred k(0., 1.)

x = rand(100)
@btime kernelmatrix($k, $x)
# before: 1.670 ms (258 allocations: 6.08 MiB)
# after:    1.646 ms (258 allocations: 6.08 MiB)
@btime Zygote.pullback(kernelmatrix, $k, $x)
# before: 2.979 ms (3144 allocations: 18.44 MiB)
# after:    2.861 ms (2901 allocations: 18.41 MiB)
out, pb = Zygote.pullback(kernelmatrix, k, x)
@btime $pb($out)
# before: 1.510 ms (3519 allocations: 6.38 MiB)
# after:    1.483 ms (3223 allocations: 6.35 MiB)

```

simsurace:
Closes #458

simsurace:
Ok, seems to be fine after all. Closes #458.

willtebbutt:
This PR looks like it's basically ready to go, other than syncing it up with master. @simsurace is there anything else that you think needs doing?

simsurace:
I don't think so.

willtebbutt:
Cool. Do you have a few minutes to merge in the master branch to this branch so that we can merge this back into master, or are you happy for me to handle it from here?

willtebbutt:
@simsurace have merged in changes from master and re-bumped the patch. Will merge when CI passes.

willtebbutt:
Eurgh. Somehow the build has been broken. Going to need to figure out how before merging.

willtebbutt:
Okay, I'm pretty sure that the failures aren't related to this PR, so I'm going to merge when CI (other than BaseKernels) passes

PR #465, opened 2022-08-14
MOInput Type Stability

willtebbutt:
**Summary**
Makes the type of `out_dim` field of `MOInput` a type parameter of `MOInput`. This has been done in a non-breaking way.

**Proposed changes**

See summary

* ...

https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/405 consides an alternative approach in which the type is constrained to `Int`. This has the disadvantage of being more constrained, and constituting a break change. The potential advantage over the approach used here is slight changes in compile time, but my instinct is that these will be negligible.

**Breaking changes**

None

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
       matrix:
         version:
           - '1'
-          - '1.3'
+          - '1.6'
           - 'nightly'
         os:
           - ubuntu-latest
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "KernelFunctions"
 uuid = "ec8451be-7e33-11e9-00cf-bbf324bd1392"
-version = "0.10.40"
+version = "0.10.41"
 
 [deps]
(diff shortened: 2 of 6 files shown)


theogf:
Maybe add a @inferred test?

willtebbutt:
@theogf done!

theogf (approved):

willtebbutt:
Thanks for the swift review @theogf  -- will merge when CI passes

willtebbutt:
@theogf looks like Others on 1.3 is failing on the doctests -- the formatting of the type printing appears to be different for that version. I don't know how tests were passing previously, because presumably this hasn't changed... 😬 

Any thoughts?

theogf:
Probably the doctestfilter don't catch so many types or something? Didn't we mention dropping 1.3 anyways?

willtebbutt:
Cool. I'm just going to drop CI for 1.3

willtebbutt:
@theogf any idea how I tell github to not require the 1.3 tests to be run?

PR #466, opened 2022-08-19
ChainTransform AD performance

willtebbutt:
**Summary**

The `ChainTransform` has some performance issues on master.

Evidence:
```julia
using BenchmarkTools, KernelFunctions, Zygote

kernel(θ) = with_lengthscale(Matern12Kernel(), 0.5) ∘ PeriodicTransform(θ)

foo(x) = KernelFunctions._map(PeriodicTransform(1 / 5), x)
bar(θ, x) = kernelmatrix(kernel(θ), x)

const x = randn(500);
out, pb = Zygote.pullback(bar, 5.0, x);

Δ = copy(out);
@benchmark $pb($Δ)
```

master:
```julia
BenchmarkTools.Trial: 36 samples with 1 evaluation.
 Range (min … max):  118.074 ms … 234.881 ms  ┊ GC (min … max): 18.20% … 24.77%
 Time  (median):     140.383 ms               ┊ GC (median):    18.11%
 Time  (mean ± σ):   140.637 ms ±  21.592 ms  ┊ GC (mean ± σ):  20.45% ±  4.18%

  █ █▃█ ▃     ██  ▃ ▃
  █▇███▇█▇▇▁▇▁██▇▇█▇█▇▁▇▇▁▁▁▁▁▁▁▇▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▇ ▁
  118 ms           Histogram: frequency by time          235 ms <

 Memory estimate: 128.37 MiB, allocs estimate: 1586367.
```

This branch:
```julia
BenchmarkTools.Trial: 1946 samples with 1 evaluation.
 Range (min … max):  1.979 ms … 10.005 ms  ┊ GC (min … max): 0.00% … 77.81%
 Time  (median):     2.236 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   2.562 ms ±  1.175 ms  ┊ GC (mean ± σ):  8.04% ± 12.87%

  ██▇▆▆▅▃▃▁                                                  ▁
  ███████████▇▄▆▄▅▄▁▁▁▇▇▄▄▄▄▁▄▄▁▁▁▁▄▄▄▁▄▄▅▄▆▄▄▁▄▅▄▆▇▄▅▅▅▅▄▄▄ █
  1.98 ms      Histogram: log(frequency) by time     8.98 ms <

 Memory estimate: 7.70 MiB, allocs estimate: 272.
```

**Proposed changes**

1. use a tuple rather than a vector to contain the things being chained together. This enables type-stable composition.
1. call `_map` rather than `map`, because that's the API

Note that the way I'm testing that this change has been successful is by checking that the number of allocations required to compute the `kernelmatrix`, its forwards-pass and pullback (using Zygote) is invariant to the size of input vector considered. I plan to roll this out more widely in the coming days.

None

**Breaking changes**

This only widens the set of permissible types in the `ChainTransform`, and which one gets used by default. On the basis of this, my inclination is to suggest that we shouldn't consider this breaking, but I might have missed something obvious.

### DIFF
--- a/src/transform/chaintransform.jl
+++ b/src/transform/chaintransform.jl
@@ -1,5 +1,5 @@
 """
-    ChainTransform(ts::AbstractVector{<:Transform})
+    ChainTransform(transforms)
 
 Transformation that applies a chain of transformations `ts` to the input.
 
@@ -19,7 +19,7 @@ julia> map(t2 ∘ t1, ColVecs(X)) == ColVecs(A * (l .* X))
 true
 ```
 """
-struct ChainTransform{V<:AbstractVector{<:Transform}} <: Transform
+struct ChainTransform{V} <: Transform
     transforms::V
 end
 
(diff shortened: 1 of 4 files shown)


src/transform/chaintransform.jl
```diff
@@ -28,23 +28,23 @@ end
 Base.length(t::ChainTransform) = length(t.transforms)
```
theogf, line 29: Do you think we could add a constructor `ChainTransform(transforms...) = ChainTransform(tuple(transforms...))` or that would be breaking with the `ChainTransform(v, \theta)` constructor?

test/test_utils.jl
```diff
@@ -274,3 +274,68 @@ function test_AD(AD::Symbol, k::MOKernel, dims=(in=3, out=2, obs=3))
...
+
+Assesses `constant_allocs_heuristic` for `f`, `Zygote.pullback(f, args...)` and its
+pullback for both of `args1` and `args2`.
+
+`Δ1` and `Δ2` are passed to the pullback associated with `Zygote.pullback(f, args1...)`
+and `Zygote.pullback(f, args2...)` respectively. If left as `nothing`, it is assumed that
+the output of the primal is an acceptable cotangent to be passed to the corresponding
+pullback.
+"""
+function ad_constant_allocs_heuristic(
```
theogf, line 312: I am afraid I don't really understand why we want to check that the number of allocations is equal?

theogf (commented):
Looks good. I also don't think it's breaking as it is

theogf (approved):

test/test_utils.jl
willtebbutt: Ah. I should add more docs then. The logic is:
1. if we've implemented this well, the number of allocations shouldn't change for different input sizes (although the total amount of memory allocated obviously will)
2. a really common way for performance to be bad is some kind of type instability that introduces at least one allocation per kernelmatrix element
3. We can measure the number of allocations really quickly

It's definitely not a sufficient condition for us to know that we've got good performance, but we at least know that the number of allocations is independent of the size of the output of `kernelmatrix`, which rules out a decent number of performance bugs.

Does this make sense?

src/transform/chaintransform.jl
willtebbutt: Hmm yeah, I think that would break that

willtebbutt:
@theogf let me know whether my explanation of the testing is sufficient, and I'll add a docstring + merge

test/test_utils.jl
theogf: Yes this is much clearer thanks! If you put this in the docstring we are good to go :)

willtebbutt:
Will squash + merge when CI passes

PR #467, opened 2022-08-22
AD Performance

willtebbutt:
**Summary**

Extends work from #466 to many (nearly all) of our kernels. Identifies various performance bugs in the process, and fixes many of them.

**Proposed changes**

* ...
Introduces the `test_interface_ad_perf` function in `test/test_utils.jl`. This utilises the allocations-invariant-to-input-size heuristic to check performance.

N/A

**Breaking changes**

None


test/runtests.jl
```diff
@@ -19,7 +20,7 @@ using Compat: only
 
 using KernelFunctions: SimpleKernel, metric, kappa, ColVecs, RowVecs, TestUtils
 
-using KernelFunctions.TestUtils: test_interface
+using KernelFunctions.TestUtils: test_interface, __example_inputs
```
theogf, line 23: If this is you actually exporting, maybe don't make it an internal?

test/test_utils.jl
```diff
@@ -349,6 +349,127 @@ function ad_constant_allocs_heuristic(
...
+"""
+    test_zygote_perf_heuristic(
+        f, name::String, args1, args2, passes, Δ1=nothing, Δ2=nothing
+    )
+
+Executes `ad_constant_allocs_heuristc(f, args1, args2; Δ1, Δ2)` and creates a testset out
+of the results.
+`passes` is a 3-tuple of booleans. `passes[1]` indicates whether the checks should pass
+for `f(args1...)` etc, `passes[2]` for `Zygote.pullback(f, args1...)`, and `passes[3]` the
+pullback generated by `Zygote.pullback(f, args1...)`. This is useful when it is known that
```
theogf, line 364: I don't get the difference between passes[2] and passes[3]

test/test_utils.jl
```diff
@@ -349,6 +349,127 @@ function ad_constant_allocs_heuristic(
...
+function test_zygote_perf_heuristic(
+    f, name::String, args1, args2, passes, Δ1=nothing, Δ2=nothing
+)
+    @testset "$name" begin
+        primal, fwd, pb = ad_constant_allocs_heuristic(f, args1, args2; Δ1, Δ2)
+        if passes[1]
+            @test primal[1] == primal[2]
+        else
+            @test_broken primal[1] == primal[2]
+        end
```
theogf, line 376: I guess you can simplify those as
```suggestion
             @test primal[1] == primal[2] broken=!passes[1]
```

src/transform/ardtransform.jl
```diff
@@ -35,9 +35,10 @@ dim(t::ARDTransform) = length(t.v)
 (t::ARDTransform)(x::Real) = only(t.v) * x
 (t::ARDTransform)(x) = t.v .* x
 
-_map(t::ARDTransform, x::AbstractVector{<:Real}) = t.v' .* x
-_map(t::ARDTransform, x::ColVecs) = ColVecs(t.v .* x.X)
-_map(t::ARDTransform, x::RowVecs) = RowVecs(t.v' .* x.X)
+# Quite specific implementations required to pass correctness and performance tests.
+_map(t::ARDTransform, x::AbstractVector{<:Real}) = x * only(t.v)
```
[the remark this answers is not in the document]
willtebbutt: Should it not be if the `ARDTransform` is being applied to a vector of real numbers? i.e. there's only a single dimension whose relevance should be determined

src/basekernels/rational.jl
```diff
@@ -121,12 +182,32 @@ end
 
 @functor GammaRationalKernel
 
-function kappa(κ::GammaRationalKernel, d::Real)
-    return (one(d) + d^only(κ.γ) / only(κ.α))^(-only(κ.α))
-end
+__grk_kappa(α::Real, γ::Real, d::Real) = (one(d) + d^only(γ) / α)^(-α)
```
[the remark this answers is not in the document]
willtebbutt: Good catch. This is wrong

test/runtests.jl
willtebbutt: I'd rather not export because I don't want to make it part of the official KernelFunctions API. Note that we're not exporting `test_interface` either

test/test_utils.jl
willtebbutt: `passes[2]` is for the output of `Zygote.pullback(f, args1...)`. Let
```julia
out, pb = Zygote.pullback(f, args1)
```
then `passes[3]` is for `pb(out)`.
I'll clarify in the docstring

test/runtests.jl
theogf: Sorry, I meant, if this is part of the TestUtils API, one should probably remove the `__` at the beginning. Using this in the tests would probably feel hacky for someone who does not know what this is.

src/transform/ardtransform.jl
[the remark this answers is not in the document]
theogf: `ARDTransform` is meant to be a vector of (inverse) lengthscales I think.

test/test_utils.jl
theogf: Arf, I thought it was older than this

theogf (approved):

test/test_utils.jl
theogf: Yes looks good now!

theogf:
Don't forget to bump the version.

src/basekernels/fbm.jl
```diff
@@ -50,8 +50,9 @@ _mod(x::RowVecs) = vec(sum(abs2, x.X; dims=2))
 
 function kernelmatrix(κ::FBMKernel, x::AbstractVector)
     modx = _mod(x)
+    modx_wide = modx * ones(1, length(modx)) # ad perf hack -- is unit tested
```
devmotion, line 53: This will cause promotions of e.g. `Float32` types to `Float64`, won't it?

src/basekernels/nn.jl
```diff
@@ -37,6 +37,26 @@ function (κ::NeuralNetworkKernel)(x, y)
     return asin(dot(x, y) / sqrt((1 + sum(abs2, x)) * (1 + sum(abs2, y))))
 end
 
+function kernelmatrix(
+    k::NeuralNetworkKernel, x::AbstractVector{<:Real}, y::AbstractVector{<:Real}
+)
+    return kernelmatrix(k, ColVecs(reshape(x, 1, :)), ColVecs(reshape(y, 1, :)))
```
devmotion, line 43: Maybe one could add an (internal?) function for `ColVecs(reshape(...))`? A while ago, sometimes I also used `ColVecs(permutedims(x))` (IIRC mainly to replace the (recursive) `x'` which can be problematic sometimes, it seemed).

src/basekernels/polynomial.jl
```diff
@@ -68,10 +86,28 @@ function Functors.functor(::Type{<:PolynomialKernel}, x)
     return (c=x.c,), reconstruct_polynomialkernel
 end
 
-kappa(κ::PolynomialKernel, xᵀy::Real) = (xᵀy + only(κ.c))^κ.degree
+@noinline __make_polynomial_kappa(degree) = (c::Real, xᵀy::Real) -> (xᵀy + c)^degree
```
devmotion, line 89: Alternatively one could use a callable struct but I guess it is not necessary to dispatch on the resulting function and this is good enough?

devmotion: Ah maybe a callable struct could be better if one wants to define `rrule` etc.?

src/transform/ardtransform.jl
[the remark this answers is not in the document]
willtebbutt: Sure, but should it not error if the transformation contains more than one element, and the input collection contains only a single element? i.e. that would a dimension mismatch of some kind. It's like trying to apply an ARDTransform with 5 parameters to 3-dimensional data, no?

src/basekernels/fbm.jl
willtebbutt: Good catch

src/kernels/normalizedkernel.jl
```diff
@@ -20,15 +20,17 @@ end
 (κ::NormalizedKernel)(x, y) = κ.kernel(x, y) / sqrt(κ.kernel(x, x) * κ.kernel(y, y))
 
 function kernelmatrix(κ::NormalizedKernel, x::AbstractVector, y::AbstractVector)
-    return kernelmatrix(κ.kernel, x, y) ./
-           sqrt.(
-        kernelmatrix_diag(κ.kernel, x) .* permutedims(kernelmatrix_diag(κ.kernel, y))
-    )
+    x_diag = kernelmatrix_diag(κ.kernel, x)
+    x_diag_wide = x_diag * ones(1, length(y)) # ad perf hack. Is unit tested.
```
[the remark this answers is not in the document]
willtebbutt: Another good catch

src/basekernels/polynomial.jl
willtebbutt: Hmmm maybe? If you think that this is passable though, I think I'd prefer to stick with it

src/basekernels/nn.jl
willtebbutt: Could you provide an example of the kind of thing that you have in mind? Are you imagining a constructor of the form `ColVecs(::AbstractVector)` or something?

devmotion: No, not sure if that would be a good idea. Maybe it is. But I thought rather some simple internal function for such cases to 1) save typing, 2) consistently use the same approach, and 3) be able to change the approach easily (e.g. if we figure out `permutedims` works better for AD etc.).

src/basekernels/polynomial.jl
devmotion: I don't like closures as I agree with https://github.com/SciML/SciMLStyle#closures-should-be-avoided-whenever-possible. I'm OK with changing that in the future if the closure causes problems. (On the other hand, one reason to not use a closure from the start is that - in contrast to a possible need for dispatch later on - potential type inference and performance issues can be difficult to debug and discover.)

src/basekernels/rational.jl
```diff
@@ -69,18 +89,59 @@ struct RationalQuadraticKernel{Tα<:Real,M} <: SimpleKernel
...
+
+kappa(κ::RationalQuadraticKernel, d::Real) = __rq_kappa(only(κ.α), d)
+kappa(κ::_RQ_Euclidean, d²::Real) = __rq_kappa_euclidean(only(κ.α), d²)
 
 metric(k::RationalQuadraticKernel) = k.metric
 metric(::RationalQuadraticKernel{<:Real,<:Euclidean}) = SqEuclidean()
 
+# AD-performance optimisation. Is unit tested.
+function kernelmatrix(k::RationalQuadraticKernel, x::AbstractVector, y::AbstractVector)
+    return __rq_kappa.(only(k.α), pairwise(metric(k), x, y))
```
devmotion, line 107: Seems this line is not tested?

src/basekernels/rational.jl
```diff
@@ -69,18 +89,59 @@ struct RationalQuadraticKernel{Tα<:Real,M} <: SimpleKernel
...
+end
+
+# AD-performance optimisation. Is unit tested.
+function kernelmatrix(k::RationalQuadraticKernel, x::AbstractVector)
+    return __rq_kappa.(only(k.α), pairwise(metric(k), x))
```
devmotion, line 112: Same here?

src/basekernels/rational.jl
```diff
@@ -69,18 +89,59 @@ struct RationalQuadraticKernel{Tα<:Real,M} <: SimpleKernel
...
+end
+
+# AD-performance optimisation. Is unit tested.
+function kernelmatrix_diag(k::RationalQuadraticKernel, x::AbstractVector, y::AbstractVector)
+    return __rq_kappa.(only(k.α), colwise(metric(k), x, y))
```
devmotion, line 117: Not tested?

src/basekernels/rational.jl
```diff
@@ -69,18 +89,59 @@ struct RationalQuadraticKernel{Tα<:Real,M} <: SimpleKernel
...
+end
+
+# AD-performance optimisation. Is unit tested.
+function kernelmatrix_diag(k::RationalQuadraticKernel, x::AbstractVector)
+    return __rq_kappa.(only(k.α), colwise(metric(k), x))
```
devmotion, line 122: Not tested?

src/basekernels/nn.jl
willtebbutt: Ahh okay, so just something like
```julia
_to_row_vec(x::AbstractVector{<:Real}) = reshape(x, 1, :)
```
and
```julia
kernelmatrix(k, ColVecs(_to_row_vec(x)), ColVecs(_to_row_vec(y)))
```
?

src/basekernels/polynomial.jl
willtebbutt: Cool. I'll change to a `struct` :)

src/transform/ardtransform.jl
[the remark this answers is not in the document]
theogf: Ooooh right. I forgot that `_map` acts on the whole data, so it makes sense here.

src/basekernels/nn.jl
devmotion: Or even simpler just
```julia
_to_colvecs(x::AbstractVector{<:Real}) = ColVecs(reshape(x, 1, :))
```
and
```suggestion
kernelmatrix(k, _to_colvecs(x), _to_colvecs(y))
```

willtebbutt:
@devmotion thanks for the careful review -- I believe I've resolved everything that you discussed

willtebbutt:
All CI appears to be passing at last!

devmotion (approved):
LGTM :+1:

PR #472, opened 2022-08-31
GPU Attempt 2

willtebbutt:
**Summary**

Supercedes #470 . I messed up some of my gitconfig stuff and had commits with the wrong username etc. Easiest thing was to start again

**Proposed changes**

* ...

**Breaking changes**

edit:

TODO:
1. sort out codecov on buildkite branch
2. sort out formatting


willtebbutt:
@theogf there are a couple of minor things left to do, but I think this is defintely ready for review

src/TestUtils.jl
```diff
@@ -176,27 +190,108 @@ function test_interface(k::Kernel, T::Type{<:Real}=Float64; kwargs...)
     return test_interface(Random.GLOBAL_RNG, k, T; kwargs...)
 end
 
+const FloatType = Union{Float32, Float64}
```
theogf, line 193: Why not AbstractFloat?

src/KernelFunctions.jl
```diff
@@ -47,6 +47,7 @@ export IndependentMOKernel,
 export tensor, ⊗, compose
 
 using Compat
+using CUDA
```
theogf, line 50: Unless things changed, that's quite a massive dependency no? I saw CuArrays.jl got deprecated, does this mean that CUDA is light now?

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,14 @@
+steps:
```
theogf, line 1: Noice!

src/TestUtils.jl
```diff
@@ -176,27 +190,108 @@ function test_interface(k::Kernel, T::Type{<:Real}=Float64; kwargs...)
...
+    end
 end
 
+function test_gpu_against_cpu(rng::AbstractRNG, T::Type, k::Kernel, data_type::Type)
+    _, x1, x2, _ = example_inputs(rng, data_type)
+    test_gpu_against_cpu(k, T, x1, x2)
+end
+
+_to_cpu(x::CuArray{<:Real}) = Array(x)
+_to_cpu(x::ColVecs{<:Real, <:CuMatrix}) = ColVecs(_to_cpu(x.X))
```
theogf, line 282: Can't we just overload `cpu` on our type? (or `cpu` is part of Flux? I don't remember...)

src/basekernels/matern.jl
```diff
@@ -77,7 +77,7 @@ end
 
 Matern32Kernel(; metric=Euclidean()) = Matern32Kernel(metric)
 
-kappa(::Matern32Kernel, d::Real) = (1 + sqrt(3) * d) * exp(-sqrt(3) * d)
+kappa(::Matern32Kernel, d::T) where {T<:Real} = (1 + T(sqrt(3)) * d) * exp(-T(sqrt(3)) * d)
```
theogf, line 80: Can't T(sqrt(3)), or at least sqrt(3) be precalculated?

src/basekernels/matern.jl
```diff
@@ -108,7 +108,7 @@ end
 
 Matern52Kernel(; metric=Euclidean()) = Matern52Kernel(metric)
 
-kappa(::Matern52Kernel, d::Real) = (1 + sqrt(5) * d + 5 * d^2 / 3) * exp(-sqrt(5) * d)
+kappa(::Matern52Kernel, d::T) where {T<:Real} = (1 + T(sqrt(5)) * d + 5 * d^2 / 3) * exp(-T(sqrt(5)) * d)
```
theogf, line 111: Same here with sqrt(5)

src/basekernels/wiener.jl
```diff
@@ -62,22 +62,22 @@ function (::WienerKernel{1})(x, y)
     X = sqrt(sum(abs2, x))
     Y = sqrt(sum(abs2, y))
     minXY = min(X, Y)
-    return 1 / 3 * minXY^3 + 1 / 2 * minXY^2 * euclidean(x, y)
+    return minXY^3 / 3 + minXY^2 * euclidean(x, y) / 2
```
theogf, line 65: Nice catches here!

theogf (commented):
Wow that's an amazing load of good work. My only concern is that a lot of the code consist in getting pairwise done right and in a better world this should be done in an external package. Especially because of the new CUDA dependency...

src/basekernels/matern.jl
devmotion: One could use
```suggestion
function kappa(::Matern32Kernel, d::Real)
    x = sqrt3 * d
    return (1 + x) * exp(-x)
end
```
and load `IrrationalConstants.sqrt3` in src/KernelFunctions.jl

src/basekernels/matern.jl
devmotion: Unfortunately, IrrationalConstants doesn't define `sqrt5` (maybe should be added?) but one could still improve it a bit: ```suggestion
function kappa(::Matern52Kernel, d::Real)
    x = sqrt(oftype(d, 5))
    return (1 + x + 5 * d^2 / 3) * exp(-x)
end
```
This would also be a bit safer as it also works if `d` is not a floating point number.

File: src/KernelFunctions.jl (hunk 2 above)
Comment (devmotion) [reply src/KernelFunctions.jl:50]: As we already depend on Requires, maybe it could be hidden behind a `@require`?

Comment (devmotion) [reply src/KernelFunctions.jl:50]: Alternatively, maybe the lightweight `GPUArraysCore` (https://github.com/JuliaGPU/GPUArrays.jl/tree/master/lib/GPUArraysCore) is sufficient?

File: src/TestUtils.jl (hunk 4 above)
Comment (devmotion) [reply src/TestUtils.jl:282]: Could we use/define `Adapt.adapt`? And let users pass the types they want to adapt it to?
According to the README of Adapt, we probably would want to define
```julia
Adapt.adapt_structure(to, x::ColVecs{<:Real}) = ColVecs(adapt(to, x.X))
Adapt.adapt_structure(to, x::RowVecs{<:Real}) = RowVecs(adapt(to, x.X))
```
And then one could use e.g. `adapt(CuArray, x)` or `adapt(Array, x)` where `x` could be a `ColVecs{<:Real, Matrix{<:Real}}`/`RowVecs{<:Real,Matrix{<:Real}}` or a `ColVecs{<:Real,<:CuMatrix}`/`RowVecs{<:Real,<:CuMatrix}` (for the other direction). If the testing function allows to specify the desired type (such as `Array` or `CuArray`) users/developers could pass it and we don't have to deal with it in KernelFunctions here, maybe?

src/basekernels/fbm.jl
```diff
@@ -48,9 +48,19 @@ _mod(x::AbstractVector{<:AbstractVector{<:Real}}) = sum.(abs2, x)
...
 _mod(x::RowVecs) = vec(sum(abs2, x.X; dims=2))
 
+_ones(el_type, x::AbstractVector, sz...) = ones(el_type, sz...)
+function _ones(
+    el_type,
+    x::Union{CuVector{<:Real}, ColVecs{<:Real, <:CuMatrix}, RowVecs{<:Real, <:CuMatrix}},
+    sz...,
+)
+    return CUDA.ones(el_type, sz...)
+end
```
devmotion, line 58: Could we use `adapt` here?

src/distances/cuda.jl
```diff
@@ -0,0 +1,123 @@
+
+const CuSubArray{T} = SubArray{T, <:Any, <:CuArray}
+const CuColVecs{T} = ColVecs{T, <:Union{CuMatrix{T}, CuSubArray{T}}}
+const CuRowVecs{T} = RowVecs{T, <:Union{CuMatrix{T}, CuSubArray{T}}}
+
+const _CuVector{T} = Union{CuVector{T}, SubArray{T, 1, <:CuArray}}
```
devmotion, line 6: Do we need to restrict it to these types? Could we just define our `pairwise` (IIRC we own it?) in this way for generic `AbstractVector`/`ColVecs`/`RowVecs`, and only forward the `Vector{<:Real}`/`ColVecs{<:Real,Matrix{<:Real}}`/... to Distances?

devmotion (commented):
@willtebbutt Do you think it will take some time to sort out all remaining comments and issues in this PR? If so, we could add only the buildkite configuration as a first step to fix CI errors and add GPU-specific tests in a follow-up PR, similar to https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/pull/335.

willtebbutt:
> @willtebbutt Do you think it will take some time to sort out all remaining comments and issues in this PR? If so, we could add only the buildkite configuration as a first step to fix CI errors and add GPU-specific tests in a follow-up PR, similar to https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/pull/335.

That would be ideal. I thought I was going to have time to follow up on this immediately, but it's actually going to be a little while. If you could open a PR like the one you made to AbstractGPs, that would be ideal.

PR #475, opened 2022-09-21
Wct/linear spline

willtebbutt:
**Summary**

Found this interesting kernel in the Probabilistic numerics textbook.

**Proposed changes**

Add the LinearSplineKernel

* ...

Doing more work to find a more general version of the spline kernel.

**Breaking changes**

None

@theogf does our AD correctness testing infrastructure currently permit me to test with `Vector{Float64}` inputs to `kernelmatrix` etc, rather than `Matrix{Float64}`? If not, I'll not bother testing. If it does, I'll add a line.


devmotion:
> Found this interesting kernel in the Probabilistic numerics textbook.

Haha, I just started to read through that book a bit some weeks ago :smile: 

I'm not sure though if we should restrict "linear spline kernel" to this specific example. It seems a quick Google lookup yields many publications with different "linear spline kernels" (e.g., https://www.diva-portal.org/smash/get/diva2:716642/FULLTEXT01.pdf, https://people.vcu.edu/~ymotai/Publications/journals/TNNLS2015.pdf, https://gpy.readthedocs.io/en/deploy/_modules/GPy/kern/src/spline.html). Maybe we could generalize the definition to accommodate for (some of) these other forms as well?

Could you also add it to the docuementation?

willtebbutt:
Hmmmm I'm going to have to leave this open for a while. I don't currently have access to Wahba's 1980 paper in which I assume the stuff that Hennig talks about is discussed in more depth. At the minute, I'm not able to determine what the bounds are on the domain of the kernels. If anyone has access and can figure it out, I would greatly appreciate it.

st--:
@willtebbutt do you mean this one: [Spline bases, regularization, and generalized cross validation for solving approximation problems with large quantities of noisy data](https://pages.stat.wisc.edu/~wahba/TRarchive/TRsEARLY2/tr597_1.pdf) ?

willtebbutt:
I'm not going to find time to finish this off, so I'm going to close.

PR #476, opened 2022-09-26
Extend lengthscale tip

willtebbutt:
**Summary**

It was unclear to a Stheno.jl user how to do ARD stuff -- see [this issue](https://github.com/JuliaGaussianProcesses/Stheno.jl/issues/237). This PR extends our lengthscale tip to multiple dimensions.

**Proposed changes**

Extend lengthscale tip to cover ARD and factor analysis cases.

I considered adding something in Stheno, but I didn't want to dilute the docs.
I also considered not adding the factor analyis example, since it is so closely-related to the ARD example, it felt important to include.

**Breaking changes**

None


docs/src/userguide.md
```diff
@@ -21,6 +21,19 @@ For example, a squared exponential kernel is created by
     ```
     [`with_lengthscale`](@ref) also works with vector-valued lengthscales for ARD.
     Check the [Input Transforms](@ref input_transforms) page for more details.
+    Per-dimension lengthscales for automatic relevance determination work in a similar manner:
+    ```julia
+      inverse_length_scales = [1.0, 2.0]
+      k = SqExponentialKernel() ∘ ARDTransform(inverse_length_scales)
+      k = compose(SqExponentialKernel(), ARDTransform(inverse_length_scales))
+    ```
```
devmotion, line 29: I thought commonly one would just use `with_lengthscale`? ARD is actually mentioned two lines above :slightly_smiling_face:

willtebbutt: Huh. I hadn't realised that is a thing! Maybe I should just add an example with a vector of lengthscales to the tip then, rather than extending substantially? And maybe change the title to "How do I set the lengthscale(s)?"?

devmotion: I think these are good suggestions.

willtebbutt: Have tweaked -- let me know if you think it works.

devmotion (approved):

PR #478, opened 2022-09-28
Fix Matern Grads

willtebbutt:
**Summary**

We've been seeing test failures for AD with the `MaternKernel`. This fixes these.

**Proposed changes**

Modifies the implementation of the `kappa` function for the `MaternKernel` to utilise `if ... else ... end` instead of `ifelse.` The reason that this fixes the problem is that, if `iszero(d)`, the branch which calls the bessel function etc (rather that the one that just returns `1`) has some `NaN`s and some `Inf`s floating around in it.
The `ifelse` branch evaluates this branch-with-`NaN`s, and when Zygote is performing the reverse pass, this `NaN` somehow gets propagated backwards.
Conversely, `if ... else ... end` entirely avoids computing any `NaN`s, thereby avoiding the problem.

This has the potential downside of yielding worse performance under Zygote, the `MaternKernel` performance was already quite bad, so I don't view this as a problem (it might not even make it much worse tbh).

The solution that's defnitely going to work + be performant here would be a custom rule, but I don't have the time, and I don't know that this can be improved without sorting out the `besselk` `rrule` anyway.

**Breaking changes**

Not breaking.

### DIFF
--- a/src/basekernels/matern.jl
+++ b/src/basekernels/matern.jl
@@ -37,14 +37,15 @@ MaternKernel(; nu::Real=1.5, ν::Real=nu, metric=Euclidean()) = MaternKernel(ν,
 
 @functor MaternKernel
 
-@inline function kappa(k::MaternKernel, d::Real)
-    result = _matern(only(k.ν), d)
-    return ifelse(iszero(d), one(result), result)
-end
+@inline kappa(k::MaternKernel, d::Real) = _matern(only(k.ν), d)
 
 function _matern(ν::Real, d::Real)
-    y = sqrt(2ν) * d
-    return exp((one(d) - ν) * logtwo - loggamma(ν) + ν * log(y) + log(besselk(ν, y)))
+    if iszero(d)
+        return one(d)
+    else
+        y = sqrt(2ν) * d
+        return exp((one(d) - ν) * logtwo - loggamma(ν) + ν * log(y) + log(besselk(ν, y)))
+    end
 end
 
 metric(k::MaternKernel) = k.metric
--- a/src/basekernels/matern.jl
+++ b/src/basekernels/matern.jl
@@ -44,7 +44,8 @@ function _matern(ν::Real, d::Real)
         return one(d)
     else
         y = sqrt(2ν) * d
-        return exp((one(d) - ν) * logtwo - loggamma(ν) + ν * log(y) + log(besselk(ν, y)))
+        b = log(besselk(ν, y))
+        return exp((one(d) - ν) * oftype(y, logtwo) - loggamma(ν) + ν * log(y) + b)
     end
 end
 
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 Manifest.toml
 coverage/
 .DS_store
+dev
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "KernelFunctions"
 uuid = "ec8451be-7e33-11e9-00cf-bbf324bd1392"
-version = "0.10.45"
+version = "0.10.46"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"


willtebbutt:
Codecov is taking a little while to update. This is ready for review.

src/basekernels/matern.jl
```diff
@@ -37,14 +37,15 @@ MaternKernel(; nu::Real=1.5, ν::Real=nu, metric=Euclidean()) = MaternKernel(ν,
...
-    result = _matern(only(k.ν), d)
-    return ifelse(iszero(d), one(result), result)
-end
+@inline kappa(k::MaternKernel, d::Real) = _matern(only(k.ν), d)
 
 function _matern(ν::Real, d::Real)
-    y = sqrt(2ν) * d
-    return exp((one(d) - ν) * logtwo - loggamma(ν) + ν * log(y) + log(besselk(ν, y)))
+    if iszero(d)
+        return one(d)
```
devmotion, line 44: I think this is problematic as it very easily creates type stability issues.

devmotion: Maybe something like
```suggestion
        return one(float(Base.promote_typeof(ν, d)))
```
would be sufficient.

devmotion: Maybe a better approach though (also e.g. regarding the computation in the other branch) would be to define
```julia
_matern(ν::Real, d::Real) = _matern(promote(ν, d)...)
function _matern(ν::T, d::T) where {T<:Real}
....
```

src/basekernels/matern.jl
```diff
@@ -37,14 +37,15 @@ MaternKernel(; nu::Real=1.5, ν::Real=nu, metric=Euclidean()) = MaternKernel(ν,
...
+    else
+        y = sqrt(2ν) * d
+        return exp((one(d) - ν) * logtwo - loggamma(ν) + ν * log(y) + log(besselk(ν, y)))
```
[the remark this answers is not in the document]
devmotion: Also maybe one could be a bit more careful with undesired promotions of the irrational and use `oftype(y, logtwo)` instead of just `logtwo`.

src/basekernels/matern.jl
willtebbutt: Good point

src/basekernels/matern.jl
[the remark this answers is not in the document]
willtebbutt: Also a good point. Have implemented.

src/basekernels/matern.jl
willtebbutt: Promotion + ForwardDiff / ReverseDiff don't appear to play nicely. Neither of the above patterns seems to work unfortunately.

willtebbutt: I don't believe the new implementation to be any _worse_ than the previous, so unless you strongly object, I would like to overlook this for now.

src/basekernels/matern.jl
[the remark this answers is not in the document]
devmotion: OK :slightly_smiling_face:

src/basekernels/matern.jl
devmotion: Ah that's unfortunate and a bit surprising. I'm fine with keeping the PR as it is and to improve it at some point.

devmotion (commented):
Looks good to me, I just think it would be good to add a test for it that fails on the master branch.

src/basekernels/matern.jl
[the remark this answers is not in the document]
devmotion: BTW `besselkx` should be fixed on ForwardDiff/ReverseDiff by https://github.com/JuliaDiff/DiffRules.jl/pull/79.

willtebbutt:
> Looks good to me, I just think it would be good to add a test for it that fails on the master branch.

Well I implemented this change _to_ fix a test on master, so is it not sufficient that CI now passes?

devmotion (approved):
Ah I didn't realize that :smile: Somehow I thought you ran into issues in some external example. If it fixes CI on master that's sufficient, of course :+1:

Issue #479, opened 2022-10-03
Delta type restrictions


willtebbutt:
@theogf do you know why our `Delta` metric is constrained to work for numeric-ish types? Should it not work for anything that has `==` defined on it?

cc @glennmoy

theogf:
Yeah no idea why I did this... This should be changed...

willtebbutt:
Cool. We should add a test when we do. I think we should probably do a more general audit, but we shouldn't block this change in the short-term.

@molet do you have any other examples of kernels where you'd have thought things should just work out-of-the-box with non-numeric data types and they don't?

glennmoy:
`_wrap` is another example (where we initially came across this)

Restricted to `AbstractVector{<:Real}` and `AbstractMatrix{<:Real}`

https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/d425983cd734c5c9c2e8c7573a97491ea5ecaa52/src/transform/selecttransform.jl#L31-L32

molet:
Although there might be some exceptions like kernels in `constant.jl`, I think in general it is a good approach to require numeric types for most of the kernels.

So far, I successfully used `FunctionTransform` for many situations when I needed to convert a non-numeric type to a numeric one to be able to use specific numeric type based kernels. I think this is a good practice.

Therefore, I would rather focus on making some transform functions more general and let them take non-numerical types: `SelectTransform` and `FunctionTransform` would definitely need this feature (so we might want to change this [line](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/master/src/transform/functiontransform.jl#L26) as well to be more general).

glennmoy:
> Although there might be some exceptions like kernels in `constant.jl`, I think in general it is a good approach to require numeric types for most of the kernels.

If this is the desired interface it would be good to document it.

theogf:
> Although there might be some exceptions like kernels in constant.jl, I think in general it is a good approach to require numeric types for most of the kernels.

I think that's already true with the `kappa` function. But for things like `Transform` and `Metric` types, this should not be the case. We aim at being as agnostic as possible to the type of data.

willtebbutt:
I believe that this has been resolved via #480  so will close this issue. Please let me know if there's anything missing.

PR #480, opened 2022-10-09
Ensure accepted input types are general where possible

willtebbutt:
**Summary**

As discussed in #479 , some of our kernels / transforms etc is overly strict in terms of the types that it accepts, prompting users to find workaround. The goal of this is to resolve these problems and add some tests to prevent regression.

**Proposed changes**
- Some type restrictions have been removed / generalised.
- Some tests have been added.

N/A

**Breaking changes**

N/A

### DIFF
--- a/src/distances/delta.jl
+++ b/src/distances/delta.jl
@@ -2,9 +2,8 @@
 struct Delta <: Distances.UnionPreMetric end
 
 (dist::Delta)(a, b) = a == b
-Base.@propagate_inbounds function (dist::Delta)(
-    a::AbstractArray, b::AbstractArray
-)
+
+Base.@propagate_inbounds function (dist::Delta)(a::AbstractArray, b::AbstractArray)
     @boundscheck if length(a) != length(b)
         throw(
             DimensionMismatch(
--- a/src/distances/delta.jl
+++ b/src/distances/delta.jl
@@ -1,9 +1,9 @@
 # Delta is not following the PreMetric rules since d(x, x) == 1
 struct Delta <: Distances.UnionPreMetric end
 
-(dist::Delta)(a::Number, b::Number) = a == b
+(dist::Delta)(a, b) = a == b
 Base.@propagate_inbounds function (dist::Delta)(
-    a::AbstractArray{<:Number}, b::AbstractArray{<:Number}
+    a::AbstractArray, b::AbstractArray
 )
     @boundscheck if length(a) != length(b)
         throw(
--- a/src/transform/selecttransform.jl
+++ b/src/transform/selecttransform.jl
@@ -28,7 +28,7 @@ _maybe_unwrap(x::AbstractArray{<:Any,0}) = x[]
 _map(t::SelectTransform, x::ColVecs) = _wrap(view(x.X, t.select, :), ColVecs)
 _map(t::SelectTransform, x::RowVecs) = _wrap(view(x.X, :, t.select), RowVecs)
 
-_wrap(x::AbstractVector{<:Real}, ::Any) = x
-_wrap(X::AbstractMatrix{<:Real}, ::Type{T}) where {T} = T(X)
+_wrap(x::AbstractVector, ::Any) = x
+_wrap(X::AbstractMatrix, ::Type{T}) where {T} = T(X)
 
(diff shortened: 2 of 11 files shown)


src/distances/delta.jl
```diff
@@ -1,9 +1,9 @@
 # Delta is not following the PreMetric rules since d(x, x) == 1
 struct Delta <: Distances.UnionPreMetric end
 
-(dist::Delta)(a::Number, b::Number) = a == b
+(dist::Delta)(a, b) = a == b
 Base.@propagate_inbounds function (dist::Delta)(
-    a::AbstractArray{<:Number}, b::AbstractArray{<:Number}
+    a::AbstractArray, b::AbstractArray
```
devmotion, line 6: I wonder if we should demand that the inputs are of the same (element) type, i.e., `(dist::Delta)(a::T, b::T) where {T} = a == b` and `(dist::Delta)(a::AbstractArray{T}, b::AbstractArray{T}) where {T}` (maybe in the latter case even the dimension should be equal?)?

src/transform/selecttransform.jl
```diff
@@ -28,7 +28,7 @@ _maybe_unwrap(x::AbstractArray{<:Any,0}) = x[]
 _map(t::SelectTransform, x::ColVecs) = _wrap(view(x.X, t.select, :), ColVecs)
 _map(t::SelectTransform, x::RowVecs) = _wrap(view(x.X, :, t.select), RowVecs)
 
-_wrap(x::AbstractVector{<:Real}, ::Any) = x
-_wrap(X::AbstractMatrix{<:Real}, ::Type{T}) where {T} = T(X)
+_wrap(x::AbstractVector, ::Any) = x
```
devmotion, line 31: Is that always correct and what does this change imply? E.g., now `x::ColVecs` is also covered here?

src/distances/delta.jl
```diff
@@ -1,9 +1,9 @@
...
 )
```
devmotion, line 7: Formatter is unhappy here:
```suggestion
Base.@propagate_inbounds function (dist::Delta)(a::AbstractArray, b::AbstractArray)
```

test/transform/functiontransform.jl
```diff
@@ -26,6 +26,15 @@
         end
     end
 
+    @testset "String input" begin
+        f = x -> x * "hello"
+        t = FunctionTransform(f)
+        x = [randstring(rng) for _ in 1:3]
+        x′ = map(t, x)
```
devmotion, line 33: Maybe use `y` here? I don't know how to type that accent and I'm not even sure it is possible to do it easily on my keyboard. So basically I always have to copy-paste.

devmotion:
```suggestion
        y = map(t, x)
```

test/transform/functiontransform.jl
```diff
@@ -26,6 +26,15 @@
...
+        @test all([t(x[n]) == x′[n] for n in eachindex(x)])
+        @test all([f(x[n]) == x′[n] for n in eachindex(x)])
```
devmotion, line 35:
```suggestion
        @test all([t(x[n]) == y[n] for n in eachindex(x)])
        @test all([f(x[n]) == y[n] for n in eachindex(x)])
```

src/distances/delta.jl
willtebbutt: Insisting on the same element types would be breaking, and I'm not sure it's strictly necessary. In order for this kernel to be defined, we really just need `==` to be defined between whatever is passed in.

src/transform/selecttransform.jl
willtebbutt: This is fine -- the point of `_wrap` is to turn `x` into an `AbstractVector` (i.e. `ColVecs` or `RowVecs`) if it's a matrix, and leave `x` alone if it's a vector. This relaxation of the type constraints just means that this can happen for more types -- it was overly restricted before.

src/distances/delta.jl
willtebbutt: Yeah. Weird that it won't let me commit in the GH interface. Will make the change locally.

devmotion (approved):
Looks good to me. I think the coverage comments by codecov are wrong?

PR #496, opened 2023-03-21
Small clarification in train-kernel-parameters notebook

st--:
**Summary**

When I tried to walk someone through the train-kernel-parameters notebook, the notation was confusing, so we fixed it.


willtebbutt (approved):
Good catch

PR #499, opened 2023-04-18
Add check_args and drop 1.3

theogf:
**Summary**
Similarly to Distributions.jl we do not necessarily want to check for argument correctness all the time.

**Proposed changes**

This adds a `check_args` keyword to all constructors that require it to deactivate the check for the correctness for the arguments.
Additionally this drops 1.3 for 1.6 because I am lazy to write `kwarg=kwarg` (appears in 1.5 or 1.6 can't remember0

* ...

We could define a reparametrization of the parameters but we already discussed that it should be the duty of the user to do this.

**Breaking changes**
None! Crazy right?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "KernelFunctions"
 uuid = "ec8451be-7e33-11e9-00cf-bbf324bd1392"
-version = "0.10.53"
+version = "0.10.54"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
@@ -35,4 +35,4 @@ SpecialFunctions = "0.8, 0.9, 0.10, 1, 2"
 StatsBase = "0.32, 0.33"
 TensorCore = "0.1"
 ZygoteRules = "0.2"
-julia = "1.3"
+julia = "1.6"
--- a/src/basekernels/constant.jl
+++ b/src/basekernels/constant.jl
@@ -98,8 +98,8 @@ See also: [`ZeroKernel`](@ref)
 struct ConstantKernel{Tc<:Real} <: SimpleKernel
     c::Vector{Tc}
 
-    function ConstantKernel(; c::Real=1.0)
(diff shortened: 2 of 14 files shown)


willtebbutt (approved):
Approved subject to CI etc. I'd also like an approval from either @devmotion or @st-- before merging

devmotion:
I'm a bit worried that this might cause performance regressions, or at least is suboptimal, based on the fixes that were needed in Distributions: https://github.com/JuliaStats/Distributions.jl/pull/1492

willtebbutt:
Good point @devmotion . @theogf could you check some simple examples with Zygote?

theogf:
> I'm a bit worried that this might cause performance regressions, or at least is suboptimal, based on the fixes that were needed in Distributions: [JuliaStats/Distributions.jl#1492](https://github.com/JuliaStats/Distributions.jl/pull/1492)

Then shouldn't I just reuse the `@check_args` from Distributions.jl ?

theogf:
@devmotion do you think that would be enough?

devmotion:
Did you do compare performance with Zygote?

devmotion:
That's a quite noticeable regression. Do we know what exactly causes it?

theogf:
For completeness I added the constructor 
```
    function OldLinearKernel(c::Real; check_args=true)
        check_args && @old_check_args(LinearKernel, c, c >= zero(c), "c ≥ 0")
        return new{typeof(c)}([c])
    end
```
And these are the results
```
function h(x)
    k = OldLinearKernel(x[1])
    sum(k.c)
end
function i(x)
    k = OldLinearKernel(x[1]; check_args=false)
    sum(k.c)
end

@btime Zygote.gradient($h, $x)  # 6.086 μs (65 allocations: 2.58 KiB)
@btime Zygote.gradient($i, $x) # 10.404 μs (91 allocations: 3.38 KiB)
```

theogf:
Looking back at this, how about using `CheckArgs.jl` ?

Issue #506, opened 2023-05-09
Sum of independent kernels


martincornejo:
Following this [discourse discussion](https://discourse.julialang.org/t/gps-sum-of-independent-kernels/98542). Currently, there is no building block to sum independent Kernels, analog to `KernelTensorProduct` but with addition instead of multiplication:

For inputs $x = (x_1,\dots,x_n)$ and  $x' = (x_1',\dots,x_n')$, the independent sum of kernels $k_1, \dots, k_n$:

$$ k(x, x'; k_1, \dots k_n) = \sum_{i=1}^n k_i(x_i, x_i') $$

theogf:
It sounds like a reasonable composite to add, especially since the alternative is pretty ugly using `SelectTransform`, is there a standardized name for this kind of kernels? `KernelTensorSum`? `KernelDimensionwiseSum` ?

devmotion:
> What are the requirements for a fully functional kernel that can be used in AbstractGPs?

I guess you can mainly copy `KernelTensorProduct` and replace multiplication with addition.

> would an abstract type KernelTensor for both KernelTensorProduct and KernelTensorSum make sense?

Not in an initial version IMO (and maybe not at all). I would add a separate type, similar to how we distinguish between `KernelSum` and `KernelProduct`.

martincornejo:
One technical question. Is it always the case that if you compare the same input, the correlation should be 1?

```julia
julia> k = SqExponentialKernel();

julia> x = 0;

julia> k(x,x)
1.0
```

Independently adding the kernels results into the following behavior:
```julia
julia> k1 = SqExponentialKernel();

julia> k2 = ExponentialKernel();

julia> k = KernelTensorSum(k1, k2)
Tensor sum of 2 kernels:
        Squared Exponential Kernel (metric = Distances.Euclidean(0.0))
        Exponential Kernel (metric = Distances.Euclidean(0.0))

julia> x = zeros(2);

julia> k(x,x)
2.0

```
So, should the kernel take the `mean` instead of `sum` so the correlation is _normalized_?

theogf:
No it does not have to be!
I would not "normalize" cause that might be something unexpected from the user side. The scaling should be dealt with each kernel individually.

PR #507, opened 2023-05-10
Add `KernelTensorSum`

martincornejo:
Adressess https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/506. Implements a new kernel composition `KernelTensorSum` and related `⊕` operator.

The naming should be discussed since the meaning of `KernelTensorSum` might not be appropriate. Suggestions are welcome.


src/kernels/kerneltensorsum.jl
```diff
@@ -0,0 +1,102 @@
...
+function (kernel::KernelTensorSum)(x, y)
+    if !(length(x) == length(y) == length(kernel))
+        throw(DimensionMismatch("number of kernels and number of features
+are not consistent"))
+    end
+    return sum(k(xi, yi) for (k, xi, yi) in zip(kernel.kernels, x, y))
+end
+
+function validate_domain(k::KernelTensorSum, x::AbstractVector)
+    return dim(x) == length(k) ||
```
willtebbutt, line 60: It would probably be helpful if this provided a little more information to the user. In particular, could you please also print out what `dim(x)` and `length(k)` are in the error message?

src/kernels/kerneltensorsum.jl
```diff
@@ -0,0 +1,102 @@
...
+end
+
+function kernelmatrix(k::KernelTensorSum, x::AbstractVector)
+    validate_domain(k, x)
+    return mapreduce(kernelmatrix, +, k.kernels, slices(x))
+end
+
+function kernelmatrix(k::KernelTensorSum, x::AbstractVector, y::AbstractVector)
+    validate_domain(k, x)
+    return mapreduce(kernelmatrix, +, k.kernels, slices(x), slices(y))
```
willtebbutt, line 71: Is it clear why codecov is complaining about this line not getting hit in the tests? I would have thought the interface tests would pick this up 🤔

src/kernels/kerneltensorsum.jl
```diff
@@ -0,0 +1,102 @@
...
+function KernelTensorSum(kernel::Kernel, kernels::Kernel...)
+    return KernelTensorSum((kernel, kernels...))
+end
+
+@functor KernelTensorSum
+
+Base.length(kernel::KernelTensorSum) = length(kernel.kernels)
+
+function (kernel::KernelTensorSum)(x, y)
+    if !(length(x) == length(y) == length(kernel))
```
willtebbutt, line 52: Should this test be conducted by `validate_domain`, or is it doing something slightly different? Do we additionally need a three-argument version of `validate_domain` which can check multiple input vectors?

src/kernels/kerneltensorsum.jl
```diff
@@ -0,0 +1,102 @@
...
+           error("number of kernels and groups of features are not consistent")
+end
+
+function kernelmatrix(k::KernelTensorSum, x::AbstractVector)
+    validate_domain(k, x)
+    return mapreduce(kernelmatrix, +, k.kernels, slices(x))
+end
+
+function kernelmatrix(k::KernelTensorSum, x::AbstractVector, y::AbstractVector)
+    validate_domain(k, x)
```
willtebbutt, line 70: It doesn't look like we check that `y` has the correct dimension here. Maybe we need to extend `validate_domain` (related to above comment)

src/kernels/kerneltensorsum.jl
```diff
@@ -0,0 +1,102 @@
...
+    return mapreduce(kernelmatrix, +, k.kernels, slices(x), slices(y))
+end
+
+function kernelmatrix_diag(k::KernelTensorSum, x::AbstractVector)
+    validate_domain(k, x)
+    return mapreduce(kernelmatrix_diag, +, k.kernels, slices(x))
+end
+
+function kernelmatrix_diag(k::KernelTensorSum, x::AbstractVector, y::AbstractVector)
+    validate_domain(k, x)
```
willtebbutt, line 80: Same thing as before about not checking `y`.

willtebbutt (commented):
Thanks for opening this PR. It's looking pretty good -- there are a few details I would like addressed before merging though

src/kernels/kerneltensorsum.jl
martincornejo: Slightly different, since here, `x`, and `y` could be tuples or vectors and in `validate_domain` they are vector of vectors

src/kernels/kerneltensorsum.jl
```diff
@@ -0,0 +1,107 @@
...
+end
+
+@functor KernelTensorSum
+
+Base.length(kernel::KernelTensorSum) = length(kernel.kernels)
+
+function (kernel::KernelTensorSum)(x, y)
+    if !((nx = length(x)) == (ny = length(y)) == (nkernerls = length(k)))
+        throw(DimensionMismatch("number of kernels ($nkernels) and number of features 
+(x=$nx, y=$ny) are not consistent"))
```
willtebbutt, line 54: Formatting error?

src/kernels/kerneltensorsum.jl
```diff
@@ -0,0 +1,107 @@
...
+function KernelTensorSum(kernel::Kernel, kernels::Kernel...)
+    return KernelTensorSum((kernel, kernels...))
+end
+
+@functor KernelTensorSum
+
+Base.length(kernel::KernelTensorSum) = length(kernel.kernels)
+
+function (kernel::KernelTensorSum)(x, y)
+    if !((nx = length(x)) == (ny = length(y)) == (nkernerls = length(k)))
```
willtebbutt, line 52: It looks like `nkernerls` is a typo here? Should be `nkernels`?

src/kernels/kerneltensorsum.jl
willtebbutt: Much better. Thanks!

martincornejo:
I've renamed `KernelTensorSum` to `KernelIndependentSum` since in my opinion it better informs what it is. I could also revert this commit, do you have any opinions on the naming?

martincornejo:
Hi, is there anything holding this PR? Let me know if I should do anything more from my side. Thanks!

src/kernels/overloads.jl
```diff
@@ -1,7 +1,10 @@
+function ⊕ end
```
devmotion, line 1: This seems too generic to be defined and exported from KernelFunctions. Is it not part of TensorCore or some other lightweight interface package? We would also a non-Unicode alias, as for other keyword arguments and functions.

devmotion:
I would have chosen `KernelTensorSum`, for consistency with `KernelTensorProduct`, similar to how we use the names `KernelSum`/`KernelProduct`. But I don't have a strong opinion.

willtebbutt:
Sorry for not getting around to reviewing this.

> I've renamed KernelTensorSum to KernelIndependentSum since in my opinion it better informs what it is. I could also revert this commit, do you have any opinions on the naming?

Could you elaborate on why this is a better name? This actually does worry me, because I value consistency with `KernelTensorProduct`.

src/kernels/overloads.jl
martincornejo: I thought about that, but `⊕` is not part of TensorCore. It is a help constructor for the new `KernelTensorSum`/`KernelIndependentSum`, so there non-Unicode function is already available.

martincornejo:
> I would have chosen `KernelTensorSum`, for consistency with `KernelTensorProduct`, similar to how we use the names `KernelSum`/`KernelProduct`. But I don't have a strong opinion.

> Could you elaborate on why this is a better name? This actually does worry me, because I value consistency with `KernelTensorProduct`.

That was my initial idea as well, but without the context of  `KernelTensorProduct`, it is difficult to understand what `KernelTensorSum` is in itself. Also, is the use of the term "tensor sum" used properly in this context? If not, we would be trading off consistency for "formal correctness".

But I'm completely open about this, both are valid alternatives. I'll leave the decision up to you.

src/kernels/overloads.jl
martincornejo: Suggestions wellcome on how to improve this.

willtebbutt:
That's fair. I agree that I'm not entirely sure that the use of the term is correct, but I still think my preference is consistency + an issue suggesting that both be renamed when we next create a breaking change / a proposal to deprecate the tensor names.

src/kernels/overloads.jl
devmotion: There's no non-Unicode alternative similar to `+`, `*`, or `tensor` yet as far as I can tell?

martincornejo: Ah! I see, you're right

martincornejo: So, actually, `Kronecker.jl` does define `⊕` (`kronekersum`) but is currently an optional "Requires" dependency. That poses a problem.

devmotion: Which workaround?

martincornejo: I've added a non-Unicode alias for `⊕`: `kernel_sum`

theogf:
I only found one reference to the term "Tensor sum" here: https://www.degruyter.com/document/doi/10.1515/spma-2019-0009/html?lang=en

src/kernels/overloads.jl
martincornejo: > Which workaround?

https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/ef6d4591b36194fca069d8bc7ae8c1e2ee288080/src/matrix/kernelkroneckermat.jl#L1-L3

devmotion: Ah I see! IMO that's not a workaround though 😅 I think generally it is good practice to qualify calls and not load packages with `using XX` (in a package, scripts and REPL are different).

martincornejo: Resolve?

devmotion: > This seems too generic to be defined and exported from KernelFunctions.

This problem is not fixed yet, is it?

martincornejo: But since `TensorCore.jl`  does not define `⊕`, what should we do? Here are the [packages that use  `⊕`](https://juliahub.com/ui/Search?q=%E2%8A%95&type=symbols). `Kronecker.jl` is one, but I guess we do not want to add this as a dependency, only to re-use the symbol.

devmotion: We should make a PR to TensorCore. I think the operator should not be owned by KernelFunctions.

martincornejo: https://github.com/JuliaMath/TensorCore.jl/pull/10

Issue #504, opened 2023-05-02
Efficient AD for the simulation of gradients


FelixBenning:
## Introduction: How would you simulate gradients

If you want to simulate the gradient of a random function $Z$, it turns out that you simply need to take gradient of the covariance funcion, as (w.l.o.g. $Z$ is centered)
```math
\text{Cov}(\nabla Z(x), Z(y)) = \mathbb{E}[\nabla Z(x) Z(y)] = \nabla_x \mathbb{E}[Z(x)Z(y)] = \nabla_x C(x,y)
```

And similarly
```math
\text{Cov}(\nabla Z(x), \nabla Z(y)) = \nabla_x\nabla_y C(x,y)
```

For stationary covariance functions $C(x,y) = C(x-y)$ this simplifies to
```math
\begin{aligned}
\nabla_x C(x,y)  &= \nabla C(x-y)\\
\nabla_x\nabla_y C(x,y) &= -\nabla^2 C(x-y)
\end{aligned}
```

So if we consider the multivariate random function $T_1(x) = (Z(x), \nabla Z(x))$, then its covariance kernel is given by
```math
\begin{aligned}
C_{T_1}(x,y) 
&= \begin{pmatrix}
     C(x,y) & \nabla_y C(x,y)^T\\
     \nabla_x C(x,y) & \nabla_x \nabla_y C(x,y)
\end{pmatrix}\\
&\overset{\mathllap{\text{stationary}}}= 
\begin{pmatrix}
     C(x-y) & -\nabla C(x-y)^T\\
     \nabla C(x-y) & -\nabla^2C(x-y)
\end{pmatrix}\\
&\overset{\mathllap{\text{isotrope}}}=
 \begin{pmatrix}
     C(d) & -f'\bigl(\frac{\|d\|^2}{2}\bigr)d^T\\
     f'\bigl(\frac{\|d\|^2}{2}\bigr) d & -\Bigl[f''\bigl(\frac{\|d\|^2}2\bigr)dd^T + f'\bigl(\frac{\|d\|^2}2\bigr) \mathbb{I}\Bigr]
\end{pmatrix}
\end{aligned}
```
where we use $d=x-y$ and $C(d) = f\bigl(\frac{\|d\|^2}2\bigr)$ in the last equation.

## Performance Considerations

[...]

PR #508, opened 2023-05-17
Introduce DiffPt for the covariance function of derivatives

FelixBenning:
**Summary**

This is a minimal implementation to enable the simulation of gradients (and higher order derivatives) of GPs (see also https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/504)

**Proposed changes**

For a covariance kernel k of GP Z, i.e.
```julia
k(x,y) # = Cov(Z(x), Z(y)),
```
a `DiffPt` allows the differentiation of Z, i.e.	
```julia
k(DiffPt(x, partial=1), y) # = Cov(∂₁Z(x), Z(y))
```
for higher order derivatives partial can be any iterable, i.e.
```julia
k(DiffPt(x, partial=(1,2)), y) # = Cov(∂₁∂₂Z(x), Z(y))
```

the code for this feature is extremely minimal but allows the simulation of arbitrary derivatives of Gaussian Processes. It only contains

* `DiffPt`
* an `_evaluate(::T, x::DiffPt, y::DiffPt) where {T<: Kernel}` function which calls `partial` functions that take the derivatives.

This is the implementation with the smallest footprint but not the most performant. What essentially happens here is the simulation of the multivariate GP $f = (Z, \nabla Z)$ which is a $d+1$ dimensional GP if $Z$ is a univariate GP with input dimension $d$. Due to the ["no multi-variate kernels" design philosophy of KernelFunctions.jl](https://juliagaussianprocesses.github.io/KernelFunctions.jl/stable/design/#inputs_for_multiple_outputs) we are forced to calculate the entries of the covariance matrix one-by-one. It would be more performant to calculate the entire matrix in one go using backward diff for the first pass and forward diff for the second derivative.

It might be possible to somehow specialize on ranges to get back this performance. But it is not completely clear how. Since we do not call
```julia
k.(1:d, 1:d)
```
which could easily be caught by specializing on `broadcast` but in reality we do something like
```julia
k.([(x, 1),...(x,d)], [(y,1),...(y,d)])
```
And this is still not quite true as we consider all pairs of these lists and not just a zip.

**Breaking changes**

None.


src/diffKernel.jl
```diff
@@ -0,0 +1,108 @@
...
+
+#=
+This is a hack to work around the fact that the `where {T<:Kernel}` clause is
+not allowed for the `(::T)(x,y)` syntax. If we were to only implement
+```julia
+	(::Kernel)(::DiffPt,::DiffPt)
+```
+then julia would not know whether to use
+`(::SpecialKernel)(x,y)` or `(::Kernel)(x::DiffPt, y::DiffPt)`
+```
```
FelixBenning, line 85: To avoid this hack, no kernel type T should implement
```julia
	(::T)(x,y)
```
and instead implement
```julia
	_evaluate(k::T, x, y)
```
Then there should be only a single
```julia
	(k::Kernel)(x,y) = _evaluate(k, x, y)
```
which all the kernels would fall back to.

This ensures that `_evaluate(k::T, x::DiffPt{Dim}, y::DiffPt{Dim}) where T<:Kernel` is always
more specialized and called.

FelixBenning: but this is a much more intrusive change so to not blow up the lines changed, I did not do this yet.

FelixBenning: this is also why "detect ambiguities" fails right now

Crown421:
I don't think `ForwardDiff` should be an explicit dependency for KernelFunctions. To me this would make more sense as an [extension](https://julialang.org/blog/2023/04/julia-1.9-highlights/#package_extensions) (which might also allow for different implementations, i.e. Enzyme forward).

src/KernelFunctions.jl
```diff
@@ -125,6 +127,7 @@ include("chainrules.jl")
 include("zygoterules.jl")
 
 include("TestUtils.jl")
+include("diffKernel.jl")
```
devmotion, line 130: Kernels are contained in the kernel subfolder.

src/diffKernel.jl
```diff
@@ -0,0 +1,81 @@
+using OneHotArrays: OneHotVector
+import ForwardDiff as FD
+import LinearAlgebra as LA
+
+""" 
+	DiffPt(x; partial=())
```
devmotion, line 6: Why is a separate type needed? Wouldn't it be better to use the existing input formats for multi-output kernels?

src/diffKernel.jl
```diff
@@ -0,0 +1,81 @@
...
+a DiffPt allows the differentiation of Z, i.e.	
+```julia
+	k(DiffPt(x, partial=1), y) # = Cov(∂₁Z(x), Z(y))
+```
+for higher order derivatives partial can be any iterable, i.e.
+```julia
+	k(DiffPt(x, partial=(1,2)), y) # = Cov(∂₁∂₂Z(x), Z(y))
+```
+"""
+struct DiffPt{Dim}
```
devmotion, line 21: Is it necessary to hardcode the dimension?

src/diffKernel.jl
```diff
@@ -0,0 +1,81 @@
...
+    pos # the actual position
+    partial
```
devmotion, line 23: Fields with abstract types lead to bad performance.

src/diffKernel.jl
```diff
@@ -0,0 +1,81 @@
...
+end
+
+DiffPt(x; partial=()) = DiffPt{length(x)}(x, partial) # convenience constructor
```
devmotion, line 26: The result can't be inferred here.

src/diffKernel.jl
```diff
@@ -0,0 +1,81 @@
...
+
+"""
+Take the partial derivative of a function `fun`  with input dimesion `dim`.
+If partials=(i,j), then (∂ᵢ∂ⱼ fun) is returned.
+"""
+function partial(fun, dim, partials=())
```
devmotion, line 32: IMO a simpler API would be preferable. Why a tuple instead of just separate arguments?

src/diffKernel.jl
```diff
@@ -0,0 +1,81 @@
...
+    if !isnothing(local next = iterate(partials))
```
devmotion, line 33: This is related to the comment about the function signature above: no check is needed if the syntax is just `partial(f, dim, i)` and `partial(f, dim, i, j)`.

src/diffKernel.jl
devmotion: I don't think we should switch to `_evaluate(k, x, y)`. Instead of changing all other kernels, I think you should just create a single wrapper of existing kernels.

src/diffKernel.jl
```diff
@@ -0,0 +1,81 @@
...
+    local f(x, y) = partial(t -> k(t, y), dim, partials_x)(x)
+    return (x, y) -> partial(t -> f(x, t), dim, partials_y)(y)
+end
+
+"""
+	_evaluate(k::T, x::DiffPt{Dim}, y::DiffPt{Dim}) where {Dim, T<:Kernel}
+
+implements `(k::T)(x::DiffPt{Dim}, y::DiffPt{Dim})` for all kernel types. But since
+generics are not allowed in the syntax above by the dispatch system, this
+redirection over `_evaluate` is necessary
```
devmotion, line 58: See my other comment, I think this should not be done and a simple wrapper would be sufficient.

src/diffKernel.jl
```diff
@@ -0,0 +1,81 @@
...
+    end
+    return fun
+end
+
+"""
+Take the partial derivative of a function with two dim-dimensional inputs,
+i.e. 2*dim dimensional input
+"""
+function partial(k, dim; partials_x=(), partials_y=())
+    local f(x, y) = partial(t -> k(t, y), dim, partials_x)(x)
```
devmotion, line 49: `local` is not needed?

src/diffKernel.jl
devmotion: Moreover, based on the code below `dim` can be removed.

src/diffKernel.jl
```diff
@@ -0,0 +1,81 @@
...
+"""
+Take the partial derivative of a function `fun`  with input dimesion `dim`.
+If partials=(i,j), then (∂ᵢ∂ⱼ fun) is returned.
+"""
+function partial(fun, dim, partials=())
+    if !isnothing(local next = iterate(partials))
+        idx, state = next
+        return partial(
+            x -> FD.derivative(0) do dx
+                fun(x .+ dx * OneHotVector(idx, dim))
```
devmotion, line 37: Why `OneHotVector`? Can't you just use
```suggestion
                y = similar(x)
                copyto!(y, x)
                y[idx] += dx
                fun(y)
```
Using `OneHotVector` does not seem to save any allocations here.

FelixBenning: I wanted to avoid mutation as autodiff seems to be unstable if you do use mutation. In principle the code above is probably equivalent - as long as autodiff does not stumble over it.

FelixBenning: This is also the only reason OneHotArrays was used. If that truly works with mutation this dependency could be removed.

Project.toml
```diff
@@ -8,10 +8,12 @@ Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
 CompositionsBase = "a33af91c-f02d-484b-be07-31d278c5ca2b"
 Distances = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
 FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
+ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 Functors = "d9f16b24-f501-4c13-a1f2-28368ffc5196"
 IrrationalConstants = "92d709cd-6900-40b7-9082-c6be49f344b6"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 LogExpFunctions = "2ab3a3ac-af41-5b50-aa03-7779005ae688"
+OneHotArrays = "0b1bfda6-eb8a-41d2-88d8-f5af5cad476f"
```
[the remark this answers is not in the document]
FelixBenning: https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/508/files#r1196306623

src/KernelFunctions.jl
FelixBenning: Well it is not a new kernel - it is extending the functionality of all kernels which is why I did not put it there. But maybe it makes sense to put it there anyway.

src/diffKernel.jl
FelixBenning: Where can I find the existing input formats? https://juliagaussianprocesses.github.io/KernelFunctions.jl/stable/design/#inputs_for_multiple_outputs here it says that you would use `Tuple{T, Int}` and while you could in principle convert a `partial` tuple to an `Int`, you reall don't want to.

I mean even if we are gracious and start counting with zero to make this less of a mess:
```
() -> 0 # no derivative
(1) -> 1 # partial derivative in direction 1
(2) -> 2
...
(dim) -> dim # partial derivative in direction dim
(1,1) -> dim+1 # twice partial derivative in direction 1 
(1,2) -> dim +2
...
(k, j) -> k * dim + j
```
So the user would have to do this conversion from carthesian coordinates to linear coordinates. And then the implementation would revert this transformation back from linear coordinates to cartesian coordinates. Cartesian coordinates with ex ante unknown tuple length that is.

That all seems annoying without being really necessary. But sure - you could do it.

src/diffKernel.jl
FelixBenning: this would very much be true if you wanted to do the conversion from cartesian coordinates to linear coordinates. I actually started this way until I thought screw it this is much more annoying than it is worth. Later on the OneHotVector needs to know the dimension. If we can get rid of the one hot vector, then no the dimension is not needed (but then you definitely can not use linear coordinates)

src/diffKernel.jl
FelixBenning: what do you mean by that?

src/diffKernel.jl
FelixBenning: what do you mean separate arguments? The tuple is of variable length if it is empty (no derivative), then function is returned as is. If it contains one element or is a number, then take the partial derivative in that direction. If it contains two elements do two partial derivatives, if it contains n element do n partial derivatives

src/diffKernel.jl
FelixBenning: I guess - felt right to explicitly say that this function is only temporary and is immediately going to be transformed again

Project.toml
```diff
@@ -8,10 +8,12 @@ Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
 CompositionsBase = "a33af91c-f02d-484b-be07-31d278c5ca2b"
 Distances = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
 FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
+ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
```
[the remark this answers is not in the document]
FelixBenning: yeah maybe this should be a plugin somehow

src/diffKernel.jl
FelixBenning: The ugly thing about wrapping existing kernels is, that you essentially say: This is a different gaussian process model.

I.e.

```julia
f = GP(MaternKernel())
```
is a different (non-differentiable) model while

```julia
g = GP(DiffWrapper(MaternKernel()))
```
is differentiable. But fundamentally the matern kernel implies that the gaussian process should always be $\lfloor \nu\rfloor$-differentiable. So `f` _ought_ to be differentiable.

And with this abstraction it is. I.e. you use

 ```julia
 x = 1:10
 fx = f(x)
 y = rand(fx)
 ```
 
 to simulate `f` at points `1:10`. If you now wanted to simulate its gradient at point `0` too, you would just have to modify its input
  ```julia
 x = [DiffPt(0, partial=1), 1:10... ]
 fx = f(x)
y0_grad, y... =  rand(fx)
 ```
and `y0_grad` would be the gradient in `0` as expected.

FelixBenning: Also note how this abstraction lets you mix and match normal points and `DiffPt`s

src/diffKernel.jl
devmotion: Don't you just need `X \times \{0, \ldots, dim\}` to represent all desired partials?

src/diffKernel.jl
devmotion: > autodiff seems to be unstable if you do use mutation.

Definitely not the case for ForwardDiff.

For immutable arrays you would probably just dispatch/branch above and then use `y = Base.setindex(x, x[idx] + dx, idx)` - a bit unfortunate that Base does not support this for `Array`. But then again usually we only work with `Array`, so this special case is not even needed (definitely not initially).

src/diffKernel.jl
devmotion: There's no way for the compiler to figure out what type this function returns.

src/diffKernel.jl
FelixBenning: I mean kernels themselves have abstract type inputs, no? I assume that the struct would be jited once the type is known (just like the kernel methods)

Project.toml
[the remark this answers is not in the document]
devmotion: IMO it's not ready for proper use. But hopefully it will at some point.

src/diffKernel.jl
devmotion: > Also note how this abstraction lets you mix and match normal points and DiffPts

You can perform exactly the same computations with a wrapper type.

> The ugly thing about wrapping existing kernels is, that you essentially say: This is a different gaussian process model.

I don't view it this way. The wrapper does not change the mathematical model, it just allows you to query derivatives as well.

src/diffKernel.jl
FelixBenning: And this is a bit tricky

```math
\begin{aligned}
f: \mathbb{R}^d\to  \mathbb{R}\\
\nabla f(x) \in \mathbb{R}^d\\
\nabla^2 f(x) \in \mathbb{R}^{d^2}
\end{aligned}
```
So $(f(x), \nabla f(x), \nabla^2 f(x)) \in \mathbb{R} \times \mathbb{R}^d \times \mathbb{R}^{d\times d} = \mathbb{R}^{1+d + d^2}$

So more generally $(f(x), \nabla f(x), ..., f^{(n)}) \in \mathbb{R}^m$ with 
```math
m=\sum_{k=0}^{n} d^k = \frac{n(n+1)}2
```
In essence a variable length tuple carries more information than just the longest tuple. Plus there is no longest tuple.

src/diffKernel.jl
FelixBenning: The goal is to have an arbitrarily differentiable `f` here, so manually defining `partial(f, i, j, k, l, ....)` is not really an option. What would be possible is to use varargs

```julia
function partial(fun, partials...)
     ....
     return partial(
          ...
          Base.rest(partials, state)...
```

But this only really adds a ton of splatting and unsplatting.

src/diffKernel.jl
devmotion: No, it's not superfluous as the wrapped kernel has a different API (i.e., you have to provide different types/structure of inputs) that is more inconvenient to work with if you're not interested in evaluating those. There's a difference - but IMO it's not a mathematical one but rather regarding user experience.

As you've already noticed I think it's also just not feasible to extend every implementation of differentiable kernels out there without making implementations of kernels more inconvenient for people that do not want to work with derivatives. So clearly separating these use cases seems simpler to me from a design perspective.

The wrapper would be a `Kernel` as well, of course, so if the existing implementations in KernelFunctions such as sums etc. are written as generally as they were intended no new definitions for compositions are needed. You would add them only if there is a clear performance gain that outweighs code complexity.

src/diffKernel.jl
FelixBenning: the `similar(x)` thing breaks if `x` is just a value not a vector :(

src/diffKernel.jl
FelixBenning: no - you do not have to provide different types/structures of inputs. So it does not make it more inconvenient for people not interested in gradients. That is the entire point of specializing on `DiffPt` the kernel function still works on everything that is not a `DiffPt`

src/diffKernel.jl
FelixBenning: I tried replacing this function with these two:
```julia
function partial(fun, dim, idx)
	return x -> FD.derivative(0) do dx
		fun(x .+ dx * OneHotVector(idx, dim))
	end
end

function partial(fun, dim, indices...)
	idx, state = iterate(indices)
	return partial(partial(fun, dim, idx), Base.rest(indices, state)...)
end
```

but I can not get my own tests to work then. I will have to take a look after this weekend. I unfortunately won't have time in the next couple days.

Crown421:
I have been playing around with the ideas in PR, and realized that to make this work there are some open questions. 

The first issue is that `kernelmatrix` is specialized for `SimpleKernels` and `MOKernels`, which would require some additional thought and changes to make work. 
For `SimpleKernels`  just adding the `_evaluate` method is insufficient, as `kernelmatrix` uses `pairwise(metric(...), ...`. In principle one could think about going deeper, and start extending all necessary methods for `Distances.jl`, but at this point the question is whether it is worth it. 

At this point a wrapper might be easier, because then the only thing needed are some additional methods. At this point `GP(DiffWrapper(Kernel))` would indeed be a different object, since it might use be less specialized methods for the `kernelmatrix`. 

Additionally,  GP `gd` that expresses the derivative of some GP `g` is not quite the same object. At least for the exact GP posteriors defined in `AbstractGPs.jl`, each instance stores the Cholesky decomposition `C` as well as `C\y` for the undifferentiated input.kernel, which can then be efficiently re-used each time the `mean` or `var` are computed. 
To get the variance `var(::GP, ::DiffPt)` we can't use the existing `C`, so we would need to compute the whole `C` for `d^2/dx1dx2 k(x1,x2)`  or store this matrix in addition.

FelixBenning:
## kernelmatrix

@Crown421 the `kernelmatrix` thing is an issue I have not considered. Taking derivatives breaks isotropy leaving only stationarity intact. But I am also not sure why this specialization for kernelmatrix

```julia
function kernelmatrix(κ::SimpleKernel, x::AbstractVector)
    return map(x -> kappa(κ, x), pairwise(metric(κ), x))
end
```
is more performant than
```julia
function kernelmatrix(κ::SimpleKernel, x::AbstractVector)
    return broadcast(κ, x, permutedims(x))
```

I mean `pairwise(metric(κ), x) = broadcast(metric(κ), x, permutedims(x))`. So the specialized implementation does essentially
```julia
K = broadcast(x, permutedims(x) do (x,y)
      metric(κ)(x,y)
end # first pass over K
map( x->kappa(κ, x), K) # second pass over K
```
which accesses the elements of `K` twice. On the other hand
```julia
K = broadcast(x, permutedims(x) do (x,y)
      κ(x,y)
end
``` 
only requires one access. Since memory access is typically the bottleneck, the general definition should be more performant. That is unless
```julia
(κ::SimpleKernel)(x,y) = kappa(κ, metric(κ)(x,y))
```
is not inlined and causes more function calls. But in that case it probably makes more sense to force inline the code above with `@inline` this should ensure the general implementation would be reduced to
```julia
K = broadcast(x, permutedims(x) do (x,y)
      kappa(κ, metric(κ)(x,y))
end
``` 
by the compiler. Which should be faster than the two pass version.

## Issues with a wrapper

While `DiffWrapper(kernel)` may be of type `kernel` its compositions are not obvious. I mean for sums it is fine, since sum and differentiation commute. But for a function transform you do not have

```math
\frac{d}{d x_i} k(f(x), f(y)) = (\frac{\partial}{\partial x_i} k) (f(x), f(y))
```

So
```julia
DiffWrapper(kernel) ∘ FunctionTransform(f) != DiffWrapper(kernel ∘ FunctionTransform(f))
```
So if you wanted to treat `k=DiffWrapper(SqExponentialKernel())` as "the" mathematical squared exponential kernel

```math
k(x,y) = \exp\Bigl(\frac{(x-y)^2}{2}\Bigr)
```
which is differentiable, then you would expect the behavior of `DiffWrapper(kernel ∘ FunctionTransform(f))`. So you would have to specialize all the function composition for `DiffKernel`. And that feels like the main selling point of KernelFunctions.jl to me.

## Caching the cholesky decomposition

I do not understand this point. This would automatically happen with this implementation too. I mean `DiffPt(x, partial=1)` is just a special point `x` which is not in $\mathbb{R}^n$. But it still has an evaluation $y$ and a row in the cholesky matrix which can be cached. Everything should just work as is with `AbstractGP.jl`

FelixBenning:
This is really weird...
```julia
julia> @btime kernelmatrix(k, Xc);
  23.836 ms (5 allocations: 61.05 MiB)

julia> @btime map(x -> KernelFunctions.kappa(k, x), KernelFunctions.pairwise(KernelFunctions.metric(k), Xc));
  103.780 ms (8000009 allocations: 183.12 MiB)

julia> @btime k.(Xc, permutedims(Xc));
  78.818 ms (4 allocations: 30.52 MiB)

julia> size(Xc)
(2000,)

julia> size(first(Xc))
(2,)
```

Why is the performance of the implementation worse than the function itself?

src/diffKernel.jl
devmotion: Simple fix is to dispatch on `x` being a vector or a scalar.

devmotion:
> worse than the function itself?

Not sure what function you mean here and what you expect to be worse/better.

The main issue is that your benchmarking is flawed, variables etc. have to be interpolated since otherwise you suffer, sometimes massively, from type instabilities and inference issues introduced by global variables.

So instead you should perform benchmarks such as
```julia
julia> using KernelFunctions, BenchmarkTools

julia> Xc = ColVecs(randn(2, 2000));

julia> k = GaussianKernel();

julia> @btime kernelmatrix($k, $Xc);
  38.594 ms (5 allocations: 61.05 MiB)

julia> @btime kernelmatrix($k, $Xc);
  35.585 ms (5 allocations: 61.05 MiB)

julia> @btime map(x -> KernelFunctions.kappa($k, x), KernelFunctions.pairwise(KernelFunctions.metric($k), $Xc));
  37.478 ms (5 allocations: 61.05 MiB)

julia> @btime map(x -> KernelFunctions.kappa($k, x), KernelFunctions.pairwise(KernelFunctions.metric($k), $Xc));
  33.321 ms (5 allocations: 61.05 MiB)

julia> @btime $k.($Xc, permutedims($Xc));
  45.019 ms (2 allocations: 30.52 MiB)

julia> @btime $k.($Xc, permutedims($Xc));
  45.339 ms (2 allocations: 30.52 MiB)
```

> I mean pairwise(metric(κ), x) = broadcast(metric(κ), x, permutedims(x)).

No, not generally. `pairwise` for standard distances (such as `Euclidean`) is implemented in highly optimized ways in Distances (e.g., by exploiting and ensuring symmetry of the distance matrix).

> Since memory access is typically the bottleneck, the general definition should be more performant. 

Therefore this statement also does not hold generally. If you are concerned about memory allocations, probably you also might want to use `kernelmatrix!` instead of `kernelmatrix` which minimizes allocations:

```julia
julia> # Continued from above

julia> K = Matrix{Float64}(undef, length(Xc), length(Xc));

julia> @btime kernelmatrix!($K, $k, $Xc);
  25.775 ms (1 allocation: 15.75 KiB)

julia> @btime kernelmatrix!($K, $k, $Xc);
  30.012 ms (1 allocation: 15.75 KiB)
```

Another disadvantage of broadcasting is that generally it means more work for the compiler (the whole broadcasting machinery is very involved and complicated) and hence increases compilation times.

FelixBenning:
@devmotion ahh 🤦  I only looked into the `distances/pairwise.jl` file for the pairwise function. I did not know that `Distances.jl` defines this as well. This is why I hate julias `using` import mechanism and the `include` instead of import of files. You never know where functions are coming from. It is basically like `from module import *` in python which everyone dislikes for the same reason.

I guess if pairwise actually uses the symmetry of distances, then I see where the speedup in the isotropic case comes from.

devmotion:
Yes, I try to avoid `using XX` in packages nowadays and rather use `import XX` or `using XX: f, g, h` to make such relations clearer (still convenient to use `using XX` in the REPL IMO).

Crown421:
> ## Caching the cholesky decomposition
> 
> I do not understand this point. This would automatically happen with this implementation too. I mean `DiffPt(x, partial=i)` is just a special point x+∂i which is not in Rn. But it still has an evaluation y and a row in the cholesky matrix which can be cached. Everything should just work as is with `AbstractGP.jl`

My apologies, I had an error in thinking here, I was convinced that an additional matrix would need to be cached, not sure why. 

> ## Issues with a wrapper
> 
> While `DiffWrapper(kernel)` may be of type `kernel` its compositions are not obvious. I mean for sums it is fine, since sum and differentiation commute. But for a function transform you do not have
> 
> So
> 
> ```julia
> DiffWrapper(kernel) ∘ FunctionTransform(f) != DiffWrapper(kernel ∘ FunctionTransform(f))
> ```
> 
> So if you wanted to treat `k=DiffWrapper(SqExponentialKernel())` as "the" mathematical squared exponential kernel
> 
> which is differentiable, then you would expect the behavior of `DiffWrapper(kernel ∘ FunctionTransform(f))`. So you would have to specialize all the function composition for `DiffWrapper`. And that feels like the main selling point of KernelFunctions.jl to me. Of course you could tell people to only use `DiffWrapper` at the very end. But man is that ugly for zero reason.

Well, not zero reason. There are multiple reasons for using a wrapper in this PR, and therefore it comes down to opinion. It would be easy to define some fallback functions that throw an error in problematic cases, advising users to use the wrapper at the end. 

Given that differentiable kernels would not be a core feature, but rather an Extension when also loading a compatible autodiff package, any changes in main part of KernelFunctions should be minimal, and not reduce any performance. 

Therefore I would personally prefer starting with a wrapper, at least for now, to have the key functionality available and see additional issues during use. For example, I have already wondered:

1. How `DiffPt` should be treated in combination with "normal" points. You mention above mixing the two, but what should for example `vcat(ColVecs(X), DiffPt(x, partial=1))` look like?  We get performance benefits from storing a points as columns/ rows of a matrix of concrete types (i.e. `Matrix{Float64}`). Where do we put the  `partial` "annotation" of a `DiffPt`? One option could be to define new types and a load of convenience functions to make it seamless to combine them with existing ones. 

2. How does `DiffPt` combine with `MOInputs`?

3. For `MOPinput`s there is a [`prepare_isotopic_multi_output_data`](https://juliagaussianprocesses.github.io/KernelFunctions.jl/stable/api/#KernelFunctions.prepare_isotopic_multi_output_data-Tuple{AbstractVector,%20ColVecs}) method, should there be something similar for `DiffPt`s? 

For me these are important usability questions, with a much higher "ugliness" potential than where one can put a wrapper. During a normal session, I manipulate a lot of inputs and input collections, but only define a GP/ kernel once.

src/diffKernel.jl
```diff
@@ -33,18 +32,30 @@ function DiffPt(x::T, partials::NTuple{Order,KeyT}) where {T,Order,KeyT}
...
-        return func(x .+ dx * OneHotVector(idx, length(x)))
+"""
+    tangentCurve(x₀, i::IndexType)
+returns the function (t ↦ x₀ + teᵢ) where eᵢ is the unit vector at index i
+"""
+function tangentCurve(x0::AbstractArray{N,T}, idx::IndexType) where {N, T}
+    return t -> begin
+        x = similar(x0)
+        copyto!(x, x0)
+        x[idx] +=t
```
FelixBenning: @devmotion I don't know - but mutating code does feel unstable to me... this results in this error:
```julia
smoke test: Error During Test at /Users/felix/code/KernelFunctions.jl/test/diffKernel.jl:2
  Got exception outside of a @test
  InexactError: Int(Int64, Dual{ForwardDiff.Tag{ComposedFunction{KernelFunctions.var"#203#207"{Vector{Int64}, MaternKernel{Float64, Euclidean}}, KernelFunctions.var"#195#196"{Vector{Int64}, Int64}}, Int64}}(1,1))
```

FelixBenning: At least I can not make it work it seems

Crown421: Maybe you can build your own onehot implementation? Something like 
```
bv(i, n) = [j == i for j in 1:n]
```
is very performant.

FelixBenning:
@Crown421 
> Well, not zero reason. There are multiple reasons for using a wrapper in this PR, and therefore it comes down to opinion. It would be easy to define some fallback functions that throw an error in problematic cases, advising users to use the wrapper at the end.

I am starting to agree, given that I can not come up with a good solution to the kernelmatrix problem at the moment.

> How DiffPt should be treated in combination with "normal" points. You mention above mixing the two, but what should for example vcat(ColVecs(X), DiffPt(x, partial=1)) look like? We get performance benefits from storing a points as columns/ rows of a matrix of concrete types (i.e. Matrix{Float64}). Where do we put the partial "annotation" of a DiffPt? One option could be to define new types and a load of convenience functions to make it seamless to combine them with existing ones.

That is something I am currently thinking about a lot. I would think that custom composite types would be a good idea. Storing
```julia
(x, 2)
(x, n),
(y,1),
...
(y,n)
```
could be replaced and emulated by some sort of dictionary
```julia
x => [2, n]
y => 1:n
```
the advantage is, that you could specialize on index ranges to take more than one partial derivative (and use backwarddiff to get the entire gradient).

But you would still need the ability to interleave points
```julia
(x,1)
(y,2)
(x,2)
```
and I am not yet sure how to do that.

src/diffKernel.jl
devmotion: Mutating code is not type-unstable if written correctly. Writing
```julia
return t -> begin
        x = similar(promote_type(eltype(x), typeof(t)), x0)
        copyto!(x, x0)
        x[idx] +=t
```
instead should be sufficient here.

BTW the problem here is not type instability but rather that the type of `x` is sometimes not correct - the element type is not wide enough to support `t`.

devmotion: Mutation is commonly used in Julia and packages (eg SciML) to obtain highly optimized and performant implementations.

Issue #512, opened 2023-05-30
Perhaps avoid validating input dims on AbstractVector{<:AbstractVector{<:Real}}


grero:
I would suggest either changing this, or perhaps better, skipping it all together.
https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/ef6d4591b36194fca069d8bc7ae8c1e2ee288080/src/utils.jl#L194

I'm developing kernels that work on pairs of `Vector{Float64}`, where the kernel essentially sums over all pairwise distances in `x` and `y`. For this kernel, the check above does not really make sense, and indeed passes if I construct the kernel matrix like this:

```julia
K = kernelmatrix(kernel, x,x)
```
However it fails if I instead do

```julia
K = kernelmatrix(kernel, x,y)
```
unless `length(first(x))==length(first(y))`. I realise that I can just create my own type essentially wrapping `Vector{Vector{Float64}}`, but that seems like unnecessary work, especially since the kernel machinery works beautifully as it is, as long as the inputs pass the validation.

willtebbutt:
Ah, interesting. So you've got a case where the dimensionality of the input changes from element to element?

willtebbutt:
Are you hitting the checks in the various methods in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/master/src/matrix/kernelmatrix.jl ? I agree that they're making overly strong assumptions about input dimensionality, but I would like to check that they're actually the source of your problem.

grero:
Yes, that's right. For context, I'm working on using kernel methods for analysing spike trains, which are basically sets of event timestamps. The obvious way to represent this is as a vector of vector, where each inner vector represents a spike train for one trial. Since the number of events changes from trial to trial, the dimension of the inner vectors also change.

willtebbutt:
Excellent. This makes complete sense. Would you mind opening a PR which removes the checks, and adds tests which fail without the removal of the checks?

grero:
> Are you hitting the checks in the various methods in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/master/src/matrix/kernelmatrix.jl ? I agree that they're making overly strong assumptions about input dimensionality, but I would like to check that they're actually the source of your problem.

Yes, as I indicated at the top, the `validate_inputs` method fails on my inputs, unless the first element of `x` and `y` happen to have the same number of items.

grero:
PR submitted.

PR #513, opened 2023-05-31
Remove input check for vector of vectors

grero:
**Summary**

This attempts to address  #512, which prevents kernels operating on vectors of unequal length from working properly. 

**Proposed changes**

* ...
Remove the check on input dimensions when the inputs are vectors of vectors, since there is nothing wrong in principal with having kernels operate on these kinds of inputs.

The most obvious alternative would be to implement my own type, wrapping AbstractVector{<:AbstractVector{<:Real}}, but since this change would potentially benefit others with similar needs, a fix in KernelFunctions itself seemed more appropriate. 
**Breaking changes**

The main change is that inputs checks that are currently failing would pass. This is potentially breaking, as it is possible that downstream code could rely on these checks.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -191,7 +191,6 @@ pairwise(d::PreMetric, x::RowVecs, y::ColVecs) = pairwise(d, ColVecs(permutedims
 
 dim(x) = 0 # This is the passes-by-default choice. For a proper check, implement `KernelFunctions.dim` for your datatype.
 dim(x::AbstractVector) = dim(first(x))
-dim(x::AbstractVector{<:AbstractVector{<:Real}}) = length(first(x))
 dim(x::AbstractVector{<:Real}) = 1
 
 function validate_inputs(x, y)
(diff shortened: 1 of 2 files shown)


willtebbutt:
Thanks for opening this PR.

I understand why you've taken this approach, but I don't believe it's the correct approach, as whether or not the input-dimension comparison should be performed is a function of the kernel, not the input type. For example, it (should be) perfectly fine to feed a `Vector{Vector{Float64}}` into an `SEKernel`, and in that case we _would_ generally like to check that the input dimensions match.

To my mind, a better implementation would be to switch the check on / off depending upon which kernel is employed.

@theogf @devmotion @st-- do any of you have thoughts on this problem? I'm starting to wonder whether having this check is more hassle than it's worth.

devmotion:
> To my mind, a better implementation would be to switch the check on / off depending upon which kernel is employed.

Yes, I agree, that seems to be the best and most general approach. Maybe also `dim` is too specific and instead we should add a function such as `are_valid_inputs(kernel, x, y)::Bool` as the basis of `valid_inputs`?

Issue #93, opened 2020-04-23
iskroncompatible


willtebbutt:
1. `iskroncompatible` appears to be incorrectly implemented in a couple of places. For example, an `ExponentialKernel` should not be marked as compatible, or have I misunderstood something?
2. Do we really want to try and do this with a flag method? Would be not be better off introducing a `RectilinearGrid` input type (that is a subtype of `AbstractVector`) and then using dispatch to figure out whether it's correct to return a `KroneckerMatrix` when `kernelmatrix` is called?

theogf:
Back to this : 
 - 1. Yep that's a mistake, I will take care of it
 - 2. How would you figure out automatically the factorization of the kernel? Also wouldn't that lead to type instability?

willtebbutt:
I wouldn't imagine that you would need to introduce a type instability. One would just implement specialised `kernelmatrix` functions for kernels that are separable when they receive a `RectilinearGrid`, that spit out `Kronecker` matrices (e.g. from `Kronecker.jl`). Everything else in the pipeline would just specialise on that.

Am I missing something?

theogf:
Coming back to this issue:
- I think the `RectilinearGrid` solution would be too restrictive (I don't think a uniform grid is needed to compute a Kronecker matrix?). I don't see a problem having it, but I think dispatching separately for every kernel able to do it is too much unnecessary work.
- The other solution would be to rely on traits, but that would be doing the exact same thing no?

Additional point we can also define the `TensorProduct` as "`kroncompatible`"

willtebbutt:
> I think the RectilinearGrid solution would be too restrictive (I don't think a uniform grid is needed to compute a Kronecker matrix?). I don't see a problem having it, but I think dispatching separately for every kernel able to do it is too much unnecessary work.

It's just unclear to me what other contexts Kronecker product structure arises in. You need to have some kind of cartesian-product-like structure, and a rectilinear grid is the most general manifestation of that in Euclidean space, if I'm not mistaken. Maybe `CartesianProductVector` would be a better name?

devmotion:
More structured output, such as Toeplitz or circulant matrices, would be useful also when [working with Gaussian random fields](https://github.com/PieterjanRobbe/GaussianRandomFields.jl/issues/32#issuecomment-820260665).

theogf:
Yes! It's even in the README! :laughing:

devmotion:
[...]

PR #515, opened 2023-06-02
Add LazyKernelMatrix and lazykernelmatrix

### LINKED ISSUE #514: Add `LazyKernelMatrix`
In some cases we do not want to build the kernel matrix explicitely but simply want to evaluate its components when necessary (for example in `K * v`).

We should have a lazy API (or just rely on an existing one) that basically builds a `LazyKernelMatrix` containing references to the kernel and the data and implementing `getindex(...)` appropriately

sethaxen:
**Summary**

This PR introduces functionality for lazily representing kernel matrices, which is necessary when the matrix might be too large to store in memory. Fixes #514 

**Proposed changes**

* new API function `lazykernelmatrix`: supports similar semantics as `kernelmatrix` but constructs a lazy representation
* new `AbstractMatrix` subtype `LazyKernelMatrix`, constructed for the `lazykernelmatrix` default.

- I considered using one of the many existing array types in the ecosystem for lazily representing an array, but defining a novel type allows us to do things like perform scalar multiplication or add kernel matrices without densifying the array.
- it is not strictly necessary to introduce `lazykernelmatrix`, but this could allow even more structured matrix types to be defined for specific kernel types and returned in the future.
- perhaps `LazyKernelMatrix` should also store `obsdim1` and `obsdim2`? Currently we require the user has passed a vector e.g. `RowVecs` or `ColVecs`.
- perhaps we should support `y` being a `nothing` to define a symmetric kernel matrix? This would allow a couple checks to  be done at compile time. In particular we could support `+(::LazyKernelMatrix{T,Tk,Tx,Nothing}, ::Diagonal) -> LazyKernelMatrix`.

**To-Do**
- [ ] add tests

### DIFF
--- a/src/KernelFunctions.jl
+++ b/src/KernelFunctions.jl
@@ -1,6 +1,7 @@
 module KernelFunctions
 
 export kernelmatrix, kernelmatrix!, kernelmatrix_diag, kernelmatrix_diag!
+export LazyKernelMatrix, lazykernelmatrix
 export duplicate, set! # Helpers
 
 export Kernel, MOKernel
@@ -106,6 +107,7 @@ include("kernels/gibbskernel.jl")
 include("kernels/scaledkernel.jl")
 include("kernels/normalizedkernel.jl")
 include("matrix/kernelmatrix.jl")
+include("matrix/lazykernelmatrix.jl")
 include("kernels/kernelsum.jl")
 include("kernels/kernelproduct.jl")
 include("kernels/kerneltensorproduct.jl")
--- a/src/matrix/lazykernelmatrix.jl
+++ b/src/matrix/lazykernelmatrix.jl
@@ -1,0 +1,109 @@
+"""
+    lazykernelmatrix(κ::Kernel, x::AbstractVector) -> AbstractMatrix
+
+Construct a lazy representation of the kernel `κ` for each pair of inputs in `x`.
+
+The result is a matrix with the same entries as [`kernelmatrix(κ, x)`](@ref) but where the
+entries are not computed until they are needed.
+"""
+lazykernelmatrix(κ::Kernel, x) = lazykernelmatrix(κ, x, x)
+
+"""
+    lazykernelmatrix(κ::Kernel, x::AbstractVector, y::AbstractVector) -> AbstractMatrix
+
+Construct a lazy representation of the kernel `κ` for each pair of inputs in `x`.
+
+The result is a matrix with the same entries as [`kernelmatrix(κ, x, y)`](@ref) but where
+the entries are not computed until they are needed.
+"""
+lazykernelmatrix(κ::Kernel, x, y) = LazyKernelMatrix(κ, x, y)
+
+"""
+    LazyKernelMatrix(κ::Kernel, x[, y])
+    LazyKernelMatrix{T<:Real}(κ::Kernel, x, y)
+
+Construct a lazy representation of the kernel `κ` for each pair of inputs in `x` and `y`.
+
+Instead of constructing this directly, it is better to call
+[`lazykernelmatrix(κ, x[, y])`](@ref lazykernelmatrix).
+"""
+struct LazyKernelMatrix{T<:Real,Tk<:Kernel,Tx<:AbstractVector,Ty<:AbstractVector} <:
+       AbstractMatrix{T}
+    kernel::Tk
+    x::Tx
+    y::Ty
+    function LazyKernelMatrix{T}(κ::Tk, x::Tx, y::Ty) where {T<:Real,Tk<:Kernel,Tx,Ty}
+        Base.require_one_based_indexing(x)
+        Base.require_one_based_indexing(y)
+        return new{T,Tk,Tx,Ty}(κ, x, y)
+    end
+    function LazyKernelMatrix{T}(κ::Tk, x::Tx) where {T<:Real,Tk<:Kernel,Tx}
+        Base.require_one_based_indexing(x)
+        return new{T,Tk,Tx,Tx}(κ, x, x)
+    end
+end
+function LazyKernelMatrix(κ::Kernel, x::AbstractVector, y::AbstractVector)
+    # evaluate once to get eltype
+    T = typeof(κ(first(x), first(y)))
+    return LazyKernelMatrix{T}(κ, x, y)
+end
+LazyKernelMatrix(κ::Kernel, x::AbstractVector) = LazyKernelMatrix(κ, x, x)
+
+Base.Matrix(K::LazyKernelMatrix) = kernelmatrix(K.kernel, K.x, K.y)
+function Base.AbstractMatrix{T}(K::LazyKernelMatrix) where {T}
+    return LazyKernelMatrix{T}(K.kernel, K.x, K.y)
+end
+
+Base.size(K::LazyKernelMatrix) = (length(K.x), length(K.y))
+
+Base.axes(K::LazyKernelMatrix) = (axes(K.x, 1), axes(K.y, 1))
+
+function Base.getindex(K::LazyKernelMatrix{T}, i::Int, j::Int) where {T}
+    return T(K.kernel(K.x[i], K.y[j]))
+end
+for f in (:getindex, :view)
+    @eval begin
+        function Base.$f(
+            K::LazyKernelMatrix{T},
+            I::Union{Colon,AbstractVector},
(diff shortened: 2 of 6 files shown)


devmotion:
> this could allow even more structured matrix types to be defined for specific kernel types and returned in the future

Yes, I imagine this would be useful also in the context of https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/93#issuecomment-820262209. Though there maybe often the non-lazy version might be sufficient (or even preferable).

src/KernelFunctions.jl
```diff
@@ -1,6 +1,7 @@
 module KernelFunctions
 
 export kernelmatrix, kernelmatrix!, kernelmatrix_diag, kernelmatrix_diag!
+export LazyKernelMatrix, lazykernelmatrix
```
devmotion, line 4: Do we have to export both? Is `lazykernelmatrix` sufficient maybe?

theogf:
It looks really nice already! Only thing which is a bit uneasy is the output type... I am not sure there is a strong guarantee that the first evaluated type would correspond to the rest of the matrix. But I also don't see how it could be solved easily...

src/matrix/lazykernelmatrix.jl
```diff
@@ -0,0 +1,109 @@
+"""
+    lazykernelmatrix(κ::Kernel, x::AbstractVector) -> AbstractMatrix
+
+Construct a lazy representation of the kernel `κ` for each pair of inputs in `x`.
+
+The result is a matrix with the same entries as [`kernelmatrix(κ, x)`](@ref) but where the
+entries are not computed until they are needed.
+"""
+lazykernelmatrix(κ::Kernel, x) = lazykernelmatrix(κ, x, x)
```
devmotion, line 9: It would be good to optimize this for the symmetric case, IMO, similar to `kernelmatrix` (which IIRC often does not use such a fallback but more optimized methods).

FelixBenning:
For this a lazy ProductArray would also be a neat abstraction:

```julia
julia> v = rand(3)
3-element Vector{Float64}:
 0.417571623820013
 0.39972694171008405
 0.9970727095536318

julia> productArray(v,v)
3×3 ProductArrays.ProductArray{Tuple{Vector{Float64}, Vector{Float64}}, Tuple{Float64, Float64}, 2}:
 (0.417572, 0.417572)  (0.417572, 0.399727)  (0.417572, 0.997073)
 (0.399727, 0.417572)  (0.399727, 0.399727)  (0.399727, 0.997073)
 (0.997073, 0.417572)  (0.997073, 0.399727)  (0.997073, 0.997073)
```

now the lazy kernelmatrix is simply a lazy [mappedarray](https://github.com/JuliaArrays/MappedArrays.jl) of this product array.

This is neat, because the productArray abstraction is also helpful as a multioutput abstraction:
```julia
julia> vec(productArray(v,1:2))
6-element reshape(::ProductArrays.ProductArray{Tuple{Vector{Float64}, UnitRange{Int64}}, Tuple{Float64, Int64}, 2}, 6) with eltype Tuple{Float64, Int64}:
 (0.417571623820013, 1)
 (0.39972694171008405, 1)
 (0.9970727095536318, 1)
 (0.417571623820013, 2)
 (0.39972694171008405, 2)
 (0.9970727095536318, 2)
```

cf. https://github.com/lazyLibraries/ProductArrays.jl (not yet a package https://github.com/JuliaRegistries/General/pull/84683)

devmotion:
I have the same feeling as @sethaxen:

> I considered using one of the many existing array types in the ecosystem for lazily representing an array, but defining a novel type allows us to do things like perform scalar multiplication or add kernel matrices without densifying the array.

A separate dedicated type provides more information about the context and allows us to define dispatches, special operations and optimizations that might be less relevant or not well defined in the general case.

src/matrix/lazykernelmatrix.jl
FelixBenning: The optimization for the symmetric case is only calculating half the matrix lets say the top half and then redirecting all queries from the bottom half to the top half. (Actually distances simply copies the top half into the bottom half). 

Since this is lazy, there is probably no point in this optimization because you do not do the calculation from the start. And when you call getindex it does not matter whether you calculate the element in the top or bottom half.

FelixBenning:
> A separate dedicated type provides more information about the context and allows us to define dispatches, special operations and optimizations that might be less relevant or not well defined in the general case.

I edited my reply to explain how you can still do special dispatches

src/matrix/lazykernelmatrix.jl
devmotion: At least with other lazy iterators in Base it's a common pattern to `collect` results at some point (e.g., after filtering, mapping, etc.). In this case it seems beneficial to know that the lazy matrix is symmetric.

FelixBenning:
I actually found this pull request because I wanted to ask: does it ever make sense for kernelmatrix to be eager? I mean memory access is expensive and you take $n^2$ space instead of the natural $n$ for just the vector. And kernelmatrices are often only accessed once (e.g. when you want take its cholesky decomposition) and afterwards tossed. Are some kernels so expensive to evaluate to justify this?

FelixBenning:
The Transform test seem very much broken to me, it breaks here, #511 and #510 in pretty much the same way. And those have nothing to do with each other.

src/KernelFunctions.jl
sethaxen: Probably.

sethaxen:
> Only thing which is a bit uneasy is the output type... I am not sure there is a strong guarantee that the first evaluated type would correspond to the rest of the matrix. But I also don't see how it could be solved easily...

I wonder of it's safe to ask Julia to infer the return type of `kernelmatrix` and use that to infer the eltype.

FelixBenning:
> > Only thing which is a bit uneasy is the output type... I am not sure there is a strong guarantee that the first evaluated type would correspond to the rest of the matrix. But I also don't see how it could be solved easily...
> 
> I wonder of it's safe to ask Julia to infer the return type of `kernelmatrix` and use that to infer the eltype.

mappedarrays does eltype inference 🤷

willtebbutt:
A completely separate option would be to be a little less ambitious with this initial implementation, and not implement a new matrix type, and just provide an interface for the operation that we want.

Specifically, add the following function to the interface:
```julia
function kernel_matrix_vector_product(k::Kernel, x::AbstractVector, y::AbstractVector, v::AbstractVector{<:Real})
    return kernelmatrix(k, x, y) * v
end
kernel_matrix_vector_product(k::Kernel, x::AbstractVector, v::AbstractVector{<:Real}) = kernelmatrix(k, x) * v
```
where the above methods are the default implementations and specify the semantics.
For large problems we could wrap the kernel in some other type which says "don't ever instantiate me", and implements a low-memory version of this operation.

The nice thing about doing things this way is that we avoid having to e.g. guess at the output type of `kernelmatrix`, or whatever. It's a little verbose, but maybe it covers our initial requirements?

Does this cover our needs? I guess really I'm just asking whether we actually need to go to the trouble of implementing a new matrix type, and can instead just implement a single function that can be overloaded.

FelixBenning:
ProductArrays v1.0.0 is now online, so you could do

```julia
using MappedArrays: mappedarray, ReadonlyMappedArray
using ProductArrays: productArray, ProductArray

struct Splat{T} 
      func::T
end

(s::Splat)(x) = s.func(x...)

lazykernelmatrix(k::Kernel, x, y) = mappedarray(Splat(k), productArray(x,y))

const LazyKernelMatrix{K<:Kernel, T<:Real, P<:ProductArray} = ReadonlyMappedArray{T, 2, P, Splat{K}}
```
and then implement additional functionality for `LazyKernelmatrix` (you would get a ton of functionality from MappedArrays for free (like Eltype inference)

```julia
julia> a = mappedarray(Splat(k), productArray(x,x))
3×3 mappedarray(Splat{SqExponentialKernel{Distances.Euclidean}}(Squared Exponential Kernel (metric = Distances.Euclidean(0.0))), ::ProductArray{Tuple{Vector{Float64}, Vector{Float64}}, Tuple{Float64, Float64}, 2}) with eltype Float64:
 1.0       0.879512  0.998656
 0.879512  1.0       0.901716
 0.998656  0.901716  1.0

julia> a isa LazyKernelMatrix
true

julia> eltype(a)
Float64

julia> a isa AbstractArray
true
```

PR #521, opened 2023-06-26
Make Test a weak dependency

devmotion:
I'm not sure if this is a good idea at all and if it works as intended - BUT I've encountered multiple repos in the last few months where people did not want to add dependencies on Test. So I was wondering if we could get away with making the test utilities an extension of the package and wanted to try it in KernelFunctions.

IMO, the main problems are that 1) the code is a bit more complex and contains a "workaround" (since only overloaded but not new functions in extensions are accessible for the user) and 2) the extension is always loaded when testing the PR in the REPL.

I assume the reason for 2) might be that the Test stdlib is included in the default system image (?). Or are stdlibs not supported as weakly dependencies (I guess @KristofferC should know this?)?


devmotion:
This approach was adopted by eg AbstractFFTs, TranscodingStreams, and (soon) Distributions. I think we should make Test a weak dependency to reduce the number of dependencies and to be nice to other packages that made Test a weak dependency.

willtebbutt:
I agree that we should do this -- I'm contemplating doing it with some of my other repos as well anyway.

PR #523, opened 2023-09-12
Make pullback error for ColVecs and RowVecs a bit more informative

torfjelde:
The adjoint defined for `ColVecs` and `RowVecs` indicates that the cause of the error might be something internal to KernelFunctions.jl. But other packages are also making use of `ColVecs` and `RowVecs`, e.g. AbstractGPs.jl, which in turn means that issues often encountered in practice are _not_ related to what the adjoint-error indicates, e.g. https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/issues/344

I ran into this yesterday and because of the error message I spent a fair bit of time looking for bugs in the kernel-related code rather than looking elsewhere. Hence the PR:)

My wording in the error message can probably do with an improvement?

EDIT: Oh and why does this adjoint definition exist? Why not just pull back the vector of vectors to a matrix? AFAIK the adjoint is still valid?

### DIFF
--- a/src/chainrules.jl
+++ b/src/chainrules.jl
@@ -151,7 +151,7 @@ function ChainRulesCore.rrule(::Type{<:ColVecs}, X::AbstractMatrix)
         return error(
             "Pullback on AbstractVector{<:AbstractVector}.\n" *
             "This might happen if you try to use gradients on the generic `kernelmatrix` or `kernelmatrix_diag`,\n" *
-            "or because some other external computation has acted on `ColVecs` to produce a vector of vectors." *
+            "or because some external computation has acted on `ColVecs` to produce a vector of vectors." *
             "If it is the former, to solve this issue overload `kernelmatrix(_diag)` for your kernel for `ColVecs`",
         )
     end
@@ -164,7 +164,7 @@ function ChainRulesCore.rrule(::Type{<:RowVecs}, X::AbstractMatrix)
         return error(
             "Pullback on AbstractVector{<:AbstractVector}.\n" *
             "This might happen if you try to use gradients on the generic `kernelmatrix` or `kernelmatrix_diag`,\n" *
-            "or because some other external computation has acted on `RowVecs` to produce a vector of vectors." *
+            "or because some external computation has acted on `RowVecs` to produce a vector of vectors." *
             "If it is the former, to solve this issue overload `kernelmatrix(_diag)` for your kernel for `RowVecs`",
         )
     end
--- a/src/chainrules.jl
+++ b/src/chainrules.jl
@@ -152,7 +152,9 @@ function ChainRulesCore.rrule(::Type{<:ColVecs}, X::AbstractMatrix)
             "Pullback on AbstractVector{<:AbstractVector}.\n" *
             "This might happen if you try to use gradients on the generic `kernelmatrix` or `kernelmatrix_diag`,\n" *
             "or because some external computation has acted on `ColVecs` to produce a vector of vectors." *
-            "If it is the former, to solve this issue overload `kernelmatrix(_diag)` for your kernel for `ColVecs`",
+            "In the former case, to solve this issue overload `kernelmatrix(_diag)` for your kernel for `ColVecs`." *
+            "In the latter case, one needs to track down the `rrule` whose pullback returns a `Vector{Vector{T}}`," *
+            " rather than a `Tangent`, as the cotangent / gradient for `ColVecs` input, and circumvent it."
         )
     end
     return ColVecs(X), ColVecs_pullback
--- a/src/chainrules.jl
+++ b/src/chainrules.jl
@@ -150,8 +150,9 @@ function ChainRulesCore.rrule(::Type{<:ColVecs}, X::AbstractMatrix)
     function ColVecs_pullback(::AbstractVector{<:AbstractVector{<:Real}})
         return error(
             "Pullback on AbstractVector{<:AbstractVector}.\n" *
-            "This might happen if you try to use gradients on the generic `kernelmatrix` or `kernelmatrix_diag`.\n" *
-            "To solve this issue overload `kernelmatrix(_diag)` for your kernel for `ColVecs`",
+            "This might happen if you try to use gradients on the generic `kernelmatrix` or `kernelmatrix_diag`,\n" *
+            "or because some other external computation has acted on `ColVecs` to produce a vector of vectors." *
+            "If it is the former, to solve this issue overload `kernelmatrix(_diag)` for your kernel for `ColVecs`",
         )
     end
     return ColVecs(X), ColVecs_pullback
@@ -162,8 +163,9 @@ function ChainRulesCore.rrule(::Type{<:RowVecs}, X::AbstractMatrix)
     function RowVecs_pullback(::AbstractVector{<:AbstractVector{<:Real}})
         return error(
             "Pullback on AbstractVector{<:AbstractVector}.\n" *
-            "This might happen if you try to use gradients on the generic `kernelmatrix` or `kernelmatrix_diag`.\n" *
-            "To solve this issue overload `kernelmatrix(_diag)` for your kernel for `RowVecs`",
+            "This might happen if you try to use gradients on the generic `kernelmatrix` or `kernelmatrix_diag`,\n" *
+            "or because some other external computation has acted on `RowVecs` to produce a vector of vectors." *
+            "If it is the former, to solve this issue overload `kernelmatrix(_diag)` for your kernel for `RowVecs`",
         )
     end
     return RowVecs(X), RowVecs_pullback
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "KernelFunctions"
 uuid = "ec8451be-7e33-11e9-00cf-bbf324bd1392"
-version = "0.10.56"
+version = "0.10.57"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"


src/chainrules.jl
```diff
@@ -150,8 +150,9 @@ function ChainRulesCore.rrule(::Type{<:ColVecs}, X::AbstractMatrix)
     function ColVecs_pullback(::AbstractVector{<:AbstractVector{<:Real}})
         return error(
             "Pullback on AbstractVector{<:AbstractVector}.\n" *
-            "This might happen if you try to use gradients on the generic `kernelmatrix` or `kernelmatrix_diag`.\n" *
-            "To solve this issue overload `kernelmatrix(_diag)` for your kernel for `ColVecs`",
+            "This might happen if you try to use gradients on the generic `kernelmatrix` or `kernelmatrix_diag`,\n" *
+            "or because some other external computation has acted on `ColVecs` to produce a vector of vectors." *
```
torfjelde, line 154:
```suggestion
            "or because some external computation has acted on `ColVecs` to produce a vector of vectors." *
```

src/chainrules.jl
```diff
@@ -162,8 +163,9 @@ function ChainRulesCore.rrule(::Type{<:RowVecs}, X::AbstractMatrix)
     function RowVecs_pullback(::AbstractVector{<:AbstractVector{<:Real}})
         return error(
             "Pullback on AbstractVector{<:AbstractVector}.\n" *
-            "This might happen if you try to use gradients on the generic `kernelmatrix` or `kernelmatrix_diag`.\n" *
-            "To solve this issue overload `kernelmatrix(_diag)` for your kernel for `RowVecs`",
+            "This might happen if you try to use gradients on the generic `kernelmatrix` or `kernelmatrix_diag`,\n" *
+            "or because some other external computation has acted on `RowVecs` to produce a vector of vectors." *
```
torfjelde, line 167:
```suggestion
            "or because some external computation has acted on `RowVecs` to produce a vector of vectors." *
```

devmotion:
> EDIT: Oh and why does this adjoint definition exist? Why not just pull back the vector of vectors to a matrix? AFAIK the adjoint is still valid?

To avoid surprisingly slow code. `ColVecs` and `RowVecs` wrap a concatenated matrix of the vectors and usually it is faster to work with that underlying matrix instead of constructing a vector of vectors.

The code was introduced in #84 initially (moved to ChainRules in #208) but as it was copied from Stheno.jl maybe @willtebbutt has some additional comments.

st--:
Looks good to me, can you also bump the patch version?

torfjelde:
> To avoid surprisingly slow code. 

That's what I suspected, but is it worth it given how difficult it can be to debug these adjoint issues for most users? :grimacing:

devmotion:
Difficult to debug implies that probably it would be even more difficult to notice the problem and the cause for the slow performance without the error 😛

torfjelde:
I think the error makes complete sense in the scenario where all usages of `ColVecs` and `RowVecs` are internal, but that it becomes somewhat more nuanced when it can interact with external code.

torfjelde:
For example, the issue referenced above, is caused because we end up silently taking the "slow path", _but_ we're doing this because there exists a default implementation for `AbstractGPs.mean_vector`. So I guess my question is then boiled down to, why allow this slow default method to be hit but not the slow pull back?

devmotion:
Without looking into all details, if AbstractGPs takes a slow path, that's a bug in AbstractGPs that should be fixed there. No method in KernelFunctions should support slow paths for ColVecs and RowVecs. But of course we want to be as generic as possible so we define methods for `AbstractVector` - we just always want to take the optimal path for `ColVecs`/`RowVecs`.

src/chainrules.jl
```diff
@@ -150,8 +150,9 @@ function ChainRulesCore.rrule(::Type{<:ColVecs}, X::AbstractMatrix)
     function ColVecs_pullback(::AbstractVector{<:AbstractVector{<:Real}})
         return error(
             "Pullback on AbstractVector{<:AbstractVector}.\n" *
-            "This might happen if you try to use gradients on the generic `kernelmatrix` or `kernelmatrix_diag`.\n" *
-            "To solve this issue overload `kernelmatrix(_diag)` for your kernel for `ColVecs`",
+            "This might happen if you try to use gradients on the generic `kernelmatrix` or `kernelmatrix_diag`,\n" *
+            "or because some external computation has acted on `ColVecs` to produce a vector of vectors." *
```
willtebbutt, line 154:
```julia
            "or because some external computation has acted on `ColVecs` to produce a vector of vectors." *
```
this isn't quite right. I think it would be more accurate to say that "a pullback has produced a vector-of-vectors as the cotangent (gradient) of something w.r.t. a `ColVecs`".

Same for `RowVecs`

willtebbutt: Could we maybe talk about how to debug this? i.e. say something about hunting through earlier bits of the reverse-pass to figure out where the `Vector{Vector{T}}` came from?

torfjelde:
> if AbstractGPs takes a slow path, that's a bug in AbstractGPs that should be fixed there. No method in KernelFunctions should support slow paths for ColVecs and RowVecs

Ah gotcha, then it makes sense :+1: 

> What I was primarily trying to guard against was the forwards-pass taking the fast path, and the pullback somehow hitting the slow path (I believe that this bit me several times when I hit a piratic rrule).

Yeah I can see how an incorrect `rrule` could cause issues, so it def seems sensible :+1: To me the confusion was mainly when a forward-pass hits the slow path, then we should be happy with the reverse also taking the slow path.

But so this is in general a "restriction" with `ColVecs` and `RowVecs` then? You're not allowed to AD through "slow paths", such as the example from AbstractGPs?

src/chainrules.jl
torfjelde: Agreed. Something like
```
In the latter case, one needs to track down the call that treats the `ColVecs` as a simple vector and either
add an explicit definition of the forward-pass for `ColVecs` input, or define an explicit adjoint.
```
?

willtebbutt: I think we can be a little more precise:
```julia
In the latter case, one needs to track down the `rrule` whose pullback returns a `Vector{Vector{T}}`, rather than a `Tangent`, as the cotangent / gradient for `ColVecs` input, and circumvent it.
```
Or is this too technical?

willtebbutt:
> But so this is in general a "restriction" with ColVecs and RowVecs then? You're not allowed to "automatically" AD through "slow paths", such as the example from AbstractGPs?

Depends on the kind of slow path. If when you AD the slow path, you produce a `Vector{Vector{T}}` cotangent for a `ColVecs`/`RowVecs`, then no. If it's a slow path, but you get the right type, it'll work fine.

torfjelde:
> If when you AD the slow path, you produce a Vector{Vector{T}} cotangent for a ColVecs/RowVecs, then no. If it's a slow path, but you get the right type, it'll work fine.

With "slow path" I mean when a method that works with `AbstractVector` or similar receives a `ColVecs`, e.g. `map`. In these scenarios, the evaluation/forward pass is allowed and works just fine, even though you're hitting a slow path, but AD is _not_ allowed. Have I understood correctly?

willtebbutt:
> With "slow path" I mean when a method that works with AbstractVector or similar receives a ColVecs, e.g. map. In these scenarios, the evaluation/forward pass is allowed and works just fine, even though you're hitting a slow path, but (specifically reverse-mode) AD is not allowed. Have I understood correctly?

Kind of. I'm just saying it's more specific than that in that the reverse-pass must return a `Vector{Vector{T}}` rather than a `Tangent{ColVecs{T}}` or whatever. It happens to be the case that this tends to align with when the forwards-pass does something slow and annoying, but you could imagine situations in which this isn't the case. e.g. `map(sum, ColVecs(randn(10, 20)))` isn't going to be particularly slow. My point is that it's not so much to do with slowness, as it is type errors.

torfjelde:
>  It happens to be the case that this tends to align with when the forwards-pass does something slow and annoying, but you could imagine situations in which this isn't the case. e.g. map(sum, ColVecs(randn(10, 20))) isn't going to be particularly slow. My point is that it's not so much to do with slowness, as it is type errors.

Sure, that I'm with; I was referring to "slow paths" because I thought that was the original motivation as to why this `rrule` was implemented as is. 

But is it understandable that it's somewhat confusing that something like `map(sum, ColVecs(randn(10, 20)))` is allowed but computing the pullback of this is not?

willtebbutt:
> But is it understandable that it's somewhat confusing that something like map(sum, ColVecs(randn(10, 20))) is allowed but computing the pullback of this is not?

It's entirely understandable, but I felt (still feel?) that it's the lesser of two evils -- it's opting for a loud error when performance is bad (read: catastrophic) on the reverse-pass, rather than allowing the code to run very slowly, but eventually produce an answer.

torfjelde:
Gotcha, gotcha :+1: 

Thanks for explaining the decision! I'm curious about this stuff partially because of the discussions we've had regarding more general batching structures in the past.

src/chainrules.jl
st--: I don't think it's too technical... better to be explicit. Maybe link to some GitHub issue open exactly for explaining this (if we assume it's something that might disappear sometime in the future)? At least then people would have a place to ask for help how to get code working, and it could include more examples of how to fix common cases...

src/chainrules.jl
```diff
@@ -150,8 +150,9 @@ function ChainRulesCore.rrule(::Type{<:ColVecs}, X::AbstractMatrix)
...
+            "If it is the former, to solve this issue overload `kernelmatrix(_diag)` for your kernel for `ColVecs`",
```
torfjelde, line 155:
```suggestion
            "This might happen if you try to use gradients on the generic `kernelmatrix` or `kernelmatrix_diag`,\n" *
            "or because some external computation has acted on `ColVecs` to produce a vector of vectors." *
            "In the former case, to solve this issue overload `kernelmatrix(_diag)` for your kernel for `ColVecs`." *
            "In the latter case, one needs to track down the `rrule` whose pullback returns a `Vector{Vector{T}}`," *
            " rather than a `Tangent`, as the cotangent / gradient for `ColVecs` input, and circumvent it."
```

willtebbutt (approved):
Thanks for sorting this out @torfjelde . I'll merge when CI passes.

willtebbutt:
Something has broken with AD 🤦 . It's clearly not this PR's fault, so I'm going to merge and tag a release anyway.

Issue #526, opened 2023-09-24
Failures on master


simsurace:
The BaseKernels failures on master can be traced back to moving ChainRules from 1.52.0 to 1.53.0. I think some weird interaction betwen ChainRules and Zygote pullbacks is at play.
The upside is that they are resolved by using https://github.com/JuliaStats/Distances.jl/pull/246.

simsurace:
Actually, I re-tested now. This works by doing one of these things:
- Pin ChainRules to v1.52.0
- Remove ZygoteDistancesExt from Zygote https://github.com/FluxML/Zygote.jl/pull/1460 and fix DistancesChainRulesCoreExt https://github.com/JuliaStats/Distances.jl/pull/256

simsurace:
Ok, we are getting there! Any ideas on what the last remaining failure is about?
```
Zygote: Error During Test at /home/runner/work/KernelFunctions.jl/KernelFunctions.jl/test/transform/selecttransform.jl:75
  Got exception outside of a @test
  ArgumentError: unable to check bounds for indices of type Symbol
  Stacktrace:
    [1] checkindex(::Type{Bool}, inds::Base.OneTo{Int64}, i::Symbol)
      @ Base ./abstractarray.jl:756
    [2] checkindex
      @ Base ./abstractarray.jl:773 [inlined]
    [3] checkbounds_indices
      @ Base ./abstractarray.jl:725 [inlined]
    [4] checkbounds
      @ Base ./abstractarray.jl:678 [inlined]
    [5] checkbounds
      @ Base ./abstractarray.jl:699 [inlined]
    [6] view(::Matrix{Float64}, ::Vector{Symbol}, ::Base.Slice{Base.OneTo{Int64}})
      @ Base ./subarray.jl:179
    [7] ∇getindex!(::Matrix{Float64}, ::Matrix{Float64}, ::Vector{Symbol}, ::Vararg{Any})
      @ ChainRules ~/.julia/packages/ChainRules/Gw0tZ/src/rulesets/Base/indexing.jl:151
    [8] ∇getindex(::AxisMatrix{Float64, Matrix{Float64}, Tuple{Axis{:row, Vector{Symbol}}, Axis{:col, Vector{Symbol}}}}, ::Matrix{Float64}, ::Vector{Symbol}, ::Vararg{Any})
      @ ChainRules ~/.julia/packages/ChainRules/Gw0tZ/src/rulesets/Base/indexing.jl:89
[...]

PR #529, opened 2023-10-06
Stop using deprecated signatures from Distances.jl

simsurace:
We are using a deprecated signature of `Distances.pairwise!`, leading to distracting warnings.
This PR fixes those.


devmotion (commented):
Good to fix those! Two main comments:
- The deprecations and in particular the new syntax was only added in https://github.com/JuliaStats/Distances.jl/pull/239, so we have to bump the Distances compat entry to 0.10.9.
- We should change the function signature of `KernelFunctions.pairwise!` in the same way.

willtebbutt:
Thanks for catching this. If you could please bump the patch version, I'm happy to approve

src/distances/pairwise.jl
```diff
@@ -6,11 +6,11 @@ end
 
 pairwise(d::PreMetric, X::AbstractVector) = pairwise(d, X, X)
 
-function pairwise!(out::AbstractMatrix, d::PreMetric, X::AbstractVector, Y::AbstractVector)
+function pairwise!(d::PreMetric, out::AbstractMatrix, X::AbstractVector, Y::AbstractVector)
```
devmotion, line 9: Are there any downstream packages that use `KernelFunctions.pairwise!`? Probably not I assume? Otherwise we would have to copy the deprecations from Distances to avoid breaking downstream code.

willtebbutt: There shouldn't be, and I don't believe that we declare it to be part of the public API of KernelFunctions, so if people are depending on it, they really shoudn't be.

simsurace: I won't be releasing this yet (I think we can wait for #526 to be fully resolved before tagging another release). So if this is discussion is not a blocker, I would propose to merge this. If you change your mind, we can still add deprecation warnings later.

simsurace:
Bump :)

devmotion (approved):
Looks good to me.

PR #530, opened 2023-10-07
Relax allocation test

simsurace:
Fixes part of #526. For some reason I have not been able to figure out, starting from Julia 1.9 the allocations in these AD calls can be off by one from call to call. This PR relaxes the test slightly. I think this still serves the original purpose of the test.


simsurace:
Indeed

devmotion (approved):

PR #539, opened 2023-11-11
Remove chainrule and test for `SqMahalanobis`

simsurace:
Just some cleanup work to remove remnants of the Mahalanobis kernel which has been absent for a while.
- They are no longer needed
- The chainrule is committing type piracy towards Distances.jl (would there be value to move the rule in the extension there?)
- The test is triggering some warnings that are distracting

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "KernelFunctions"
 uuid = "ec8451be-7e33-11e9-00cf-bbf324bd1392"
-version = "0.10.59"
+version = "0.10.60"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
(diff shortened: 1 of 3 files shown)


willtebbutt (approved):
Approved subject to patch bump, so that we can tag a release

simsurace:
Hi @willtebbutt, the version was already bumped, but I haven't released it just yet (was hoping to figure out the failures on master first). Do we merge this anyway and release it with the other fixes once they are done?

willtebbutt:
Ah okay, I hadn't realised we were in this situation. As a rule, we never leave an unreleased version of the package on master -- if a PR (with patch bump) is good enough for a release.

Could you please release the current version of master, then bump + release this? I grant you that it's annoying that there are failures on master, but unless you think that they're related to a PR that has recently been merged, I don't feel that they ought to disrupt the release cycle.

simsurace:
No the remaining failures are likely due to relying on some internals in Julia, and were introduced by the Julia 1.8 -> 1.9 update. I will do as you suggest.

PR #546, opened 2024-02-06
Fix `@test_broken` problems

simsurace:
Fixes part of #526

### DIFF
--- a/test/basekernels/periodic.jl
+++ b/test/basekernels/periodic.jl
@@ -16,6 +16,7 @@
     TestUtils.test_interface(PeriodicKernel(; r=[0.8, 0.7]), RowVecs{Float64})
 
     # test_ADs(r->PeriodicKernel(r =exp.(r)), log.(r), ADs = [:ForwardDiff, :ReverseDiff])
-    @test_broken "Undefined adjoint for Sinus metric, and failing randomly for ForwardDiff and ReverseDiff"
+    # Undefined adjoint for Sinus metric, and failing randomly for ForwardDiff and ReverseDiff
+    @test_broken false
     test_params(k, (r,))
 end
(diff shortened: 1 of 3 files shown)


willtebbutt (approved):
This is great. Please merge if you're happy.

simsurace:
> There are more bugs regarding `@test_broken` such as
> 
> https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/f71cfe0bc5d2c54e927641f897823cf6baf91ac8/test/transform/selecttransform.jl#L107
> : `@test_broken` should only applied to expressions that evaluate to a `Bool` (which in the broken case is `false` but `true` if the problem is fixed).

I see. Why don't they show up as errors though?
What is the test supposed to do anyway? It is just evaluating a gradient and assigning it to `ga`. What is broken about this?

devmotion:
Probably it throws an error currently which is handled by `@test_broken` as well. So that case is currently fine - but it's still wrong since the tests will error as soon as the gradient computation does not throw an error anymore. See https://discourse.julialang.org/t/psa-when-tests-fail-due-to-improvements-in-type-inference/102001

simsurace:
Would it be reasonable to replace these by an approximate equality comparing the chosen AD backend to `:FiniteDiff`?

devmotion:
I guess these are the intended tests: https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/f71cfe0bc5d2c54e927641f897823cf6baf91ac8/test/transform/selecttransform.jl#L76-L82 So for the ReverseDiff we could just not create the intermediate variables and directly `@test_broken` the approximate equality of the two gradients?

simsurace:
This should do it.

devmotion (approved):

PR #531, opened 2023-10-16
Fix periodic kernel AD

simsurace:
This PR reactivates the AD tests for the periodic kernel and introduces `rrules` to make them pass.
Fixes #527

PR #554, opened 2024-03-10
Fix KernelFunctions on Julia 1.3

devmotion:
https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/531 (included in KernelFunctions 0.10.62 and 0.10.63) broke KernelFunctions on Julia 1.3, which broke support and tests in downstream packages such as AbstractGPsMakie. The problem was that in CI the oldest supported Julia version was not tested.

I think it would be fine to drop support for Julia < 1.6 (even though I don't see the need as long as it works and does not impact development significantly) but before changing the Julia compat of KernelFunctions we have to fix support on Julia >= 1.3: Otherwise Pkg will always give people on such older Julia versions a broken KernelFunctions version (that currently can't even be loaded). Alternatively, we could mess with the registry and try to fix compat bounds retroactively but in my experience it's generally considered better practice to register bug fix releases than modifying existing releases (e.g. yanking a release can also break existing Manifest.toml files).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,8 +21,8 @@ jobs:
       fail-fast: false
       matrix:
         version:
+          - '1.3'
           - '1'
-          - '1.6'
           - 'nightly'
         os:
           - ubuntu-latest
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "KernelFunctions"
 uuid = "ec8451be-7e33-11e9-00cf-bbf324bd1392"
-version = "0.10.63"
+version = "0.10.64"
 
 [deps]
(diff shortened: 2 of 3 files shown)


theogf (approved):

theogf:
@devmotion It looks like the errors on 1.3 are still relevant?

willtebbutt:
I would be strongly in favour of bumping our minimum required version to the current LTS release, 1.10.8.

simsurace:
Is this still relevant after CI has dropped 1.3?

willtebbutt:
IMHO, it's not. Also, this is presumably now stale. Going to close, but @devmotion should of course feel free to reopen.

simsurace:
Shall we bump the Julia compat to 1.6?

PR #559, opened 2024-07-04
Gaussion Processes : added navbar workflow

jeetsuthar:
We are adding this to all Turing satellite packages! See here - https://github.com/TuringLang/turinglang.github.io/blob/main/assets/scripts/README.md

We also announced this navigation bar in our Turing slack's general channel!

### DIFF
--- a/.github/workflows/DocsNav.yml
+++ b/.github/workflows/DocsNav.yml
@@ -1,0 +1,44 @@
+name: Add Navbar
+
+on:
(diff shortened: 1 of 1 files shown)


willtebbutt:
See https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/pull/403#issuecomment-2209209557 -- I need more info before considering accepting this.

PR #562, opened 2025-01-06
Bump compat constraint on Functors to 0.5

axsk:
**Summary**

**Proposed changes**

* ...

**Breaking changes**


theogf:
You would need to make sure that the new behaviour of `@functor` is still correct (it looks like now being a leaf is opt-in rather than opt-out)

theogf:
I will take a closer look.

theogf:
It looks like the tests are failing like on  master, however inspecting the resolution it looks like 0.4.12 was used in the tests. @axsk could you also update the compat in `test/Project.toml`?

sathvikbhagavan:
A gentle bump on this. Its blocking upgrading Flux in Surrogates.jl

PR #568, opened 2025-03-04
Update check_zygote_type_stability()...

gnadt:
so that CI passes with Julia v1.10. Tested locally with Julia v1.10 & v1.11. Forces the expected type to be `Vector{Union{Nothing, @NamedTuple{X::Matrix{Float64}}}}` which is more robust for different package versions. There may be better solutions, but this seems to work.


willtebbutt:
I think it's a shame that we need to do this, but if @theogf is happy for it to be merged into his branch, I'm also happy for it to go in, as the lack of robustness of our AD tests is a _real_ problem.

theogf:
Ah sorry, I did not realize you were targeting my branch...

